PR #13, opened 2012-06-01
WICKET-4580 StringValue: missing methods to convert from text to wrapped classes of primitive type with defaultValue

minman:
Fix for WICKET-4580 StringValue: missing methods to convert from text to wrapped classes of primitive type with defaultValue


klopfdreh:
As of the JIRA ticket this topic wouldn't be fixed - @minman would you be so kind and close this PR? Thank you.

PR #74, opened 2014-06-22
Support of getting and setting properties customly in PropertyResolver

PhantomYdn:
There is no way to customize getting of properties from the object. All possible ways to gain properties are hardcoded in PropertyResolver. 
But, for example, there is set of NoSQL document DBs which have some class "Document" to work with DB. And to get some property of that document you should call something like this: 'doc.getProperty("myPropertyName")'.

This pull request enables possibility to specify custom GetAndSet for your type of object. Please take a look to PropertyResolverTest

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/core/util/lang/PropertyResolver.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/util/lang/PropertyResolver.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


klopfdreh:
It would be nice if you could file in a ticket in JIRA so that it can be catched up by the Wicket Dev's. For more information take a look here: https://cwiki.apache.org/confluence/display/WICKET/Contributing+to+Wicket

martin-g:
https://issues.apache.org/jira/browse/WICKET-5623

svenmeier:
I'm working on a different solution, please take a look at https://github.com/apache/wicket/tree/WICKET-5623-extensible-propertyresolver

PhantomYdn:
Looks interesting and promising:)
Could you please also fix silly current wicket issue with that stuff which currently exist?
Issue is that if there is only a setter and no getter: PropertyResolver will fail.

And what's target release? I hope that some 7.XX?

svenmeier:
> if there is only a setter and no getter, PropertyResolver will fail.

Please open a separate issue for that.

I'm planning this change for 6.x, 7.x and 8.x

PhantomYdn:
Created: https://issues.apache.org/jira/browse/WICKET-6201

Also - it will be very helfull to make getObjectWithGetAndSet public. Sometimes it's usefully to introspect object by wicket utilities. We switched in [Orienteer](http://orienteer.org) from spring bean introspection to wicket due not accurate working with beans in Spring (for example if there is a class with methods like: public String getValue(); public void setValue(Object value) - Spring is failing measurably even if property "value" was not directly requested).

svenmeier:
Please close this pull request: WICKET-5623 adds IPropertyLocator.

PhantomYdn:
Thanks for implementation of this feature!

PR #87, opened 2015-02-04
WICKET-5827 Allow to apply multiple Javascript / CSS compressors

klopfdreh:
- Allow multiple ICssCompressors and IJavaScriptCompressors
- CssUrlReplacementCompressor

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
+++ b/wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
@@ -45,14 +45,7 @@ public class CompositeCssCompressor implements ICssCompressor
 {
 
 	/* Compressors to compress the CSS content */
-	private List<ICssCompressor> compressors = new ArrayList<>();
-
-	/**
-	 * Initializes an empty composite CSS compressor
-	 */
-	public CompositeCssCompressor()
-	{
-	}
+	private final List<ICssCompressor> compressors = new ArrayList<>();
 
 	/**
 	 * Initializes the composite CSS compressor with the given {@link ICssCompressor}
@@ -62,7 +55,7 @@ public class CompositeCssCompressor implements ICssCompressor
 	 */
 	public CompositeCssCompressor(ICssCompressor... compressors)
 	{
-		this.compressors = Arrays.asList(compressors);
+		this.compressors.addAll(Arrays.asList(compressors));
 	}
 
 	/**
--- a/wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
+++ b/wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.resource;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.wicket.css.ICssCompressor;
+
+/**
+ * Used to apply several {@link ICssCompressor} to the css compression.<br>
+ * <br>
+ * Usage:
+ * 
+ * <pre>
+ * CompositeCssCompressor compositeCssCompressor = new CompositeCssCompressor();
+ * 
+ * compositeCssCompressor.getCompressors().add(new MyCssCompressor());
+ * compositeCssCompressor.getCompressors().add(new AnotherCssCompressor());
+ * 
+ * this.getResourceSettings().setCssCompressor(compositeCssCompressor);
+ * </pre>
+ * @since 6.20.0
+ * @author Tobias Soloschenko
+ * 
+ */
+public class CompositeCssCompressor implements ICssCompressor
+{
+
+	/* Compressors to compress the CSS content */
+	private List<ICssCompressor> compressors;
+
+	/**
+	 * Compresses the given original content in the order of compressors.
+	 */
+	@Override
+	public String compress(String original)
+	{
+		if (compressors != null)
+		{
+			String compressed = original;
+			for (ICssCompressor compressor : compressors)
+			{
+				compressed = compressor.compress(compressed);
+			}
+			return compressed;
+		}
+		else
+		{
+			return original;
+		}
+	}
+
+	/**
+	 * Gets a list of {@link ICssCompressor} to be used for CSS compression. They are applied in the
+	 * order of the List.
+	 * 
+	 * @return A list of {@link ICssCompressor} to be used for CSS compression.
+	 */
+	public List<ICssCompressor> getCompressors()
+	{
+		if (compressors == null)
+		{
+			compressors = new ArrayList<>();
+		}
+		return compressors;
+	}
+}
--- a/wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
+++ b/wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
@@ -17,12 +17,13 @@
 package org.apache.wicket.resource;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import org.apache.wicket.css.ICssCompressor;
 
 /**
- * Used to apply several {@link ICssCompressor} to the css compression.<br>
+ * Used to apply several {@link ICssCompressor} to the CSS compression.<br>
  * <br>
  * Usage:
  * 
@@ -34,6 +35,8 @@ import org.apache.wicket.css.ICssCompressor;
  * 
  * this.getResourceSettings().setCssCompressor(compositeCssCompressor);
  * </pre>
+ * The compressors can also be given as constructor arguments.
+ * 
  * @since 6.20.0
  * @author Tobias Soloschenko
  * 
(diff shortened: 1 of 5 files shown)


wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
```diff
@@ -0,0 +1,82 @@
...
+ * @since 6.20.0
+ * @author Tobias Soloschenko
+ * 
+ */
+public class CompositeCssCompressor implements ICssCompressor
+{
+
+	/* Compressors to compress the CSS content */
+	private List<ICssCompressor> compressors;
+
```
bitstorm: Personally I would initialize compressors in-line so we can skip 'if (compressors != null)'  checks. In addition I would also add a constructor with  multiple args for compressors : 'CompositeCssCompressor(ICssCompressor...)' .
Same for CompositeJavaScriptCompressor.

klopfdreh: Just one moment. :-)

klopfdreh: Done

wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
```diff
@@ -0,0 +1,93 @@
...
+ * 
+ * @since 6.20.0
+ * @author Tobias Soloschenko
+ * 
+ */
+public class CompositeCssCompressor implements ICssCompressor
+{
+
+	/* Compressors to compress the CSS content */
+	private List<ICssCompressor> compressors = new ArrayList<>();
```
martin-g: final

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
```diff
@@ -0,0 +1,144 @@
...
+ * 
+ */
+public class CssUrlReplacementCompressor implements ICssCompressor
+{
+
+	// Holds the names of pages
+	private Map<String, String> pageNames = Collections.synchronizedMap(new LinkedHashMap<String, String>());
+
+	// The pattern to find URLs in CSS resources
+	private Pattern urlPattern = Pattern.compile("url\\(['|\"](.*)['|\"]\\)");
```
martin-g: static final

klopfdreh: This is not possible because of Arrays.asList(compressors); in constructor

martin-g: I think `"` and `'` are optional. I.e. `url(some/path/name.png)` is also valid.

wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
klopfdreh: This is not possible because of Arrays.asList(compressors); in constructor.

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
klopfdreh: ok I check it and change the regex so that ' and " are optional

wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * Initializes an empty composite CSS compressor
+	 */
+	public CompositeCssCompressor()
```
martin-g: This constructor can be removed. The second one covers the use case of empty argument list.
Define the field as: `private final List<ICssCompressor> compressors;`

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
```diff
@@ -0,0 +1,144 @@
...
+ * 
+ * <pre>
+ * this.getResourceSettings().setCssCompressor(new CssUrlReplacementCompressor(this));
+ * </pre>
+ * 
+ * @since 6.20.0
+ * @author Tobias Soloschenko
+ * 
+ */
+public class CssUrlReplacementCompressor implements ICssCompressor
```
martin-g: If the class doesn't compress then I think it should not have `Compressor` as a suffix.

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
klopfdreh: url\\(['|\"]*(.*)['|\"]*\\)

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
```diff
@@ -0,0 +1,144 @@
...
+	 */
+	@SuppressWarnings("unchecked")
+	@Override
+	public String compress(String original)
+	{
+		Matcher matcher = this.urlPattern.matcher(original);
+		// Search for urls
+		while (matcher.find())
+		{
+			Collection<String> pageNames = this.pageNames.keySet();
```
martin-g: This logic looks very expensive :-/

wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
klopfdreh: Changed the line to:
this.compressors.addAll(Arrays.asList(compressors));

wicket-core/src/main/java/org/apache/wicket/resource/CompositeCssCompressor.java
klopfdreh: Done

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
```diff
@@ -0,0 +1,144 @@
...
+			for (String pageName : pageNames)
+			{
+				try
+				{
+					Class<Page> pageClass = (Class<Page>)Class.forName(pageName);
```
martin-g: Use `WicketObjects.resolveClass()` instead.
OSGi users will be happier!

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
```diff
@@ -0,0 +1,144 @@
...
+					String url = matcher.group(1);
+					if (!url.contains("/"))
+					{
+						URL urlResource = pageClass.getResource(url);
```
martin-g: This is the heavy part.
I'll have to think for a better solution.

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
klopfdreh: Mhh, any other suggestions? - the matcher has to be applied once so that I can find all url positions.

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
martin-g: Or even better store org.apache.wicket.util.reference.ClassReference

klopfdreh: done

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
klopfdreh: I used this solution because then you don't need to add PackageResourceReferences to be resolved to the CssUrlReplacer

martin-g: I'm not even sure this will work for all cases.
Imagine I have a Panel in `com.example.panels.MyPanel`, it contributes a CSS from `res/css/my.css`, and this CSS has image: `../images/my.png`.
My pages are in `com.example.pages.**`.
So having knowledge about pages won't help me to find the image.

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
```diff
@@ -0,0 +1,144 @@
...
+							urlBuilder.append(replacedUrl);
+							urlBuilder.append("')");
+							original = matcher.replaceFirst(urlBuilder.toString());
+						}
+					}
+				}
+				catch (Exception e)
+				{
+					StringWriter stringWriter = this.printStack(e);
+					throw new WicketRuntimeException(stringWriter.toString());
```
martin-g: Why ?! :-)
Just do: `throw new WicketRuntimeException("A problem occurred during CSS url replacement.", e);`

wicket-core/src/main/java/org/apache/wicket/resource/CompositeJavaScriptCompressor.java
```diff
@@ -0,0 +1,94 @@
...
+	/**
+	 * Initializes the composite javascript compressor with the given {@link IJavaScriptCompressor}
+	 * 
+	 * @param compressors
+	 *            several {@link IJavaScriptCompressor} the composite javascript compressor is
+	 *            initialized with
+	 */
+	public CompositeJavaScriptCompressor(IJavaScriptCompressor... compressors)
+	{
+		this.compressors = Arrays.asList(compressors);
```
martin-g: Same as CompositeCssCompressor.

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
klopfdreh: Yes thats right - thats a limitation.

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
klopfdreh: done

wicket-core/src/main/java/org/apache/wicket/resource/CompositeJavaScriptCompressor.java
klopfdreh: done

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
klopfdreh: done

wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacementCompressor.java
klopfdreh: Mhh, the problem here is that you can use any scope in PackageResourceReferences, even classes which are not the components that use the image. In this case it is hard to find a generic way to get those images. What we could do is let away the check of pages so that the images of panels could also be found.

martin-g:
The knowledge is there: https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/request/resource/CssPackageResource.java#L65
The problem is that it is not propagated.
I.e. CssPackageResource knows the scope class, but it doesn't give it to the ICssCompressor.

martin-g:
A possible improvement:
- introduce IScopeAwareCssProcessor {process(String nonCompressed, Class scope, String cssName)}
- check if the resolved ICssCompressor implements the new interface and use it if it does

klopfdreh:
What do you think of leaving the check of page away in the first step and then implement your suggestions in a further pull request?

klopfdreh:
Ok I think I got it, but the changes are not backward compatible to wicket 6.x

klopfdreh:
So for now, yep the CssUrlReplacer is a bit slow, because it is scanning for the images in different scopes and it is a limitation that you could not put images in other packages as those of the components, but I would recommend the changes of @martin-g in a separate pull request, because they require changes in various classes and they are not backward compatible. What do you think?

martin-g:
I'll take of this PR!

klopfdreh:
Thank you very much! I also applied the changes we discussed here for the wicket-6.x branch - https://github.com/apache/wicket/pull/88

PR #90, opened 2015-02-14
WICKET-5837 JUnit tests may fail because of AbstractDefaultAjaxBehavior

klopfdreh:


klopfdreh:
Linked ticket: https://issues.apache.org/jira/browse/WICKET-5837

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
```diff
@@ -233,12 +246,6 @@ protected final CharSequence renderAjaxAttributes(final Component component,
 				attributesJson.put(AjaxAttributeName.MARKUP_ID.jsonName(), componentId);
 			}
```
bitstorm: I'm just curious...why have you moved the 5 lines below at the beginning of the method (line 223)?

klopfdreh: Because the Unit Tests check the Order of the Parameter

wicket-core/src/test/java/org/apache/wicket/markup/html/form/upload/MultiFileUploadFieldTest.java
```diff
@@ -63,13 +64,16 @@ protected void onSubmit()
 				for (int i = 1; i < 2; i++)
 				{
 					FileUpload fileUpload = uploads.get(i);
-					assertEquals(MultiFileUploadFieldTest.class.getSimpleName()+i+".txt", fileUpload.getClientFileName());
+					assertEquals(MultiFileUploadFieldTest.class.getSimpleName() + (i - 1) + ".txt",
```
bitstorm: There's a problem with '(i - 1)'. I've reverted it to simply 'i' and test passes with success.  Can you confirm that this change is not necessary?

klopfdreh:
I run the test with i-1 and it was success, but maybe the code at this place changed a bit. So if it passing revert it.

Thanks a lot!

wicket-core/src/test/java/org/apache/wicket/markup/html/form/upload/MultiFileUploadFieldTest.java
klopfdreh: I run the test with i-1 and it was success, but maybe the code at this place changed a bit. So if it passing revert it.

Thanks a lot!

klopfdreh:
@bitstorm - everything ok with this PR - or should I change something?

There are two ways to proceed - one is to change all 14 unit tests and the other is to change the order of the parameters at the central place.

bitstorm:
actually I have already applied your commit to 6.x (hash 1a6f1bde1a6b576f39bf6fd6d5129d577319c2a7 ). So now its ok. I forgot to notify you about it.

klopfdreh:
Ah ok! :-) - thanks a lot - so I'm closing this PR, now!

PR #107, opened 2015-03-06
WICKET-5851 Fix PackageResourceTest and PackageResourceGuardTest under Windows

shuraa:
Please, review PR.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/PackageResourceGuard.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/PackageResourceGuard.java
@@ -85,7 +85,8 @@ public class PackageResourceGuard implements IPackageResourceGuard
 		int len = path.length();
 		final String ext;
 		if (ixExtension <= 0 || ixExtension == len ||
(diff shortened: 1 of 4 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/html/PackageResourceGuard.java
```diff
@@ -85,7 +85,8 @@ protected boolean acceptAbsolutePath(String path)
 		int len = path.length();
 		final String ext;
 		if (ixExtension <= 0 || ixExtension == len ||
-			(path.lastIndexOf(File.separator) + 1) == ixExtension)
```
martin-g: What is wrong here with `File.separator` ?

shuraa: Check PackageResourceGuardTest:

assertTrue(guard.acceptAbsolutePath("c:\\test\\org\\apache\\.java"));
assertTrue(guard.acceptAbsolutePath("c:/test/org/apache/.java"));

Under Win one can use both separator styles.

PR #114, opened 2015-03-31
IoC: Optionally use objensis for proxy creation

selckin:
This allows creating proxies for concrete classes without a default
constructor.

Guice IoC cache proxies similar to spring

-----

It may need some  more finishing touches, but would something like this be considered for inclusion?

### DIFF
--- a/wicket-guice/src/main/java/org/apache/wicket/guice/GuiceFieldValueFactory.java
+++ b/wicket-guice/src/main/java/org/apache/wicket/guice/GuiceFieldValueFactory.java
@@ -19,21 +19,25 @@ package org.apache.wicket.guice;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.util.concurrent.ConcurrentMap;
 
 import javax.inject.Qualifier;
 
 import org.apache.wicket.injection.IFieldValueFactory;
-import org.apache.wicket.proxy.IProxyTargetLocator;
 import org.apache.wicket.proxy.LazyInitProxyFactory;
 
 import com.google.inject.BindingAnnotation;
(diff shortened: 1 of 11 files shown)


martin-g:
Hi,

I see this PR introduces some API changes.
We want to release Wicket 7.0.0 next week so I am a bit hesitant to merge it.
Please start a discussion at dev@wicket.apache.org. If others think it is something important then we will merge it.

selckin:
Now with less api changes

wicket-guice/src/main/java/org/apache/wicket/guice/GuiceFieldValueFactory.java
```diff
@@ -64,15 +68,34 @@ public Object getFieldValue(final Field field, final Object fieldOwner)
 				{
 					boolean optional = injectAnnotation != null && injectAnnotation.optional();
 					Annotation bindingAnnotation = findBindingAnnotation(field.getAnnotations());
-					final IProxyTargetLocator locator = new GuiceProxyTargetLocator(field, bindingAnnotation, optional);
+					final GuiceProxyTargetLocator locator = new GuiceProxyTargetLocator(field, bindingAnnotation, optional);
+
+					Object cachedValue = cache.get(locator);
```
martin-g: The caching is not related to Objenesis, right?

selckin: It's not, but i'm not sure about the objensis performance, and the spring ioc also caches the proxies, so it keeps those 2 closer together aswel

martin-g: I'm going to revert all non-Objenesis caches.
This should be in separate ticket.

PR #118, opened 2015-05-15
WICKET-5906 Move StringResourceModel to a fluid API

gsmet:
I was wondering if maybe setModel and setParameters should be more clear, such as setSubstitutionModel and setSubstitutionParameters.

Comments welcome.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/model/StringResourceModel.java
+++ b/wicket-core/src/main/java/org/apache/wicket/model/StringResourceModel.java
@@ -26,6 +26,7 @@ import org.apache.wicket.Localizer;
 import org.apache.wicket.Session;
 import org.apache.wicket.core.util.string.interpolator.PropertyVariableInterpolator;
 import org.apache.wicket.resource.loader.ComponentStringResourceLoader;
+import org.apache.wicket.util.lang.Args;
 import org.apache.wicket.util.string.Strings;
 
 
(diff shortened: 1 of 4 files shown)


wicket-core/src/main/java/org/apache/wicket/model/StringResourceModel.java
```diff
@@ -275,110 +276,70 @@ public void setObject(String object)
...
 	 * @param component
 	 *            The component that the resource is relative to
-	 * @param model
-	 *            The model to use for property substitutions
-	 * @param parameters
-	 *            The parameters to substitute using a Java MessageFormat object
 	 */
-	public StringResourceModel(final String resourceKey, final Component component,
-		final IModel<?> model, final Object... parameters)
+	public StringResourceModel(final String resourceKey, final Component component)
```
martin-g: To make the migration easier I think it would be good to have one more constructor: 
`public StringResourceModel(final String resourceKey, final Component component, final IModel<?> model)`.

gsmet: And keep the setModel? Or should I also remove it?

My goal was to have "model" and "parameters" on the same level.

martin-g: Keep it.

gsmet:
Done. The tests tests both flavours.

martin-g:
The PR is merged but GitHub doesn't recognize it.
@gsmet Please close the PR.
Thanks!

gsmet:
Thanks for merging it.

PR #119, opened 2015-05-19
WICKET-5910 Do not intercept the protected methods with CGLib.

gsmet:


wicket-ioc/src/main/java/org/apache/wicket/proxy/LazyInitProxyFactory.java
```diff
@@ -157,14 +164,26 @@ else if (type.isInterface())
 		}
 		else
 		{
-			CGLibInterceptor handler = new CGLibInterceptor(type, locator);
+			CGLibInterceptor handler = new CGLibInterceptor(type, locator) {
+				@Override
+				public Object intercept(Object object, Method method, Object[] args, MethodProxy proxy)
+						throws Throwable {
+					// TODO Auto-generated method stub
+					return super.intercept(object, method, args, proxy);
```
martin-g: What is the purpose of this override? It seems to just call super.

wicket-ioc/src/main/java/org/apache/wicket/proxy/LazyInitProxyFactory.java
```diff
@@ -349,6 +368,33 @@ public Object writeReplace() throws ObjectStreamException
...
+	{
+		private static final long serialVersionUID = 1L;
+	}
+
+	/**
+	 * CGLib callback filter: does not intercept protected methods.
+	 * 
+	 * Otherwise CGLib throws an IllegalArgumentException: protected method.
+	 */
+	private static class CGLibCallbackFilter implements CallbackFilter {
```
martin-g: Please add a link to the description of the problem with protected methods in the Javadoc.

wicket-ioc/src/main/java/org/apache/wicket/proxy/LazyInitProxyFactory.java
gsmet: Urgh, sorry, it was a leftover of my work on it. Shouldn't have been there.

gsmet:
I think I addressed your concerns.

PR #127, opened 2015-06-16
Added constructor to Websocket filters 

pulse00:
Currently it's not possible to use the Websocket filters in pure java-based servlet 3.0 configurations without subclassing the Filter and  overriding the `getApplicationFactory` method. 

This PR adds the `WicketFilter(Webapplication application)` constructor to all websocket filters.

### DIFF
--- a/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/AbstractUpgradeFilter.java
+++ b/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/AbstractUpgradeFilter.java
@@ -16,6 +16,13 @@
  */
 package org.apache.wicket.protocol.ws;
 
+import org.apache.wicket.ThreadContext;
+import org.apache.wicket.protocol.http.WebApplication;
+import org.apache.wicket.protocol.http.WicketFilter;
+import org.apache.wicket.request.cycle.RequestCycle;
+import org.apache.wicket.request.http.WebResponse;
+import org.apache.wicket.util.string.Strings;
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Enumeration;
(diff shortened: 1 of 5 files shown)


martin-g:
What exactly is the use case ?

pulse00:
In our application, we setup the servlet context purely in java, without any `web.xml`. This looks something like this:

```java
Constructor<? extends CustomWebApplication> constructor = applicationClass.getConstructor(StageType.class);
CustomWebApplication CustomWebApplication = constructor.newInstance(activeStateType);
CustomWebApplication.setApplicationContext(rootContext);

WicketFilter wicketFilter = new WicketFilter(CustomWebApplication) {

  @Override
  public void init(boolean isServlet, FilterConfig filterConfig) throws ServletException {
    setFilterPath("");
    super.init(isServlet, filterConfig);
  }
};

servletContext.addFilter("hstsFilter", HSTSFilter.class).addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), false, "/*");

FilterRegistration.Dynamic wicketFilterRegistration = servletContext.addFilter("wicketFilter", wicketFilter);
wicketFilterRegistration.addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), true, "/*");
```

Now if we want to add websocket support by adding using a `WebsocketFilter`, we need to create a custom subclass because the constructor with the `Webapplication` argument is not public in the subclasses of `WicketFilter`: 

```java
private class CustomWebsocketFilter extends Tomcat7WebSocketFilter {

  private WebApplication application;

  public CustomWebsocketFilter(WebApplication application) {
    this.application = application;
  }

  @Override
  protected IWebApplicationFactory getApplicationFactory() {
    return new IWebApplicationFactory() {
      @Override
      public WebApplication createApplication(WicketFilter filter) {
        return application;
      }

      @Override
      public void destroy(WicketFilter filter) {
        if (application != null)
        {
          try
          {
            ThreadContext.setApplication(application);
            application.internalDestroy();
          }
          finally
          {
            ThreadContext.detach();
            application = null;
          }
        }
      }
    };
  }
}
```
At least that's how i understood the [documentation](https://github.com/pulse00/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/protocol/http/WicketFilter.java#L109) of this specific constructor.

martin-g:
I see.
In this case you should also add a constructor without any arguments, like https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/protocol/http/WicketFilter.java#L100-L106. Otherwise now instantiating from web.xml will fail.

pulse00:
@martin-g thanks, i didn't know that. The no-args constructor has been added.

pulse00:
thanks.

PR #130, opened 2015-06-24
WICKET-5935: Guice IoC: cache proxies & fail fast

selckin:
- cache proxies in the same way the spring ioc caches them

- If a bean has no binding the exception is now throw on construction
instead of when that bean/proxy is used.


martin-g:
The PR has been merged!
Please close it because we do not own this mirror.
Thank you!

PR #132, opened 2015-07-17
Add untouchPage() to PageManagerDecorator

frido37:


wicket-core/src/main/java/org/apache/wicket/page/PageManagerDecorator.java
```diff
@@ -59,6 +59,14 @@ public void touchPage(IManageablePage page)
 		delegate.touchPage(page);
 	}
 
+  public void untouchPage(IManageablePage page)
```
martin-g: Different whitespaces ...
I'll make the change manually.

martin-g:
The change is applied!
Thanks for testing before the release!

martin-g:
Please close the PR. We don't have the rights.

PR #134, opened 2015-08-13
Added return type for chaining

tatjana19:
As it is in Component

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/IGenericComponent.java
+++ b/wicket-core/src/main/java/org/apache/wicket/IGenericComponent.java
@@ -39,16 +39,18 @@ public interface IGenericComponent<T>
 	 *
 	 * @param model
 	 *            the new model
+	 * @return This for chaining
 	 */
-	void setModel(IModel<T> model);
+	Component setModel(IModel<T> model);
 
 	/**
 	 * Typesafe setter for the model object
 	 *
 	 * @param object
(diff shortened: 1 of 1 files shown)


martin-g:
This is an API break and will have to wait Wicket 8.x.

tatjana19:
Of course because of semantic versioning i think

tatjana19:
Another option would be to create a new inteface like this:

----------------------------------------------------
package org.apache.wicket;

import org.apache.wicket.model.IModel;

/**
 * An interface for all {@link Component components} with type-safe accessors and mutators
 * for the model and its object.
 *
 * @param <T>
 *     the type of the model object
 */
public interface IGenericModel<T>
{
	/**
	 * Typesafe getter for the model
	 *
	 * @return the model
	 */
	IModel<T> getModel();

	/**
	 * Typesafe setter for the model
	 *
	 * @param model
	 *            the new model
	 * @return This for chaining
	 */
	Component setModel(IModel<T> model);

	/**
	 * Typesafe setter for the model object
	 *
	 * @param object
	 *            the new model object
	 * @return This for chaining
	 */
	Component setModelObject(T object);

	/**
	 * Typesafe getter for the model's object
	 *
	 * @return the model object
	 */
	T getModelObject();
} 

----------------------------------------------------

The IGenericComponent is for use for generic Model so the new name IGenericModel would fit better i think

PR #135, opened 2015-08-13
Removed some final modifiers

Ragnaroek:
We needed to overwrite this methods in our projects, would be nice to get this upstream to get rid of our patched version.


shuraa:
Please, do not change both code formatting and code refactoring.

PR #136, opened 2015-08-13
Removed some final modifiers

Ragnaroek:
New pull request (now without formatting pollution).

We needed to remove those modifiers in our project.


martin-g:
Please start a discussion at Wicket's dev@ (or users@) mailing list with explanation of your use cases.
IMO only the method in Localizer could be improved. Removing `final` from the other two could break Wicket internals really bad if the application is not really careful.

Ragnaroek:
Hi Martin,

renderAll was overwritten by us to add some performance profiling, something like this:
<startPerfProfiling>
super.renderAll()
<stopPerfProfiling>


addStateChange() was overwritten to prevent the increment of a page version number, since we manipulated the component hierarchy after rendering, but before the page gets stored. This is a very deep manipulation of Wicket, we are fully aware of that, but it was essential for our high traffic scenario.

martin-g:
Please discuss at dev@ mailing list.
Very few people actually read here.

PR #137, opened 2015-09-24
Changed return type of

astrapi69:
Program to Interface, not Implementation


martin-g:
I'm afraid this change could not be applied.
It is an API break and we could not merge it in 7.x branch (currently master).

The methods are `static` so they are not overrideable, so I think there is no big issue here.

Thanks for the PR anyway!

astrapi69:
Ok than for release 8.x :-)

PR #138, opened 2015-10-04
Fixed 'uberjar' support for initializers

robertbraeutigam:
Comment for 'break' says no need to have multiple initializer properties files. Actually when creating a single jar file as a distribution (with jetty, wicket, everything included) as opposed to a 'war' file, it is necessary to read *all* initializer files, as all of them will be included in the single distributed jar file.

In my case, there are 2 initializers: core and extensions. With this modifications it all works in a single 'executable' jar file.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Application.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Application.java
@@ -514,7 +514,6 @@ public abstract class Application implements UnboundListener, IEventSink
 							Properties properties = new Properties();
 							properties.load(jarEntryStream);
 							load(properties);
-							break; // atm there is no need to have more than one .properties file
 						}
 					}
 				}


martin-g:
This will slow down the startup for all "normal" (WAR) applications :-/
Could you please create a quickstart to play with? I think we can improve further by `break`-ing once we exit `/META-INF/wicket/` directory. I.e. I think iterating over the jar entries will go over all entries in `/META-INF/wicket/` and then process other directories. So once it enters in `/META-INF/wicket/` we can raise a flag, later when it exits then we should `break`.

robertbraeutigam:
Here you go: https://github.com/robertbraeutigam/wicket-uberjar-sandbox

As far as I know, there is no "exiting" and "entering" directories in Jar files, since it is composed of 'entries' in no particular order. I don't think we can get around that, unless we can get rid of scanning the classpath altogether somehow.

robertbraeutigam:
As for slowing down "normal" applications, I think the real solution would be not to do classpath scanning, but to use java.util.ServiceLoader or similar approach. The "shade" maven plugin for example can deal with merging these standard Service files at least, so that would solve this "uberjar" issue and would be even faster (and more standard-compliant) than the current solution.

martin-g:
> java.util.ServiceLoader or similar approach

Maybe for Wicket 8 ...
Unless there is a backward compatible way to do this now.

I remember both maven-shade-plugin and OSGi had problems with ServiceLoader. Maybe they are fixed in latest versions.

martin-g:
The was a discussion at Wicket dev@ mailing list and we decided to deprecate the current way and use ServiceLoader from now on.
Please see https://issues.apache.org/jira/browse/WICKET-5997 and http://markmail.org/message/o5jvwjmmezu4xebe for more details.
Please close this PR because we don't have permissions to do it!
Thank you for raising this issue!

PR #139, opened 2015-10-14
WICKET-5999 fix IFeedback visitor in AjaxFormValidatingBehavior

shuraa:


shuraa:
@svenmeier I've reworked the patch.

svenmeier:
That looks better.

PR #140, opened 2015-10-17
Allow the Wicket ChannelManager to be used by plain Ajax requests

dhs3000:
Allows to replace the jQuery ajax function by one that uses Wicket’s channel manager to coordinate ajax requests in third party libraries like “select2”.

Could be used for example for an autocomplete select2 component where some requests might take longer. The Wicket channel manager drops all new request (by users input) but the last request as long as one request is still processed by the server.


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -1253,6 +1253,48 @@
...
+			this.type = channelType;
+			this.channel = this.name + "|" + this.type;
+			this.channelManager = channelManager;
+			this.transport = ajaxTransport? ajaxTransport : jQuery.ajax;
+		},
+
+		ajax: function(url, options) {
+			if (typeof url === "object") {
+				options = url;
+				url = undefined;
```
martin-g: This looks wrong. IMO if `url` is Object then:
 * if `options` is Object too then merge them
 * if `options` is undefined then: `options = url; url = options.u || options.url`

There is no unit test for `url` being Object.

martin-g:
Thanks for the contribution!
But since Wicket itself won't use it I think it should not be part of wicket-ajax-jquery.js. It could be part of Wicket-Select2 project, for example.

Let's see what other Wicket devs think about it.

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
dhs3000: That sounds right. I updated the argument handling and the unit tests.

svenmeier:
I agree with Martin here: Why have a method in Wicket, when it's not used by Wicket?

martin-g:
@dhs3000 Please close this PR. We don't have permissions to do it ourselves.

PR #142, opened 2015-11-01
FileSystemResourceReference Implementation

klopfdreh:
http://apache-wicket.1842946.n4.nabble.com/VOTE-Integration-of-the-FileSystemResourceReference-td4672449.html


wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -0,0 +1,131 @@
...
+ * For more information see {@link org.apache.wicket.markup.html.media.FileSystemResourceReference}
+ * 
+ * @author Tobias Soloschenko
+ *
+ */
+public class FileSystemResource extends AbstractResource
+{
+	private static final long serialVersionUID = 1L;
+
+	private Path path;
```
martin-g: final

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -0,0 +1,131 @@
...
+
+	/**
+	 * Creates a new file system resource based on the given path
+	 * 
+	 * @param path
+	 *            the path to be read for the resource
+	 */
+	public FileSystemResource(Path path)
+	{
+		this.path = path;
```
martin-g: Args.notNull

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -0,0 +1,131 @@
...
+
+	/**
+	 * Adjusts the resource response by overriding it with additional information
+	 * 
+	 * @param resourceResponse
+	 *            the resource response to adjust
+	 * 
+	 * @return the resource response to be adjusted
+	 */
+	protected ResourceResponse adjustResourceResponse(ResourceResponse resourceResponse)
```
martin-g: Maybe rename to `configure'. I think we use 'configure' elsewhere.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -0,0 +1,131 @@
...
+	/**
+	 * Gets the mime type to be used for the response
+	 * 
+	 * @return the mime type to be used for the response
+	 * @throws IOException
+	 *             if the mime type could'nt be resoulved
+	 */
+	protected String getMimeType() throws IOException
+	{
+		return Files.probeContentType(path);
```
martin-g: At https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/core/util/resource/UrlResourceStream.java#L122-L129 we first give the application the chance to return custom MIME type. I think we should do the same here.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
```diff
@@ -0,0 +1,175 @@
...
+ * </code>
+ * </pre>
+ * 
+ * @author Tobias Soloschenko
+ */
+public class FileSystemResourceReference extends ResourceReference
+{
+	private static final long serialVersionUID = 1L;
+
+	private Path path;
```
martin-g: final

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
```diff
@@ -0,0 +1,175 @@
...
+		if (fileSystem == null)
+		{
+			if (env == null)
+			{
+				env = new HashMap<>();
+				env.put("create", "true");
+				env.put("encoding", "UTF-8");
+			}
+			fileSystem = FileSystems.newFileSystem(new URI(zipFile), env);
+			fileSystemURIs.put(uri, fileSystem);
```
martin-g: What is the reason of the caching of the file systems here?
It has some issues related to concurrency. Better use ConcurrentMap with `#putIfAbsent()`.
I'd prefer if Wicket doesn't cache it though. I can foresee some issues with redeployments and OSGi usage.

wicket-core/src/test/java/org/apache/wicket/resource/FileSystemResourceReferenceTestFileTypeDetector.java
```diff
@@ -0,0 +1,38 @@
...
+ * @author Tobias Soloschenko
+ *
+ */
+public class FileSystemResourceReferenceTestFileTypeDetector extends FileTypeDetector
+{
+
+	@Override
+	public String probeContentType(Path path) throws IOException
+	{
+		return "text/plain_provided_by_detector";
```
martin-g: use `if (path.getFileName().toString().endsWith("our test file name here"))` to avoid potential issues in the future

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
klopfdreh: I used the cache here because I didn't find a way to check if the file system has been created before. Thats the only reason.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
klopfdreh: Okay - good point!

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
klopfdreh: Will be done!

wicket-core/src/test/java/org/apache/wicket/resource/FileSystemResourceReferenceTestFileTypeDetector.java
klopfdreh: Maybe check the name of the path and else throw an exception.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
klopfdreh: Well be done!

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
klopfdreh: Will ne done!

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
klopfdreh: Will be done!

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
martin-g: The question is: is it expensive to create a new FileSystem instance every time it is needed? Maybe it is cached by the provider

wicket-core/src/test/java/org/apache/wicket/resource/FileSystemResourceReferenceTestFileTypeDetector.java
martin-g: java.nio.file.spi.FileTypeDetector#probeContentType's javadoc says: 
`@return  The content type or {@code null} if the file type is not recognized`

martin-g: Because of `wicket-core/src/test/java/META-INF/services/java.nio.file.spi.FileTypeDetector` this impl is available for *all* tests in the classpath. By using 'if' we can isolate it to do its job only for this particular test.

klopfdreh: Yep I know why you wanted to add this condition - ok then we return null

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
klopfdreh: An exception is thrown if the code is executed twice with the same path (newFileSystem)

klopfdreh:
@martin-g any other hints? :smiley:

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Configures the resource response by overriding it with additional information
+	 * 
+	 * @param resourceResponse
+	 *            the resource response to be configured
+	 * 
+	 * @return the resource response to be configured
+	 */
+	protected ResourceResponse configureResourceResponse(ResourceResponse resourceResponse)
```
martin-g: I think there is no need of `#configureResourceResponse()`. 
`org.apache.wicket.request.resource.AbstractResource#setResponseHeaders

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -0,0 +1,145 @@
...
+	protected String getMimeType() throws IOException
+	{
+		String mimeType = null;
+		if (Application.exists())
+		{
+			mimeType = Application.get().getMimeType(path.getFileName().toString());
+		}
+		if (mimeType == null)
+		{
+
```
martin-g: extra empty line that could be removed

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	/**
+	 * Gets the input stream of the given path
+	 * 
+	 * @return the input stream of the given path
+	 * @throws IOException
+	 *             if there is an exception while receiving the input stream
+	 */
+	protected InputStream getInputStream() throws IOException
```
martin-g: I am not sure this should be overridable. What is the use case ?

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
```diff
@@ -0,0 +1,181 @@
...
+ * 
+ * @author Tobias Soloschenko
+ */
+public class FileSystemResourceReference extends ResourceReference
+{
+	private static final long serialVersionUID = 1L;
+
+	private final Path path;
+
+	private static Map<URI, FileSystem> fileSystemURIs = new HashMap<URI, FileSystem>();
```
martin-g: I still believe this should not be `static`, but should use Application's metadata.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
```diff
@@ -0,0 +1,181 @@
...
+	 * Creates a file system resource reference based on the given path
+	 * 
+	 * @param name
+	 *            the name of the resource reference to expose data
+	 * @param path
+	 *            the path to create the resource reference
+	 */
+	public FileSystemResourceReference(String name, Path path)
+	{
+		super(Args.notNull(name, "name"));
```
martin-g: org.apache.wicket.request.resource.ResourceReference#ResourceReference(java.lang.Class<?>, java.lang.String, java.util.Locale, java.lang.String, java.lang.String) cares about the non-null name. No need to check it here too.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
```diff
@@ -0,0 +1,181 @@
...
+
+	/**
+	 * Creates a file system resource reference based on the given path
+	 * 
+	 * @param path
+	 *            the path to create the resource reference (the name is used to expose the data)
+	 */
+	public FileSystemResourceReference(Path path)
+	{
+		super(Args.notNull(path, "path").getFileName().toString());
```
martin-g: This will fail if you have two files with the same file names, e.g. `/one/a` and `/two/a`. Both will try to use `a` as a name.
And since the name is used for the url for this resource reference we cannot use the absolute path. IMO this constructor should be removed in favor of `(public FileSystemResourceReference(String name, Path path))` above.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	/**
+	 * Override to apply a custom mime type without implementing a mime type detection
+	 * 
+	 * @return the mime type
+	 * @throws IOException
+	 *             if the mime type could'nt be read
+	 */
+	protected String getMimeType() throws IOException
```
martin-g: Is this method used anywhere by Wicket ?
I see it is used only by the unit tests. So maybe there is no need to have this method at all.

klopfdreh: This method is used to override the MimeType inline and not use either the Application nor the FileTypeDetector MimeType detection. I wrote an example in the mailing list (link).

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
klopfdreh: Good point - I am going to remove this constructor.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
klopfdreh: I am going to remove the notNull check.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
klopfdreh: Good point - forgot to use metadata here - will be done!

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
klopfdreh: I made all methods which uses Files overridable, so that it can be customized if required.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
klopfdreh: Will be done!

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
klopfdreh: Oh yes, you are right - so I am going to remove that method.

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResourceReference.java
martin-g: FileSystemResource has `resourceResponse.setContentType(getMimeType());`, so its `#getMimeType()` is used. 
But FSRR#getMimeType() is not used, so there is no need of it.

klopfdreh: Yes you are right - originally I wanted the FSRR#getMimeType() to be used for the FSR if overridden. I will reimplement it.

klopfdreh:
Hope the implementation gets closer to the final state.

martin-g:
PathResourceStream is still missing.
Give it a go! It shouldn't be something complex. Once ready with its impl see whether it could be used by FileSystemResource.

klopfdreh:
That was the point I asked for help. I don't actualy want to break the API with that changes.

martin-g:
Don't worry. There is nothing to break. It won't be used anywhere else for now.
Use FileResourceStream as inspiration. They should look almost the same.

klopfdreh:
@martin-g Would you be so kind and review the PathResourceStream class?

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+import java.nio.file.Path;
+import java.nio.file.attribute.BasicFileAttributes;
+
+import org.apache.wicket.util.file.File;
+import org.apache.wicket.util.lang.Args;
+import org.apache.wicket.util.lang.Bytes;
+import org.apache.wicket.util.time.Time;
+
+/**
+ * A PathResourceStream is an IResource implementation for paths.
```
martin-g: s/IResource/IResourceStream/

martin-g: for Java NIO paths

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+ * @author Tobias Soloschenko
+ */
+public class PathResourceStream extends AbstractResourceStream
+	implements
+		IFixedLocationResourceStream
+{
+	private static final long serialVersionUID = 1L;
+
+	/** Any associated path */
+	private Path path;
```
martin-g: final

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+			inputStream = null;
+		}
+	}
+
+	@Override
+	public String getContentType()
+	{
+		try
+		{
+			String contentType = URLConnection.getFileNameMap().getContentTypeFor(path.getFileName().toString());
```
martin-g: Instead of using `URLConnection.getFileNameMap().getContentTypeFor()` this should use `Application#getMimeType()`. The fallback below is OK.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+		{
+			throw new RuntimeException("content type of path " + path + " could not be acquired",
+				e);
+		}
+	}
+
+	/**
+	 * @return The path this resource resides in, if any.
+	 */
+	public Path getPath()
```
martin-g: This method should either be:
- removed completely 
- made final 
- used everywhere instead of using the member field `path`

I'm for option 2 (made final)

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+	{
+		return path;
+	}
+
+	@Override
+	public Time lastModifiedTime()
+	{
+		try
+		{
+			return Time.millis(Files.readAttributes(path, BasicFileAttributes.class)
```
martin-g: please extract some local variables
chained calls like this are hard to debug when something fails in the middle

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+				.lastModifiedTime()
+				.toMillis());
+		}
+		catch (IOException e)
+		{
+			throw new RuntimeException("length of path " + path + " could not be acquired", e);
```
martin-g: s/length/Modification time/

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+		if (inputStream == null)
+		{
+			try
+			{
+				inputStream = Files.newInputStream(path);
+			}
+			catch (IOException e)
+			{
+				throw new ResourceStreamNotFoundException(
+					"Inputstream of path " + path + " could not be acquired", e);
```
martin-g: Input stream

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+		{
+			String contentType = URLConnection.getFileNameMap().getContentTypeFor(path.getFileName().toString());
+			if(contentType == null){
+				contentType = Files.probeContentType(path);
+			}
+			return contentType;
+		}
+		catch (IOException e)
+		{
+			throw new RuntimeException("content type of path " + path + " could not be acquired",
```
martin-g: Capitalize 'content', for consistency.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+	@Override
+	public Bytes length()
+	{
+		try
+		{
+			return Bytes.bytes(Files.readAttributes(path, BasicFileAttributes.class).size());
+		}
+		catch (IOException e)
+		{
+			throw new RuntimeException("length of path " + path + " could not be acquired", e);
```
martin-g: Capitalize 'length', for consistency

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+
+	@Override
+	public Bytes length()
+	{
+		try
+		{
+			return Bytes.bytes(Files.readAttributes(path, BasicFileAttributes.class).size());
```
martin-g: Extract local variables for the attrs and size

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,163 @@
...
+	@Override
+	public String locationAsString()
+	{
+		return path.toString();
+	}
+
+	@Override
+	public String toString()
+	{
+		return path.toString();
```
martin-g: delegate to `#locationAsString()`. If anyone wants to customize this then only one method has to be overridden

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: This is not available in wicket-util.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: Made it final.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: done.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
```diff
@@ -0,0 +1,170 @@
...
+	@Override
+	public String getContentType()
+	{
+		try
+		{
+			String contentType = URLConnection.getFileNameMap()
+				.getContentTypeFor(path.getFileName().toString());
+			if (contentType == null)
+			{
+				contentType = Files.probeContentType(path);
```
martin-g: Maybe `Files.probeContentType(path)` should be used first? It uses Java NIO APIs.

martin-g:
No other comments from me.

wicket-util/src/main/java/org/apache/wicket/util/resource/PathResourceStream.java
klopfdreh: Good point - yep I will change it to be executed first.

klopfdreh:
Should I provide an additional FileSystemStreamReference (using path) - like org.apache.wicket.core.util.resource.locator.caching.FileResourceStreamReference?

martin-g:
Good catch!
Yes, we will need it.

klopfdreh:
Ok, FileSystemResourceStreamReference is implemented.

wicket-core/src/main/java/org/apache/wicket/core/util/resource/locator/caching/FileSystemResourceStreamReference.java
```diff
@@ -0,0 +1,29 @@
...
+import java.nio.file.Paths;
+
+import org.apache.wicket.util.resource.PathResourceStream;
+
+/**
+ * A reference which can be used to recreate {@link PathResourceStream}
+ * 
+ * @author Tobias Soloschenko
+ */
+public class FileSystemResourceStreamReference extends AbstractResourceStreamReference
```
martin-g: We should use the same name for all classes - either `FileSystem**` or `Path**`.

klopfdreh: I prefer FileSystem

martin-g: Then let's rename PathRS to FileSystemRS for consistency.

klopfdreh:
Done.

martin-g:
No further comments by me.
If no one objects then please merge it to master and 7.x.

klopfdreh:
Closed due to commits https://github.com/apache/wicket/commit/e0c2f1bbae576d26f2dc54996905e9acb63ccca0 and https://github.com/apache/wicket/commit/cdd9a6d56f05d37639f01b31091fb4bfd13ab926

PR #143, opened 2015-11-30
WICKET-6042 - Implementation of ExternalImage component

klopfdreh:


wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,274 @@
...
+	 * Creates an external image
+	 * 
+	 * @param id
+	 *            the component id
+	 * @param src
+	 *            the model source URL
+	 * @param srcSetModels
+	 *            a model list of URLs placed in the srcset attribute
+	 */
+	public ExternalImage(String id, IModel<?> src, IModel<?>... srcSetModels)
```
martin-g: Why `IModel<?>` and not `IModel<String>`?
Same for srcSet.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,274 @@
...
+	 * @param values
+	 *            the x values to be used in the srcset
+	 */
+	public void setXValues(String... values)
+	{
+		if (xValues == null)
+		{
+			xValues = new ArrayList<>();
+		}
+		xValues.clear();
```
martin-g: `.clear()` could be in an `else` clause

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,274 @@
...
+	 * @param sizes
+	 *            the sizes to be used in the size
+	 */
+	public void setSizes(String... sizes)
+	{
+		if (this.sizes == null)
+		{
+			this.sizes = new ArrayList<>();
+		}
+		this.sizes.clear();
```
martin-g: could be in `else`

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,274 @@
...
+
+	/** The sizes of the responsive images */
+	private List<String> sizes = null;
+
+	/**
+	 * Cross origin settings
+	 */
+	private Cors crossOrigin = null;
+
+	private IModel<?>[] srcSet;
```
martin-g: These models should be detached.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Mhhh - thought that if someone puts in an modelobject which provides the url with the toSting() method this way it can be used more flexible.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Ok - I improve it here and in Image / Source.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Ok - I improve it here and in Image / Source.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Thanks! Will be done! :smile:

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
martin-g: The the first constructor should use `Object`, not `String` :-)
See `Label`. It was using String, now it uses Serializable.

svenmeier: I find IModel<?>... highly unusual for a Wicket component. Shouldn't that be IModel<List<?>> ?

sebfz1: Shouldn't be `src` named `srcModel` in that case?

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,274 @@
...
+	 * @param id
+	 *            the component id
+	 * @param src
+	 *            the source URL
+	 * @param srcSet
+	 *            a list of URLs placed in the srcset attribute
+	 */
+	public ExternalImage(String id, String src, String... srcSet)
+	{
+		this(id, src != null ? Model.of(src) : null, convertToModel(srcSet));
```
sebfz1: Hummm, IMHO having a null model-object and a null model is different. 
Thus, might be Serializable instead of String if the second constructor becomes IModel<? extends Serializable>
Something like this:
```
public ExternalImage(String id, Serializable src, Serializable... srcSet)
{
    this(id, Model.of(src), convertToModel(srcSet));
or 
    this(id, Model.of(Args.notNull(src)), convertToModel(srcSet)); if we consider src is mandatory (don't think so but...)
}

File: wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java [hunk 6]
```diff
@@ -0,0 +1,274 @@
...
+	 * @param srcSetModels
+	 *            a model list of URLs placed in the srcset attribute
+	 */
+	public ExternalImage(String id, IModel<?> src, IModel<?>... srcSetModels)
+	{
+		super(id, src);
+		this.srcSet = srcSetModels;
+	}
+
+	/**
```
sebfz1: @klopfdreh, would it be possible to have a constructor like this? (for CPM usage) :p
```
public ExternalImage(String id, IModel<?>... srcSetModels)
```

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,274 @@
...
+	 *
+	 * @param tag
+	 *            the component tag
+	 * @param srcModel
+	 *            the model containing the src URL
+	 */
+	protected void buildSrcAttribute(final ComponentTag tag, IModel<?> srcModel)
+	{
+		// The first model is the one put into src attribute
+		tag.put("src", srcModel.getObject().toString());
```
sebfz1: Might be good to test `srcModel.getObject()` for null to not throw a NPE, or using String.valueOf(). I prefer the second solution but the first allows to deal with the null (log warning for instance)...

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,274 @@
...
+	 * Builds the src attribute
+	 *
+	 * @param tag
+	 *            the component tag
+	 * @param srcModel
+	 *            the model containing the src URL
+	 */
+	protected void buildSrcAttribute(final ComponentTag tag, IModel<?> srcModel)
+	{
+		// The first model is the one put into src attribute
```
sebfz1: I guess this comment is not relevant anymore...

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: I change it to Serializable and the names to srcModel. :-)

IModel<Serializable> srcModel

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Good point!

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Will ne removed.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: String.valueOf is good here!

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: I am going to have a look

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Mhhh - this is not possible if we have the ctr> id, srcModel, srcSetModels

sebfz1: Oops! Of course! :/ 
What if we change the vararg by a list of model or a listmodel ? The fist ctor could then call model.oflist... (just a suggestion...)

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Label uses IModel<?>, too, because if you have two constructor (String,Serializable) and (String, IModel<? extends Serializable> the second one can't be invoked by the first.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: I dislike list, because you have to preconstruct it. I like it much more to add the urls inline.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: But because it is a model I surround it with a nullcheck, instead.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: As described in a comment before IModel<? extends Serializable> is not possible, because of recursion, in the case we change the String and vararg String... to Serializable

klopfdreh:
Any other suggestions? I've not changed the IModel&lt;?&gt;, currently - because of the conflict between Serializable / and the IModel of Serializable. Any suggestions to solve that?

martin-g:
OK, I see.
Then make the other constructor accepting `Serializable` instead of `String`. As `Label` does :-)

svenmeier:
What about my remark? For srcset I'd prefer to have a model of list instead of an array of models.

klopfdreh:
@martin-g I already changed it. :smiley: 
@svenmeier in the current image / source implementation we also use varargs for the srcset. I think we should provide a uniform API across image / source components to not confound users. But lets discuss that. Maybe we find a way to uniform it even with the model of list

klopfdreh:
Maybe we can add the following constructor to also support your requirements:
ExternalImage(id,Model&lt;?&gt; src,Model&lt;List&lt;Serializable&gt;&gt; srcSet)

svenmeier:
I don't find it confusing:
- Image has IResource...
- ExternalImage has Serializable... *and* IModel<List<? extends Serializable>>

IModel... looks strange to me, IMHO the need to have #convertToModel() shows how strange it is to have an array of models.

martin-g:
I agree with Sven that `IModel<List>` is better than varargs.
We had issues with varargs in StringResourceModel and we broke the API in 7.x to improve it.
I also think it would be harder for the application to provide the parameters as varargs, than as List.
Maybe we should deprecate the varargs signature in the other new components ? Or at least new constructor with List and wait for some feedback in the coming months.

sebfz1:
+1 too, not fan of `IModel...`.

Then I guess it would be possible to add a `ExternalImage(String id, IModel<List<Seriarizable>> srcSetModel)` ctor ;)

Another remark I did earlier about my concern on `src != null ? Model.of(src) : null`. It means that supplying a null value will result of a null model, leading to a search of a compound model in the component hierarchy, which is confusing IMHO.

klopfdreh:
Well ok then! :smile: Two tasks:

1. Change varargs to List.
2. Apply src directly to super()

klopfdreh:
May one more task:

3. Add new constructors to Image / Source to support IModel&lt;List&lt;ResourceReference&gt;&gt; and set the varargs constructors to deprecated.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,295 @@
...
+	 * Builds the src attribute
+	 *
+	 * @param tag
+	 *            the component tag
+	 * @param srcModel
+	 *            the model containing the src URL
+	 */
+	protected void buildSrcAttribute(final ComponentTag tag, IModel<?> srcModel)
+	{
+		if (srcModel != null)
```
sebfz1: Caution, you do not test the nullity of the model object here...

sebfz1: IMO, you better have to use `#getDefaultModelObject()` because the model supplied in arg can be null, whereas the model can exists through the component hierarchy...

sebfz1: Oops, please forget my last comment, you already call it with `getDefaultModel()`...

klopfdreh:
Hi, I first changed only ExternalImage / ExternalSource - if everything is finished with that I am going to commit the changes on Image / Source.

Hope the last commit is like you want it to be.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: No problem :smiley:

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,298 @@
...
+
+	/** The sizes of the responsive images */
+	private List<String> sizes = null;
+
+	/**
+	 * Cross origin settings
+	 */
+	private Cors crossOrigin = null;
+
+	private IModel<List<Serializable>> srcSetModels;
```
sebfz1: Not sure it should ends by "models" in plural. It is only one model (containing a list)...

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,298 @@
...
+	 * Creates an external image
+	 * 
+	 * @param id
+	 *            the component id
+	 * @param src
+	 *            the source URL
+	 */
+	public ExternalImage(String id, Serializable src)
+	{
+		this(id, src, (List<Serializable>)null);
```
sebfz1: If I am not wrong, it usually considered as a best practice to supply an empty list instead of a null list. I guess you can use `Collections.emptyList()` here.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,298 @@
...
+	}
+
```
sebfz1: Still missing an important ctor to me: `public ExternalImage(String id)` :)

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,298 @@
...
+	 * @param id
+	 *            the component id
+	 * @param srcModel
+	 *            the model source URL
+	 * @param srcSetModels
+	 *            a model list of URLs placed in the srcset attribute
+	 */
+	public ExternalImage(String id, IModel<Serializable> srcModel)
+	{
+		this(id, srcModel, (IModel<List<Serializable>>)null);
```
sebfz1: Idem about `null` list

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
sebfz1: The first statement remains valid :) `srcModel.getObject()` is still not tested for nullity...

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,298 @@
...
+	 *            the component tag
+	 * @param srcSetModels
+	 *            the models containing the src set URLs
+	 */
+	protected void buildSrcSetAttribute(final ComponentTag tag,
+		IModel<List<Serializable>> srcSetModels)
+	{
+		int srcSetPosition = 0;
+		List<Serializable> srcSetItems = srcSetModels.getObject();
+		if (srcSetItems != null)
```
sebfz1: Can be removed if you supply an empty list in the ctor

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,298 @@
...
+	{
+		this.crossOrigin = crossOrigin;
+	}
+
+	/**
+	 * Gets a list of models containing the src set values
+	 * 
+	 * @return a list of models containing the src set values
+	 */
+	public IModel<List<Serializable>> getSrcSet()
```
sebfz1: IMO: either it is named `#getSrcSetModel` and it returns the model, or it is named `#getSrcSet` and it returns the underlying model object :). The first one sound more logical...

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
sebfz1: Sorry Tobias, I just realized I have previously asked for a `ExternalImage(String id, IModel<List<Seriarizable>> srcSetModel)` ctor without mentioning the other one: `public ExternalImage(String id)`...

sebfz1: Actually, it is not possible to have both mentioned ctor because os the same erasure.  
So to sum-up, the list of ctors I see:

```java
public ExternalImage(String id)
{
	this(id, null, Model.ofList(Collections.emptyList()));
}

public ExternalImage(String id, Serializable src)
{
	this(id, Model.of(src), Model.ofList(Collections.emptyList()));
}

public ExternalImage(String id, Serializable src, List<Serializable> srcSet)
{
	this(id, Model.of(src), Model.ofList(srcSet));
}

public ExternalImage(String id, IModel<Serializable> srcModel)
{
	this(id, srcModel, Model.ofList(Collections.emptyList()));
}

public ExternalImage(String id, IModel<Serializable> srcModel, IModel<List<Serializable>> srcSetModel)
{
	super(id, srcModel);
	
	this.srcSetModels = srcSetModel;
}
```

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: I change it to singular.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: k :wink:

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: done.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: done.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: done.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: renamed to getSrcSetModel()

klopfdreh:
Only for comparison

```java
ExternalImage externalImage2 = new ExternalImage("externalImage2",
"http://wicket.apache.org/img/wicket-7-bg.jpg",
Arrays.<Serializable>asList(
"http://wicket.apache.org/img/wicket-7-bg-1.jpg",
"http://wicket.apache.org/img/wicket-7-bg-2.jpg"));
```

```java
ExternalImage externalImage2 = new ExternalImage("externalImage2",
"http://wicket.apache.org/img/wicket-7-bg.jpg",
"http://wicket.apache.org/img/wicket-7-bg-1.jpg",
"http://wicket.apache.org/img/wicket-7-bg-2.jpg");
```

klopfdreh:
Thank you for the time you spend to review the code! :+1: Any other things to be changed? 

Anyway - you have convinced me about the list arg over the var arg - the additional code is rather small and the whole component is type safe, now.

sebfz1:
Hi Tobias,

> Thank you for the time you spend to review the code!

You are welcome, thanks a lot for your time too! :)

> Any other things to be changed? 

Well... yes ! :) You have removed the null check on the *model*  in `#buildSrcAttribute`, but there is still no null check on the model *object*. Sorry if I had not been clear enough.
As I wrote previously, I see 2 options to handle this:

1/ Check null *object*, and do not render the src tag. This will lead the image to not show at all (well, at least FF works like that)

```java
protected void buildSrcAttribute(final ComponentTag tag, IModel<?> srcModel)
{
	if (srcModel.getObject() != null)
	{
		tag.put("src", srcModel.getObject().toString());
	}
}
```

2/ use `String#valueOf`. This will lead to the src tag to have null value and display a broken image, which can also make sense to alert without throwing a runtime NPE...

```java
protected void buildSrcAttribute(final ComponentTag tag, IModel<?> srcModel)
{
	tag.put("src", String.valueOf(srcModel.getObject()));
}
```

I have a slight preference for the 2nd choice. I guess you will opt for the first choice; so if everyone is fine with this first choice, that's fine for me too! (@svenmeier , @martin-g)

Thanks again,
Sebastien

klopfdreh:
I'm going to implement the second solution, but with a check if the srcModel is not null - what do you think?

sebfz1:
I am not sure because:
- having a *non*-null *model* is the responsibility of the user.
- handling null model *object* is our responsibility because it happens at runtime

If I'm not wrong, Wicket usually let throw a NPE is thrown in case of null *model*...

klopfdreh:
Yep you are right! I changed it the way to only check the model object.

klopfdreh:
Anything else? Next step would be - squash, integrate and cherrypick onto 7.x branch

sebfz1:
To me that's all fine, thank you very much Tobias! :)

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
```diff
@@ -0,0 +1,317 @@
...
+		{
+			tag.put("sizes", sizes);
+		}
+	}
+
+	/**
+	 * @param values
+	 *            the x values to be used in the srcset
+	 */
+	public void setXValues(String... values)
```
martin-g: Should we support something like `removeXValues()`? E.g. with `setXValues(null)`.
I guess no, but I have to ask.
Same for `#setSizes()`.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalSource.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * A component which displays external images within a picture tag.
+ * 
+ * @see org.apache.wicket.markup.html.image.Source
+ * 
+ * @author Tobias Soloschenko
+ * @author Sebastien Briquet
+ * @author Sven Meier
+ * @author Martin Grigorov
```
martin-g: I did nothing but "complain" here :-)

wicket-core/src/main/java/org/apache/wicket/markup/html/image/Image.java
```diff
@@ -316,8 +316,9 @@ public void setXValues(String... values)
 		if (xValues == null)
 		{
 			xValues = new ArrayList<>();
+		}else{
```
martin-g: Empty space after `{` and `else` should be on new line. Is Eclipse formatter broken or what ? ;-)

wicket-core/src/main/java/org/apache/wicket/markup/html/image/Image.java
```diff
@@ -330,8 +331,9 @@ public void setSizes(String... sizes)
 		if (this.sizes == null)
 		{
 			this.sizes = new ArrayList<>();
+		}else{
```
martin-g: Formatting

wicket-core/src/test/java/org/apache/wicket/markup/html/image/ExternalImageTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Test cases to test the external image components
+ * 
+ * @author Tobias Soloschenko
+ *
+ */
+public class ExternalImageTest extends WicketTestCase
```
martin-g: Would be good to have a test verifying that CompoundPropertyModel actually works for ExternalImage!

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalImage.java
klopfdreh: Good idea - looks much better then setXValues(null) - I also improve Image / Source with that call.

wicket-core/src/main/java/org/apache/wicket/markup/html/image/ExternalSource.java
klopfdreh: yep but even complaining is improving and a part of development. ;-)

wicket-core/src/main/java/org/apache/wicket/markup/html/image/Image.java
klopfdreh: It is the Wicket-Formatter :-b

wicket-core/src/main/java/org/apache/wicket/markup/html/image/Image.java
klopfdreh: Wicket-Formatter :-b

wicket-core/src/test/java/org/apache/wicket/markup/html/image/ExternalImageTest.java
klopfdreh: Good point!

klopfdreh:
Other things?

klopfdreh:
Well then I'm going to integrate it! :+1:

klopfdreh:
*Integrated*

Wicket 8.x:
https://github.com/apache/wicket/commit/ee572d0ee597a357f10735cb2141b01ec19d19a4
https://github.com/apache/wicket/commit/b174c21c42220d9ccb4a01579ee9b0723a121513
https://github.com/apache/wicket/commit/4eacb2aba5f580fbbb851293834253ec7a8b3ce9
https://github.com/apache/wicket/commit/a1b19493da671dce33f258de8021713fdc45234a
https://github.com/apache/wicket/commit/93fe70d08cf255d6673243ae02e36c01c360eaf7
https://github.com/apache/wicket/commit/1ac89ee1c8c305097e34c3733c41bb217d627e57
https://github.com/apache/wicket/commit/78802c9c8a13a71000d5b6225daab80c9e6f8d46

Wicket 7.x:
https://github.com/apache/wicket/commit/2e82b8913f3cbf8d2631f8afbaba46dda8b122b0

PR #147, opened 2015-12-14
Exclude option for CssUrlReplacer

klopfdreh:
Code review required - also if a back port is available. (to wicket-7.x)

https://issues.apache.org/jira/browse/WICKET-6051

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacer.java
+++ b/wicket-core/src/main/java/org/apache/wicket/resource/CssUrlReplacer.java
@@ -16,6 +16,8 @@
  */
 package org.apache.wicket.resource;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


martin-g:
The changes look good!
But isn't it better to use includes instead?
It is much easier to control what you want in then what you want out. For example when I add some third party library I'll have to investigate what CSS resources there are and exclude them.
It'd be much easier to list only the ones I'd like to process.

klopfdreh:
Okay - I will turn around that logic.

klopfdreh:
If I turn around that logic I will break the behavior, because you have to provide those settings and if users upgrading to the new wicket version and don't have set those css file names the functionality is broken.

klopfdreh:
The performance increased way better with the new regex. We should discuss changes in the jira ticket. I made a suggestion there.

klopfdreh:
Master: https://github.com/apache/wicket/commit/274696a92128129a571b53014369d4527e27d9dc
Wicket-7.x: https://github.com/apache/wicket/commit/9b8579f078335f84357b8247401a081c9f5dc9a3
Wicket-6.x: https://github.com/apache/wicket/commit/9f6c69f8f4ccf9fc21bf42092fc88f16e35f1535

PR #149, opened 2015-12-23
Fixed calculation of byte ranges

klopfdreh:
To test the correct behavior use curl:

curl --range 0-99 &lt;pathToMovie.mp4&gt; -o /dev/null

as of the developer page of Apple this should return 100 bytes, now:

https://developer.apple.com/library/safari/documentation/AppleApplications/Reference/SafariWebContent/CreatingVideoforSafarioniPhone/CreatingVideoforSafarioniPhone.html#//apple_ref/doc/uid/TP40006514-SW6


klopfdreh:
Fixed: https://github.com/apache/wicket/commit/a169eb1125d93867c64377a8dc4066491ef510c7

PR #150, opened 2015-12-23
Fixed calculation of byte ranges 7.x

klopfdreh:
To test the correct behavior use curl:

curl --range 0-99 &lt;pathToMovie.mp4&gt; -o /dev/null

as of the developer page of Apple this should return 100 bytes, now:

https://developer.apple.com/library/safari/documentation/AppleApplications/Reference/SafariWebContent/CreatingVideoforSafarioniPhone/CreatingVideoforSafarioniPhone.html#//apple_ref/doc/uid/TP40006514-SW6


klopfdreh:
Fixed: https://github.com/apache/wicket/commit/b5ee0cbf5a1234524a57983c6bfacb2fdf8aeb97

PR #151, opened 2015-12-23
WICKET-6055 Made AjaxLazyLoadPanel non-blocking

dashorst:
When having multiple AjaxLazyLoadPanels on your page, they all block
their Wicket request thread until the content is ready to load. This
can be problematic when you try to wait for some background job to
finish and want to poll for that job to be ready, and only then update
the contents.

The improvement would be to add a method that gives the developer the
option to not update just yet (isReadyForReplacement) and when it
returns true, start the replacement. By default this would return true,
implementing the current behavior of the AjaxLazyLoadPanel.

Furthermore to improve the responsiveness of the ALLP it should add a
single timer to the page that can be used by multiple ALLPs to update
themselves. The timer would poll each second and the ALLPs would use
Wicket's event bus to update themselves. With some reference counting,
the timer can remove itself from the page when all ALLPs have updated
themselves.

This enables refreshing the page as well when outside an AJAX context,
or having a user be impatient and pressing F5.

Fixes WICKET-6055

### DIFF
--- a/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/LazyLoadingPage.java
+++ b/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/LazyLoadingPage.java
@@ -100,9 +100,7 @@ public class LazyLoadingPage extends BasePage
 				{
 					try
 					{
-						System.out.println("Starting sleep");
 						Thread.sleep(seconds * 1000);
-						System.out.println("Slept " + seconds + " seconds");
 					}
 					catch (InterruptedException e)
 					{
--- a/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/LazyLoadingPage.java
+++ b/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/LazyLoadingPage.java
@@ -16,38 +16,100 @@
  */
 package org.apache.wicket.examples.ajax.builtin;
 
+import java.util.Random;
+
 import org.apache.wicket.Component;
 import org.apache.wicket.extensions.ajax.markup.html.AjaxLazyLoadPanel;
 import org.apache.wicket.markup.html.basic.Label;
+import org.apache.wicket.markup.html.link.Link;
+import org.apache.wicket.markup.html.panel.EmptyPanel;
+import org.apache.wicket.markup.repeater.RepeatingView;
+import org.apache.wicket.util.time.Duration;
 
-/**
- * @author jcompagner
- */
+@SuppressWarnings({ "javadoc", "serial" })
 public class LazyLoadingPage extends BasePage
 {
-	/**
-	 * Construct.
-	 */
+	private Random r = new Random();
+
 	public LazyLoadingPage()
 	{
-		add(new AjaxLazyLoadPanel("lazy")
+		add(new Link<Void>("startNonblocking")
 		{
-
 			@Override
-			public Component getLazyLoadComponent(String id)
+			public void onClick()
 			{
-				// sleep for 5 seconds to show the behavior
-				try
-				{
-					Thread.sleep(5000);
-				}
-				catch (InterruptedException e)
-				{
-					throw new RuntimeException(e);
-				}
-				return new Label(id, "Lazy Loaded after 5 seconds");
+				addNonBlockingPanels();
 			}
-
 		});
+		add(new Link<Void>("startBlocking")
+		{
+			@Override
+			public void onClick()
+			{
+				addBlockingPanels();
+			}
+		});
+
+		add(new EmptyPanel("lazy"));
+		add(new EmptyPanel("lazy2"));
+	}
+
+	private void addNonBlockingPanels()
+	{
+		RepeatingView rv;
+		addOrReplace(rv = new RepeatingView("lazy"));
+
+		for (int i = 0; i < 10; i++)
+			rv.add(new AjaxLazyLoadPanel(rv.newChildId())
+			{
+				private static final long serialVersionUID = 1L;
+
+				private long startTime = System.currentTimeMillis();
+
+				private int seconds = r.nextInt(10);
+
+				@Override
+				protected boolean isReadyForReplacement()
+				{
+					return Duration.milliseconds(System.currentTimeMillis() - startTime)
+						.seconds() > seconds;
+				}
+
+				@Override
+				public Component getLazyLoadComponent(String id)
+				{
+					return new Label(id, "Lazy Loaded after " + seconds + " seconds");
+				}
+			});
+	}
+
+	private void addBlockingPanels()
+	{
+		RepeatingView rv;
+		addOrReplace(rv = new RepeatingView("lazy2"));
+
+		for (int i = 0; i < 10; i++)
+			rv.add(new AjaxLazyLoadPanel(rv.newChildId())
+			{
+				private static final long serialVersionUID = 1L;
+
+				private int seconds = r.nextInt(10);
+
+				@Override
+				public Component getLazyLoadComponent(String markupId)
+				{
+					try
+					{
+						System.out.println("Starting sleep");
+						Thread.sleep(Duration.seconds(seconds).getMilliseconds());
+						System.out.println("Slept " + seconds + " seconds");
+					}
+					catch (InterruptedException e)
+					{
+					}
+					return new Label(markupId,
+						"Lazy loaded after blocking the Wicket thread for " + seconds + " seconds");
+				}
+			});
 	}
 }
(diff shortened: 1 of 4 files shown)


wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
```diff
@@ -187,8 +237,56 @@ public Component getLoadingComponent(final String markupId)
...
+	 *            The components markupid.
+	 * @return The component that must be lazy created.
+	 */
+	public abstract Component getLazyLoadComponent(String markupId);
+
+	/**
+	 * The AJAX timer for updating the AjaxLazyLoadPanel. Is designed to be a page-local singleton
+	 * keeping track of multiple LazyLoadPanels using reference counting.
+	 */
+	private class AjaxLazyLoadTimer extends AbstractAjaxTimerBehavior
```
svenmeier: This should probably be static to prevent a dangling reference to the surrounding component.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
```diff
@@ -187,8 +237,56 @@ public Component getLoadingComponent(final String markupId)
...
+	{
+		private static final long serialVersionUID = 1L;
+
+		private int lazyLoadPanels = 0;
```
svenmeier: Counting all loading panels might be problematic: What if one (or one of its parents) is removed from the component hierarchy? The timer will keep requesting until forever.
How about broadcasting the AjaxLazyLoadTimer to all components, giving them the opportunity to keep the timer running?

    ((AjaxLazyLoadTimer)event.getPayload()).notReadyForReplacement();

If any of the panels is not ready yet, the timer has to keep running.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
dashorst: It would not pose a problem and make the code simpler. With static I have to get the component's page and ran into a order problem of removing the behavior from the page and trying to change the page's meta data. The dangling reference is only a problem when you have memory constraints, but then the lazyloadpanel's reference would not be your biggest problem. Regardless I've changed it to be static.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
dashorst: Fixed it by tracking the panels directly.

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/LazyLoadingPage.java
```diff
@@ -16,38 +16,100 @@
...
+				private static final long serialVersionUID = 1L;
+
+				private int seconds = r.nextInt(5);
+
+				@Override
+				public Component getLazyLoadComponent(String markupId)
+				{
+					try
+					{
+						System.out.println("Starting sleep");
```
martin-g: Use Logger instead.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
```diff
@@ -139,56 +182,176 @@ protected void handleCallbackScript(final IHeaderResponse response,
...
+	 */
+	private static class AjaxLazyLoadTimer extends AbstractAjaxTimerBehavior
+	{
+		private static final long serialVersionUID = 1L;
+
+		private final List<AjaxLazyLoadPanel> lazyLoadPanels = new ArrayList<>();
+
+		public AjaxLazyLoadTimer()
+		{
+			super(Duration.ONE_SECOND);
```
martin-g: +1 to make the duration configurable, e.g. overridable method in AjaxLazyLoadPanel.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
```diff
@@ -74,63 +106,74 @@ public AjaxLazyLoadPanel(final String id, final IModel<?> model)
...
+	{
+		super.onInitialize();
+
+		// when the timer is not yet installed add it
+		AjaxLazyLoadTimer timer = getLazyLoadTimer();
+		if (timer == null)
+		{
+			timer = new AjaxLazyLoadTimer();
+			setLazyLoadTimer(timer);
+			getPage().add(timer);
```
martin-g: I think there is no need to keep two references to the timer in the page.
You can use `page.getBehaviors(AjaxLazyLoadTimer.class).get(0)` to get the reference. No need of meta data.

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/LazyLoadingPage.java
dashorst: oopsie, this one should've been removed.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
dashorst: Can't. The timer is a single instance for all ALLP's on a page. The reason for using just one timer is that when you have more, you get a barrage of events fired at your server and you don't want that (I've seen the horror). A single timer is the best compromise for this. I've thought about increasing the timer the longer the wait takes, but decided against it because as a user you want direct feedback. A timer of a second is a rather nice compromise between request pressure on the server and wait time for the user.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
```diff
@@ -139,56 +182,176 @@ protected void handleCallbackScript(final IHeaderResponse response,
...
 		}
 		super.onBeforeRender();
 	}
 
+	@Override
+	public void onEvent(IEvent<?> event)
+	{
+		super.onEvent(event);
+
+		if (state == LoadingState.REPLACING_LAZY_COMPONENT)
```
svenmeier: Shouldn't #onEvent() check for a specific event? If other events are sent through the component hierarchy (e.g. when any other Ajax request is processed), the lazy component might be replaced prematurely.

martin-g:
@dashorst What is the status here? Can I help somehow ?

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
dashorst: The `ALLT` is per-page. Perhaps we can check if the Page implements a special settings interface e.g. `IAjaxLazyLoadTimerSettings` and retrieve the duration from that?

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
dashorst: No, as `isReadyForReplacement()` is actually there for preventing replacement prematurely. Why would you wait until some time in the future to do the replacement when you can replace now and the component is ready for replacement?

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxLazyLoadPanel.java
Jezza: Yeah, I was just about to say, why not delegate it to the page.
That seems like a nice compromise between making sure we don't flood the server, and we can control the polling time.

Jezza: Or go one step futher, and make it a parent interface.
So, you just call findParent(IAjaxLazyLoadTimerSettings.class), and that way we have even more control over it, and it will still work when attached only to a page.

martin-g:
@dashorst Ping!

dashorst:
Any comments about configuring the duration of the polling timer? I'm inclined to just defer that to a specific user request rather than implement it directly. Perhaps someone comes up with a better solution to configure the duration than what we have now. I'd rather wait for the need and a good solution than implement a solution that paints us in the corner.

svenmeier:
I've wrapped my ideas about this topic in #240

PR #155, opened 2016-01-06
NashornResourceReference

klopfdreh:
Allows to post javascript to a mounted resource reference, executes the code on server side and returns the result back to the client.


klopfdreh:
Will be published to https://github.com/wicketstuff/core

PR #156, opened 2016-01-12
Error page redirection support for autocomplete ajax response

iamrakesh:
Autocomplete field resulting an exception with 
org.apache.wicket.settings.ExceptionSettings#errorHandlingStrategyDuringAjaxRequests, 
set to 
org.apache.wicket.settings.ExceptionSettings.AjaxErrorStrategy.REDIRECT_TO_ERROR_PAGE

does not redirect to the error page.

This fix will enable page redirection even for autocomplete field.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -704,6 +704,7 @@
 					if (attrs.wr) {
 						self.processAjaxResponse(data, textStatus, jqXHR, context);
 					} else {
+						self.handleRedirection(jqXHR, context);
 						self._executeHandlers(attrs.sh, attrs, jqXHR, data, textStatus);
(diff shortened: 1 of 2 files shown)


iamrakesh:
Not sure about JavaScript test added, as I see all tests (part of all.html) go green always.

martin-g:
wicket-core/src/test/js/** cover only the .js files in -core.
There are no *unit* JS tests for -extensions module.
In https://github.com/apache/wicket/tree/master/wicket-examples/src/main/webapp/js-test there are *functional* tests though for some of the examples. There is no such for the autocomplete example yet.

martin-g:
Please create a ticket in JIRA with a quickstart for the problem in autocomplete.
I just checked the diff and I see that you actually modified wicket-ajax-jquery.js, not wicket-autocomplete.js. Changes here need more attention before applying!

iamrakesh:
Yes, changes to wicket-ajax-jquery.js need more attention, created a ticket for it with description
https://issues.apache.org/jira/browse/WICKET-6075

solomax:
I guess it can be dropped? :)

PR #162, opened 2016-03-07
User Guide - Wicket forms - CompoundPropertyModel - sample java code issues

amilor:
The java code example used two variables with identical name and of different type, not only this would not compile, but the used name could potentially steer wicket beginners in wrong direction of thinking (regarding the matter of topic in this code snippet).


martin-g:
Thank you!

PR #166, opened 2016-03-23
WICKET-6128 / WICKET-6127 - Request url / Session count measurement

klopfdreh:


wicket-experimental/wicket-metrics/src/main/java/org/apache/wicket/metrics/aspects/WicketFilterRequestCycleAspect.java
```diff
@@ -42,6 +44,35 @@
...
+	{
+		Object[] args = joinPoint.getArgs();
+		if (args.length >= 3)
+		{
+			Object requestAsObject = args[2];
+			if(requestAsObject instanceof HttpServletRequest){
+				HttpServletRequest httpServletRequest = (HttpServletRequest)requestAsObject;
+				if (httpServletRequest != null)
+				{
+					return measureTime("core/application/request/" + httpServletRequest.getRequestURL().toString().replace("/", "_").replace(".", "_").replace(":", ""), joinPoint,
```
martin-g: Split this in several lines/assignments.
Use `String#replace(char, char)` where possible. 
Otherwise prefer Wicket-Util's `Strings#replace(CharSequence, CharSequence, CharSequence)` than `String#replace(CharSequence, CharSequence)` because the latter uses non-compiled Regex,
i.e. it is much slower.

martin-g:
:+1:

klopfdreh:
Hi Martin,

thanks for reviewing. I am going to convert the replace to char only - are you able to test the metrics within your project? I did already and they were working for me.

I thought that a HttpSessionListener is the best choice for the session count - but if metadata-complete is set to true you have to configure the listener yourself.

The fallback from Application.get() > Application.get(String) > static members are completely configurable.

klopfdreh:
Integrated:

master: https://github.com/apache/wicket/commit/50b6cadc5d44a45c261f599501cac20b5c20c766
wicket-7.x: https://github.com/apache/wicket/commit/20c3b9982dffbb265bb176439414622169b1533a

@martin-g - One thing which has to be done so that the measurement of the session is working correctly is to clear the session storage from disc after restarting the server, because otherwise stored session leads to negative session counts, because they are restored and expire.

There is no good solution to handle this out of the box by the servlet API, because anywhere you persist the information about a session count it can be erased. (tmp folder, realPath(), etc.)

I made a hint into the guide. I also removed the jsessionid from the path of the request measurement.

PR #167, opened 2016-04-19
Enable DeltaManager to replicate PageTable in Sessions

nilsschmidthamburg:
- When using Tomcats o.a.c.h.t.SimpleTCPCluster to replicate Sessions
   in a Tomcat cluster the o.a.c.h.s.DeltaSession is the default
   implementation which handles session replication
 - Unfortunately, to calculate its deltas, it assumes that session
   attributes are only modified by calling accessor methods (i.e.
   DeltaSession.setAttribute(n,v))
 - Modifying an existing Object (e.g. wicket's PageTable) in place of
   the current session, would not trigger a delta replication
 - To make Wicket work with SimpleTCPCluster and DeltaSessions, we will
   always rewrite the PageTable into Session by calling setAttribute


martin-g:
Thanks!
I've created https://issues.apache.org/jira/browse/WICKET-6145 for the changelog.
I'll merge it soon!

martin-g:
Thanks for the pull request!
I have seen such behavior in Jetty as well, so this should help there too!

PR #169, opened 2016-04-29
now importing the correct cglib, so that wicket will play along with …

nmwael:
…pax-wicket


martin-g:
wicket-core doesn't use CGLIB!
wicket-ioc does use CGLIB.
I'm afraid your PR is invalid.

nmwael:
seems the dynamic cglib import are comming from pax wicket.. SO I will fix it there, where it should be fixed!

PR #170, opened 2016-05-10
Port of AJAX stateless artifacts from wicketstuff stateless module

bitstorm:
Port of AJAX stateless artifacts from wicketstuff stateless module

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
@@ -81,7 +81,15 @@ public abstract class AbstractDefaultAjaxBehavior extends AbstractAjaxBehavior
 	@Override
 	protected void onBind()
 	{
-		getComponent().setOutputMarkupId(true);
+		final Component component = getComponent();
+		
+		component.setOutputMarkupId(true);
+		
+		if (getStatelessHint(component))
+		{
+			//generate behavior id
+	        component.getBehaviorId(this);
+		}
 	}
 
 	/**
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -3056,6 +3056,8 @@ public abstract class Component
 	 * @param object
 	 *            The object to set
 	 * @return This
+	 * @throws IllegalStateException If the component has neither its own model nor any of its
+	 * parents uses {@link IComponentInheritedModel}
 	 */
 	@SuppressWarnings("unchecked")
 	public final Component setDefaultModelObject(final Object object)
--- a/wicket-core/src/main/java/org/apache/wicket/Session.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Session.java
@@ -488,7 +488,7 @@ public abstract class Session implements IClusterable, IEventSink
 	 */
 	private void destroy()
 	{
-		if (sessionStore != null)
+		if (getSessionStore() != null)
 		{
 			sessionStore.invalidate(RequestCycle.get().getRequest());
 			sessionStore = null;
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/form/AjaxFormSubmitBehavior.java
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/form/AjaxFormSubmitBehavior.java
@@ -168,7 +168,7 @@ public abstract class AjaxFormSubmitBehavior extends AjaxEventBehavior
 	@Override
 	protected void onEvent(final AjaxRequestTarget target)
 	{
-		getForm().getRootForm().onFormSubmitted(new AjaxFormSubmitter(this, target));
+		getForm().getRootForm().onFormSubmitted(new AjaxFormSubmitBehavior.AjaxFormSubmitter(this, target));
 	}
 
 	/**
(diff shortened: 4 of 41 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/StatelessOnChangeAjaxBehavior.java
```diff
@@ -0,0 +1,76 @@
...
+ * This behavior uses best available method to track changes on different types of form components.
+ * To accomplish this for text input form components it uses a custom event, named 'inputchange',
+ * that is handled in the best way for the specific browser. For other form component types the
+ * 'change' event is used.
+ * </p>
+ * 
+ * @author Andrea Del Bene
+ *
+ */
+public class StatelessOnChangeAjaxBehavior extends StatelessAjaxFormComponentUpdatingBehavior
```
martin-g: Except the parent class this class is an exact copy of `OnChangeAjaxBehavior`.
Can't we extract the common code somewhere and reuse it ?

wicket-core/src/main/java/org/apache/wicket/ajax/form/StatelessAjaxFormComponentUpdatingBehavior.java
```diff
@@ -0,0 +1,74 @@
...
+    {
+        final Url url = Url.parse(super.getCallbackUrl().toString());
+        final PageParameters params = getPageParameters();
+
+        return UrlUtils.mergeParameters(url, params).toString();
+    }
+
+    protected PageParameters getPageParameters()
+    {
+    	return null;
```
martin-g: Why not `getComponent().getPageParameters()` ?

wicket-core/src/main/java/org/apache/wicket/ajax/form/StatelessAjaxFormComponentUpdatingBehavior.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public CharSequence getCallbackUrl()
+    {
+        final Url url = Url.parse(super.getCallbackUrl().toString());
+        final PageParameters params = getPageParameters();
+
+        return UrlUtils.mergeParameters(url, params).toString();
+    }
```
martin-g: `#onBind()` and `#getCallbackUrl()` are the same as in StatelessAjaxEventBehavior.
Maybe we could introduce an interface with default methods that is mixed into those classes.

wicket-core/src/main/java/org/apache/wicket/ajax/form/StatelessAjaxFormSubmitBehavior.java
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * Stateless version of {@link AjaxFormSubmitBehavior}.
+ */
+public class StatelessAjaxFormSubmitBehavior extends StatelessAjaxEventBehavior
+{
+
+	/**
+	 * 
+	 */
+	private static final long serialVersionUID = -4052276703516669553L;
```
martin-g: Useless javadoc.

martin-g:
Overall it is OK, but I don't like that there is so much duplicated code.

bitstorm:
Hi @martin-g and thank you for your reviw,

as you noted the code contains some classes where code is repeated. I've tried to reduce code duplication but without using Java 8 is nearly impossible. As you mentioned above default methods might help a lot. I'm considering to post-pone this integration to Wicket 8 where I'm free to use Java 8.

martin-g:
Oh. I didn't pay attention to the destination branch. It is 'wicket-7.x', I thought it is 'master'.

bitstorm:
@martin-g  using inheritance instead of a separate stateless hierarchy reduced the code a lot :-)

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
```diff
@@ -81,7 +82,27 @@
...
+		}
+	}
+	
+	@Override
+	public CharSequence getCallbackUrl()
+	{
+		if (getStatelessHint(getComponent()))
+		{
+		    final Url url = Url.parse(super.getCallbackUrl().toString());
+	        return UrlUtils.mergeParameters(url).toString();
```
svenmeier: #mergeParameters() does nothing when no parameters are passed - are the page parameters missing here?

bitstorm: yes you are right, I haven't finished yet the refactoring. We can simplify even more moving away getCallbackUrl from AbstractDefaultAjaxBehavior and using it with page parameters only in StatelessAjaxEventBehavior.

bitstorm: @svenmeier yes you are just right. We should pass page parameters here.

svenmeier:
Actually I'd fine with adding the stateless-handling into AbstractDefaultAjaxBehavior as proposed here.

But I'd rather not add all these Stateless* classes - IMHO it's fine to let users override #isStateless() when they want to support stateless requests.

bitstorm:
BTW I'm starting to wonder if we really need a custom AbstractDefaultAjaxBehavior#getCallbackUrl() to add page parameters to the callback url. When we generate an URL for stateless pages we end up using  AbstractBookmarkableMapper or MountedMapper and both uses page.getPageParameters() to build the URL (line 182 for MountedMapper and 476 for AbstractBookmarkableMapper )

bitstorm:
@svenmeier  I also agree that the best way to proceed now would be to commit just the code to make stateless AJAX possible, i.e. to commit just the code involving the existing AJAX components/behaviors. 
If everybody agrees I will close this PR without merging anything and I will open an issue on JIRA to commit just the code we need.

martin-g:
I agree!

PR #171, opened 2016-06-01
modified class resolution to account for OSGi classloader handling

cnoelle:
The commit fixes a ClassNotFoundException that occurs when using Wicket with the Felix OSGi framework and Pax Wicket (https://github.com/ops4j/org.ops4j.pax.wicket) as OSGi-wrapper.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/core/util/lang/WicketObjects.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/util/lang/WicketObjects.java
@@ -83,7 +83,12 @@ public class WicketObjects
 		}
 		catch (ClassNotFoundException cnfx)
 		{
-			log.warn("Could not resolve class [" + className + "]", cnfx);
+			try 
+			{
+				resolved = (Class<T>)Class.forName(className);
+			} catch (ClassNotFoundException cnfx2) {
+				log.warn("Could not resolve class [" + className + "]", cnfx);
+			}
 		}
 		return resolved;
 	}


martin-g:
Hi,

Thank you for the Pull Request!
Could you please provide more information why this change is needed ?
For OSGi environment the application is supposed to use custom IClassResolver.

cnoelle:
Hi,

I tried using a custom class resolver (one provided by Pax Wicket), but it doesn't help.
Below is the stack trace that causes the problem... there is no IClassResolver involved, instead the class is loaded via WicketObject.resolveClass() (called by WicketObject.newInstance() in the stack trace). I suspect this is due to the fact that the Application#initializeComponents() is called before the Application#init() method, where one would normally register the class loader? 

Caused by: org.apache.wicket.WicketRuntimeException: Unable to create org.ops4j.pax.wicket.internal.Initializer
	at org.apache.wicket.core.util.lang.WicketObjects.newInstance(WicketObjects.java:385)
	at org.apache.wicket.Application.addInitializer(Application.java:587)
	at org.apache.wicket.Application.load(Application.java:633)
	at org.apache.wicket.Application.initializeComponents(Application.java:511)
	at org.apache.wicket.Application.initApplication(Application.java:829)
	at org.apache.wicket.protocol.http.WicketFilter.init(WicketFilter.java:427)

martin-g:
I see.
There was a discussion recently about this at dev@wicket.apache.org.
You have to use `Application#internalInit()` instead of `#init()` in that case.

cnoelle:
Ok, that does solve it. Thanks a lot!

PR #173, opened 2016-07-08
Bugfix: Ajax add header mangling

bollinger:
When an ajax update adds a new header css link, the client side javascript incorrectly adds attributes "id" and "type".

expected header
```
<head>
  <link href="status.css" rel="stylesheet">
</head>
```

before fix, the header is 
```
<head>
  <link type="null" href="status.css" rel="stylesheet" id="null">
</head>
```

The fix is to only add the attributes supplied by the server.

see https://github.com/bollinger/wicket-style-bug for a quick start which demonstrates the problem.


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -2117,11 +2117,11 @@
...
-						// copy required attributes
-						css.id = node.getAttribute("id");
-						css.rel = node.getAttribute("rel");
-						css.href = node.getAttribute("href");
-						css.type = node.getAttribute("type");
+						// copy supplied attributes only.
+						for (var attrIdx = 0; attrIdx < node.attributes.length; attrIdx++) {
+							var attr = node.attributes[attrIdx];
+							css.setAttribute(attr.name, attr.value)
+						}
```
martin-g: Let's use jQuery for that.
See http://stackoverflow.com/a/14645827/497381 for a way to list the attributes and then use `$.attr()` to set it.

bollinger: What is the advantage of using jQuery over direct dom manipulation? 
IE 5.5 is dead so using element.attributes is fine?
https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes

martin-g: There must be a reason why jQuery 3.1.0 still has this method.

martin-g:
Thank you, @bollinger !

PR #176, opened 2016-08-16
Expensive validator wicket 8.x

amichalowski:
We need to call database stored procedure in validator.
It is quite expensive and it shouldn't be called after every key press.

I would like to propose a little validation improvement.


martin-g:
X-Ref: https://github.com/apache/wicket/pull/177

martin-g:
@amichalowski Please close this PR. We don't have permissions to close it ourselves. Thanks!

PR #177, opened 2016-08-16
Expensive validator wicket 7.x

amichalowski:
We need to call database stored procedure in validator.
It is quite expensive and it shouldn't be called after every key press.

I would like to propose a little validation improvement.

Similar pull request but different branch:
[https://github.com/apache/wicket/pull/176](https://github.com/apache/wicket/pull/176)

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/form/AjaxFormComponentUpdatingBehavior.java
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/form/AjaxFormComponentUpdatingBehavior.java
@@ -142,7 +142,7 @@ public abstract class AjaxFormComponentUpdatingBehavior extends AjaxEventBehavio
 		try
 		{
 			formComponent.inputChanged();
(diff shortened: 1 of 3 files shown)


martin-g:
Thank you for your contribution!

I am not sure it is a good idea, though.
With this change we will make all Ajax form component submits "non-expensive", i.e. a subset of validator will be used. I guess some applications will need to enable an expensive validator in some use case, so we will have to make it configurable. And it will become more complex than it needs to be.

IMO it is simple enough to check whether the current request is Ajax one in the validator and decide what to do. One can even use request parameters/headers for even more flexibility.

This is just my opinion! Let's see what others think too.

martin-g:
Another Wicket developer commented in the dev@ mailing list: http://markmail.org/message/o3qsaetsloo73lwt

martin-g:
@amichalowski Please close this PR. We don't have permissions to close it ourselves. Thanks!

PR #178, opened 2016-09-05
PageRequestHandlerTracker issue

amichalowski:
PageRequestHandlerTracker doesn't work with CryptoMapper.
In mapRequest method in CryptoMapper it is created instance of RequestSettingRequestHandler and PageRequestHandlerTracker doesn't looking for IPageRequestHandler inside IRequestHandlerDelegate.

Reproduce steps:
1 application init method
```java
@Override
public void init() {
	super.init();
	getRequestCycleListeners().add(new PageRequestHandlerTracker());
	setRootRequestMapper(new CryptoMapper(getRootRequestMapper(), this));
}
```
2 Try to get last handler in page or component
```java
@Override
protected void onBeforeRender() {
	super.onBeforeRender();
	RequestCycle cycle = getRequestCycle();
	IPageRequestHandler lastHandler = PageRequestHandlerTracker.getLastHandler(cycle);
	System.out.println(lastHandler.getPageClass().getCanonicalName());
}
```


martin-g:
Thank you!

PR #182, opened 2016-10-13
Page instantiation protection by combined ruleset.

rene-d-menoto:
There is currently no way to annotate a page with instantiation rules on
multiple roles. It is possible to use an AuthorizeInstantiation. But
this can only roles using an OR ruleset.

With this annotation in use we are able to combine multiple OR rulesets
in an AND relation.

Example:
A page can be instantiate only by users with ADMIN and DATABASE role.


martin-g:
Created a ticket: https://issues.apache.org/jira/browse/WICKET-6265

martin-g:
I've cherry picked this commit and applied it at `master` (8.x) with https://git1-us-west.apache.org/repos/asf?p=wicket.git;a=commit;h=e9b812c3.
Git merge/rebase leads to too much noise in the commits@ mailing list.
@rene-d-menoto Thank you for the contribution! Please close this PR!

rene-d-menoto:
Thanks for merging.

PR #184, opened 2016-10-13
Page instantiation protection by combined ruleset.

rene-d-menoto:
There is currently no way to annotate a page with instantiation rules on
multiple roles. It is possible to use an AuthorizeInstantiation. But
this can only roles using an OR ruleset.

With this annotation in use we are able to combine multiple OR rulesets
in an AND relation.

Example:
A page can be instantiate only by users with ADMIN and DATABASE role.


martin-g:
Created a ticket: https://issues.apache.org/jira/browse/WICKET-6265

martin-g:
I've decided to not merge this to 6.x.
New features are added only to 7.x and 8.x.
If this is really needed for 6.x then we may reconsider it!
@rene-d-menoto Please change this PR! Thank you!

rene-d-menoto:
This was only a service for the wicket-6 users.

PR #185, opened 2016-10-28
TimeZone handling in examples: apply settings client time zone on DateFormat

barney2k7:
DateFormat previously used the jvm's default time zone, which leads to somewhat confusing output on the example page: 'Based on your settings, your time is: 4:46:15 AM EDT (and your time zone is Central European Time)'
After this patch, it shows '10:57:49 AM CEST (and your time zone is Central European Time)' (jvm's default time zone no longer relevant)


klopfdreh:
+1 to merge.

@martin-g - WDYT?

martin-g:
Thank you, @barney2k7 !

PR #186, opened 2016-10-31
WICKET-6264: Form: improve error message for disabled/invisible IForm…

barney2k7:
…SubmittingComponent


klopfdreh:
+1 to merge.

@martin-g - WDYT?

martin-g:
Looks good!
Please create a ticket in JIRA so it appears in the changelog.
The same improvement could be done for the problem with Fragments in 7.5.0, if the component is not rendered.

PR #188, opened 2016-11-07
Feature/locale format fix

duesenklipper:
On some systems (like my desktop ;-) ) there is not one single `Locale`, but rather different ones for text and formats. For example, I use `en_US` for nearly everything, but `de_DE` for formatting. In this situation, `DurationTest` fails, because `DurationTest` correctly uses `NumberFormat.getNumberInstance`, which uses `Locale.getDefault(Locale.Category.FORMAT)`. `Duration` itself, however, just calls `Locale.getDefault()` and thus misses the difference. This patch simply fixes `Duration` to use the formatting locale.

### DIFF
--- a/wicket-util/src/test/java/org/apache/wicket/util/lang/BytesTest.java
+++ b/wicket-util/src/test/java/org/apache/wicket/util/lang/BytesTest.java
@@ -40,10 +40,10 @@ public class BytesTest extends Assert
 	@Before
 	public void before()
 	{
-		defaultLocale = Locale.getDefault();
+		defaultLocale = Locale.getDefault(Locale.Category.FORMAT);
 
 		// these tests run in US locale.
-		Locale.setDefault(Locale.US);
+		Locale.setDefault(Locale.Category.FORMAT, Locale.US);
 	}
 
 	/**
@@ -52,7 +52,7 @@ public class BytesTest extends Assert
 	@After
 	public void after()
 	{
-		Locale.setDefault(defaultLocale);
+		Locale.setDefault(Locale.Category.FORMAT, defaultLocale);
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


klopfdreh:
+1 to merge. @martin-g WDYT?

duesenklipper:
I'd merge it myself, but I can so rarely work on wicket itself these days that I prefer having someone look at it first.

If this one is accepted, I'll add the same thing to 8.x myself.

martin-g:
1. A ticket in JIRA should be created for any PR, so that it appears in the CHANGELOG and users can see it.
2. Quick search for calls to `Locale.getDefault()` shows 40 occurrences in Wicket. IMO we should go over all of them and see whether they need such improvements as well.

My 2c.

duesenklipper:
Ticket is at https://issues.apache.org/jira/browse/WICKET-6273, and I went through the various calls to Locale.getDefault. What do you think?

wicket-util/src/test/java/org/apache/wicket/util/lang/BytesTest.java
```diff
@@ -40,10 +40,10 @@
 	@Before
 	public void before()
 	{
-		defaultLocale = Locale.getDefault();
+		defaultLocale = Locale.getDefault(Locale.Category.FORMAT);
 
 		// these tests run in US locale.
-		Locale.setDefault(Locale.US);
+		Locale.setDefault(Locale.Category.FORMAT, Locale.US);
```
martin-g: Would the test still pass without this change ?
I.e. if Category.FORMAT is not specified then does it fallback to the `generic` locale ?

duesenklipper:
Yes, it would have still worked - I changed it so it explicitly tests for a split Locale now. If no formatting locale is set, `Locale.getDefault(FORMAT)` falls back to the "normal" default locale.

martin-g:
+1 to merge

PR #189, opened 2016-11-07
make NoRecordsToolbar use onConfigure/setVisible instead of overridin…

duesenklipper:
…g isVisible

Overriding isVisible is discouraged, visibility should be set just once. I know the rowcount is cached nowadays, but some time ago this actually bit us (admittedly in some way that I don't remember now, but I think by getRowCount being called too early in a request).

If this one is accepted, I'll to the port to 8.x myself.


martin-g:
+1

PR #190, opened 2016-11-16
AjaxDownload is implemented using hidden iframe

solomax:
https://issues.apache.org/jira/browse/WICKET-6286


Jezza:
This seems overtly complex given you can achieve the same thing with resources. 
Requiring a behaviour to be attached seems a bit too much, as well as the potential page lock.

martin-g:
@Jezza Please check the implementation by @svenmeier at https://issues.apache.org/jira/browse/WICKET-6286.

martin-g:
@solomax Please close this PR! We will use the solution in https://issues.apache.org/jira/browse/WICKET-6286. Thanks!

PR #191, opened 2016-11-16
AjaxDownload is implemented using hidden iframe

solomax:
https://issues.apache.org/jira/browse/WICKET-6286


martin-g:
@solomax Please close this PR! We will use the solution in https://issues.apache.org/jira/browse/WICKET-6286. Thanks!

PR #193, opened 2016-11-24
WICKET-6287 Switch from json.org to open-json - 8.x

klopfdreh:
Repository: https://github.com/tdunning/open-json


pom.xml
```diff
@@ -447,6 +447,12 @@
 				<type>jar</type>
 			</dependency>
 			<dependency>
+		    	<groupId>com.tdunning</groupId>
+		    	<artifactId>json</artifactId>
+				<version>1.2-SNAPSHOT</version>
```
bitstorm: Hi @klopfdreh ,

why we can't use the current version 1.1 and wait for 1.2?

klopfdreh: Hi @bitstorm ,

1.2 has been just released. 😄 

Anyway the were two reasons why to wait for this version.

1. The slash / should only be escaped after a "<" bracket (https://github.com/tdunning/open-json/commit/e1541126bf3206fa7a2b243f3791e274822128c1)
2. JSONFunction was not implemented yet. (https://github.com/tdunning/open-json/commit/174ab93ac966f0213ac4a4bd3ac5b17e0920a7c8) - this is required for json like this: { "mykey": window.myfunction() }

We can switch the version to 1.2 and it should work with this PR.

kind regards

Tobias

bitstorm: @klopfdreh 
Great! Good job Tobias! I think we can go ahead with this PR and merge it

klopfdreh: @bitstorm - are you able to check this PR out an have a look if everything builds fine?

klopfdreh: Hi @bitstorm ,

that's what I mentioned here:

https://github.com/apache/wicket/pull/195#issuecomment-263071173

Sadly the API is not exact the same, but we have to move. Even if we take another json framework we can't provide exact the same methods.

But I hope with open-json it is not that difficult.

bitstorm: @klopfdreh 
I'm done with wicketstuff. I think 4-5 modules (select2, whiteboard, etc...) should be manually tested but don't expect problems.

bitstorm: @klopfdreh 
JSONFunction are rendered as regular strings, surrounded by double quotes. But I think the culprit is here:

https://github.com/tdunning/open-json/blob/master/src/main/java/org/json/JSONStringer.java#L261

We should rename JsonFunction back to JSONFunction

bitstorm: I think the problem is that the new library escapes special characters also for JSONFunction, so 'alert("")' becomes 'alert(\"\")' and can not be evaluated by javascript.

klopfdreh: Hi @bitstorm ,

I am going to write a Unit Test for this and request another PR for open-json, soon.

kind regards

Tobias

tdunning: I just pushed version 1.3 that has quoting fixes and a bean constructor.

klopfdreh: @bitstorm - is everything working with 1.3, now?

If yes we can integrate this branch I think.

tdunning:
Not sure why, but I see 1.2-SNAPSHOT in the comments coming through.

That is going to point at open-json 1.3 when you actually move forward,
right?

(just call me "better safe than sorry")

bitstorm:
@tdunning @klopfdreh 
yes, the 1.2-SNAPSHOT must turn into 1.3

klopfdreh:
Hi @bitstorm - I switched it to version 1.3.

klopfdreh:
Summary before integration:

* Because Apache Wicket 8 is at milestone level API changes are allowed and because of this all JSON library classes are removed and the external dependency of open-json has been added. This way it is easier to keep the library updated. Users have to organize imports.

bitstorm:
@klopfdreh @martin-g 
Shall we go ahead merging or do you prefer to port the pr to the other two branches first?

klopfdreh:
@bitstorm - there are two different PRs for 6.x and 7.x to preserve the API integrity.

bitstorm:
Sorry I wasn't clear. I asked if this PR is complete and if we should proceed merging it and wicketstuff/core#566.

klopfdreh:
I vote to merge the 8.x PR. For 6.x / 7.x @martin-g already did build tests and pointed out some issues and I would suggest to check out the PRs and test them even with wicketstuff to ensure not to break something.

bitstorm:
+1 to merge. I've tested with WicketStuff and with @sebfz1  wicket-jquery-ui. Two PR have been created for this two project but everything seems to work fine with them.

tdunning:
I have pushed a few additional versions of open-json that might help (or
hinder) wicket.

The basic issue was that in the bean constructor, objects were not
converted all the way down. This led to really bad behavior because the
native to-string was used on things like lists that led to ugly and
non-functional results.

This was resolved in open-json 1.6 in a way that I don't think would hurt
wicket, but it might be good for you guys to let me know if it does and if
there is a middle ground that does not.

klopfdreh:
Hi @tdunning - thanks for keeping us up to date with the changes! 👍 

I am going to set the version to 1.6 so that we get the latest version.

klopfdreh:
Closed due to:
* https://github.com/apache/wicket/commit/4e3194466c4df50c9d38ea1ed50800dea592c72b
* https://github.com/apache/wicket/commit/cc98cb0675ffbe9be956e33d6b827f4d61f9f3d7
* https://github.com/apache/wicket/commit/49a486ac9fb379426b36ee29a8c9101fa818e669

PR #194, opened 2016-11-25
WICKET-6287 Switch from json.org to open-json - 6.x

klopfdreh:
There are clirr errors, because of some classes missing in open-json


klopfdreh:
fyi @martin-g

solomax:
I believe open-json files need to be updated to correspond 1.3 ....

klopfdreh:
@solomax - yep - I updated the classes with the versions of 1.3. Anyway there are clirr issues even if we copy all missing classes and override methods / constructors to throw WicketRuntineExceptions, because some method signatures have been changed.

solomax:
Actually this can be handled using double-proxy, but I would vote for API break in this case :)

klopfdreh:
I am going to try this out, soon!

klopfdreh:
Hi @solomax ,

I managed to resolve the clirr issues: https://github.com/apache/wicket/pull/195#issuecomment-265422034

klopfdreh:
@martin-g , @dashorst , @bitstorm - would you be so kind and review the changes / check if there are no clirr issues for the 6.x branch? I don't have jdk 6 / jdk 7 installed so I get a "Invalid byte tag in constant pool: 18" of clirr.

martin-g:
This PR fails with:
```
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5:compile (default-compile) on project wicket-core: Compilation failure: Compilation failure:
[ERROR] /home/martin/git/apache/wicket-6.x/wicket-core/src/main/java/org/apache/wicket/ajax/json/JSONObject.java:[30,17] cannot find symbol
[ERROR] symbol  : class Objects
[ERROR] location: package java.util
[ERROR] /home/martin/git/apache/wicket-6.x/wicket-core/src/main/java/org/apache/wicket/ajax/json/JSONObject.java:[119,20] cannot find symbol
[ERROR] symbol: variable Objects
[ERROR] -> [Help 1]
```

klopfdreh:
Hi @martin-g - thanks for the hint - I am going to have a look at it, soon.

klopfdreh:
Hi @martin-g - this was because Objects.hashCode is only available since idk > 7.

I replaced the call with a return 0; - this is done by Objects.hashCode(null) anyway.

I also created a ticket for open-json to change the code so that there is a "real" jdk 6 compatibility. 😄

klopfdreh:
Summary before integration:

* Method calls not supported by open-json throw a WicketRuntimeException to inform the user
* If users still need json.org they have to reorganize imports and add the json.org dependency in maven (beware of the license issues in this case!!)
* All classes are preserved due to clirr checks, but are converted to dummies if there is no functionality left
* There has been a special change to JSONObject, because Objects.hashCode is only available since jdk > 7

klopfdreh:
Closed due to:
* https://github.com/apache/wicket/commit/0ba87f062cad8664d216abccb2a47126f637b87c
* https://github.com/apache/wicket/commit/6f764146a007ab38686a7d54210db0fcdfc2a7ef
* https://github.com/apache/wicket/commit/a6bf0e29bbc9ccebaa7588f566e15f0ac1d2eaa6
* https://github.com/apache/wicket/commit/ad978211a5a550e7902eba354b03ec6618649579

PR #195, opened 2016-11-25
WICKET-6287 Switch from json.org to open-json - 7.x

klopfdreh:
There are clirr errors, because of some classes missing in open-json


klopfdreh:
fyi @martin-g

klopfdreh:
Hi,

after some research I found out that there is no way to use open-json without API breaks, because it uses the same class names / methods so that you can't redirect the calls, also the API is slightly different in some places. I just tried to implement the missing classes with stubs that only throws RuntimeExceptions so that the API is preserved, but without success.

This also affects the PR https://github.com/apache/wicket/pull/194

I would suggest to go for it anyway, because open-json mostly the same as json.org

klopfdreh:
I managed to resolve the clirr issues.

What I did was:
* Revert back all class signatures that were copied from json.org and throw a corresponding exception to inform users about the switch

If the user updates Apache Wicket and is ok with the license issue of JSON.org the users code has to be switched to the corresponding maven dependency of the JSON.org library and the imports has to be reorganized, but for 90% the code will work as before.

klopfdreh:
@martin-g , @dashorst , @bitstorm - would you be so kind and review the changes / check if there are no clirr issues for the 7.x branch? I don't have jdk 7 installed so I get a "Invalid byte tag in constant pool: 18" of clirr.

martin-g:
Compilation and Clirr pass but a test fails:

```
Failed tests: 
  CssUrlReplacerTest.base64EncodedImage:165->Assert.assertThat:923->Assert.assertThat:956 
Expected: a string containing ".class {background-image: url(data:image/gif;base64,R0lGODlh1wATAXAAACH5BAEAAP8ALAAAAADXA"
     but: was ".class {background-image: url(data:image/gif;base64,R0lGODlhUAB4ANUAANbX1hQSEXV0c/fz8tCBfc3NzFBNTPO6ttzIx/LKyIGAf84jGri4t9avrainppiXlut9dqZSTI41MNzDwbCMitWQjLF0cJVzcc52cctKQ8zW1sLCwe+WkdnQz8U4MIVaWFxaWYuJiKg+Oa2sq52dnMygncvOz83QzzAuLbdbVpKRkEJAP9LDwpxmY+enpN3c3NLR0WdmZcnIyLKxsPHl5L++vc5oY8rLyuO9vMfGxuxqY6OhodZYUcXFxM7Ozv///yH5BAAAAAAALAAAAABQAHgAAAb/wJ9wSCwaj0jjAEFpfZwt54XSGSSv2Kw2OZh9QJ9paUwpfy+srXrNfkEpjdcuJqgNdiRW86Oysv+ARAhOIR01MQMOKCs7PyAqPw......
````

martin-g:
wicket-7.x fails the same way.
@duesenklipper Ping!

duesenklipper:
This might be a stupid question, but I must be missing something. How do I check out the pull request branch?

duesenklipper:
(that said, that test works in the original wicket-7.x branch. i'll try and figure this out as soon as i can get at the branch)

martin-g:
The build fails for me at branch `wicket-7.x` with the same error as in this PR.
I merge PRs from Github with https://github.com/martin-g/dotfiles/blob/master/bin/github-pr.sh: 
- $ cd ~/git/apache/wicket-7.x
- $ github-pr.sh 195

duesenklipper:
I figured out to do 

$ git fetch github-upstream refs/pull/195/head
$ git checkout FETCH_HEAD

From that I got commit 4dbdbdfe, and that builds cleanly for me. I also tried to just run that particular test in IDEA, and it worked too.

duesenklipper:
(I ran `mvn clean package`)

martin-g:
The problem seems to be here.
BuildBot passes the build until Arquilian tests: https://ci.apache.org/builders/wicket-branch-7.x/builds/333

duesenklipper:
That's a different problem though (and it worked for me too). I'm building 5bfa280b2b7fb9dc7861d582a9c6e37d106be879 locally now to see what it does (the one you linked to)

martin-g:
I was saying that if BuildBot is fine (at wicket-core) then the problem must be here.
The problem with Arquillian is known. It waits someone to debug it. Too bad the JavaEE fans do not give it more love :-)

duesenklipper:
Oh, you mean a problem on your machine? If that's the case, I'll stop trying to recreate it here ;-)

klopfdreh:
To come to the point: In which way the JSON lib change affects CssUrlReplacerTest.

duesenklipper:
@klopfdreh I don't think it does. It appears to be a problem on @martin-g's machine. The JSON stuff doesn't even come close to CssUrlReplacer.

klopfdreh:
I don't think this is regarding the JSON change. The String is to long and it is a different as the expected one and the CssUrlReplacer don't use JSON at all (only regex) - maybe it has something to do with the locale changes? (Only an assumption)

klopfdreh:
Haha @duesenklipper - you wrote a bit earlier - yep I know the CssUrlReplacer and @martin-g , too. We implemented it 😊

klopfdreh:
Summary before integration:

* Method calls not supported by open-json throw a WicketRuntimeException to inform the user
* If the users still needs json.org they have to reorganize imports and add the json.org dependency in maven (beware of the license issues in this case!!)
* All classes are preserved due to clirr checks, but are converted to dummies if there is no functionality left

duesenklipper:
@klopfdreh I know you did ;-) 

@martin-g I found it. @klopfdreh is right, it has something to do with Locale. My fault - for some reason I lost the `@Before` method in `CssUrlReplacerTest` that set the Locale to English, when I updated that test for data urls. I'll push the fix to the main branches. Sorry about that!

martin-g:
@duesenklipper I am the non-German guy here but it seems it is just me who uses de_DE locale to catch such kind of problems :-) Your fix is good!

@klopfdreh Please update this branch to latest wicket-7.x and I'll retest it again with JDK1.7

duesenklipper:
@martin-g I hate German translations of computer terms, which is the main reason my systems are all in English. But as mentioned elsewhere, I use a split locale instead, with formatting set to German. Thus I break things in other ways ;-)

klopfdreh:
@martin-g I rebased it and pushed it again. Should be up to date now.

martin-g:
The build passes with JDK1.7!

klopfdreh:
Great! 😄 - so if everything is working - we should integrate all PR for open-json. But I suggest to test all dependencies (also for wicketstuff)

klopfdreh:
Closed due to:
* https://github.com/apache/wicket/commit/3dacec4c55bb95618f604ad2ccea279346de279c
* https://github.com/apache/wicket/commit/6b1d53b17a0aa77a8b131fd161e12e08d335a251
* https://github.com/apache/wicket/commit/5348b3e00518e5636126e724f1e62ea1aebbd8b6

bitstorm:
@klopfdreh 
Hi,

One minor issue I've seen with this PR is that now ' new JSONObject' throws checked exceptions, so the code must be adapted some how.  This happened with WicketStuff [here](https://travis-ci.org/wicketstuff/core/builds/188994145). It's not a big issue, but maybe this could be a problem with other libraries.

klopfdreh:
Hi @bitstorm ,
I saw that JSONObject throws a JSONException which extends RuntimeException - I thought it is unchecked.

klopfdreh:
I changed the method propertiesAsMap to throw the JSONException, because the first statement of a constructor always has to be the super / this call.

klopfdreh:
@bitstorm - the issue about the checked exceptions is fixed in the master branch we only have to wait until the release and increase our version in the pom.xml. Then there should be no further issue.

tdunning:
Should get that release out later today.

klopfdreh:
Thanks a lot @tdunning !

tdunning:
The artifacts for version 1.8 Open Json are on central.

https://repo1.maven.org/maven2/com/tdunning/json/1.8/

klopfdreh:
@bitstorm - the changes are made - JSONObject does not throw any checked exceptions anymore.

bitstorm:
thank you all!

PR #202, opened 2016-11-28
WICKET-6290 fix CssUrlReplacer so it does not mangle data URIs

duesenklipper:


martin-g:
:+1:

klopfdreh:
Really great change! +1

PR #205, opened 2017-01-22
Add Option isQuietly to StatelessChecker

kemsakurai:
For this reason, I would like to propose adding one option so that it can handle this use case.


svenmeier:
Thanks for your patch. What do you think about a hook method instead, that people can override to do whatever they like with a failing component?

	/**
	 * The given component claims to be stateless but isn't.
	 * 
	 * @param component component failing stateless check
	 * @param reason explaining reason
	 */
	protected void fail(Component component, String reason)
	{
		throw new IllegalArgumentException("'" + component + "' claims to be stateless but isn't. " + reason);
	}

martin-g:
I don't understand the problem and the suggest solution.
What do you mean by `we throw exceptions without fail` ?
Could you please re-phrase the description of this pull request? Thanks!

svenmeier:
As I understood it, he's using a customized copy of StatelessChecker that logs instead of throwing an exception.

kemsakurai:
I'm sorry. I am not very good at English and I did not get it well.
The description of pull request has been fixed.
Since I think that the method of making a suggested hook method is a better way,
I will reimplement it. This pull request is closed.
Thanks.

PR #207, opened 2017-01-23
WICKET-6311 Russian sign out page is added

solomax:


martin-g:
I've amended the commit to add 'WICKET-6311' in the message and now the auto-close won't work.
@solomax Please close this PR! Thanks!

PR #208, opened 2017-01-25
 Add hook method for exception handling to StatelessChecker

kemsakurai:
I implemented a hook method based on #205  feedback.

### DIFF
--- a/wicket-devutils/src/main/java/org/apache/wicket/devutils/stateless/StatelessChecker.java
+++ b/wicket-devutils/src/main/java/org/apache/wicket/devutils/stateless/StatelessChecker.java
@@ -31,10 +31,10 @@ import org.apache.wicket.util.visit.IVisitor;
  * stateless. This is a utility that is intended for use primarily during development. If you add an
  * instance of this class to your application, it will check all components or pages marked with the
  * <tt>StatelessComponent</tt> annotation to make sure that they are stateless as you intended.
- * 
+ *
  * This is useful when trying to maintain stateless pages since it is very easy to inadvertently add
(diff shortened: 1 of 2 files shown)


martin-g:
Do you really need all those `#failXYZ()` methods ?
One isn't enough ?

kemsakurai:
In my use case, one `fail` method is enough.

However, in the current implementation, for components with stateful behavior, IllegalStateException is thrown.
I did not change this part.
Therefore, I created a `fail` method for each component.

To make one `fail` method,
Can I change IllegalStateException to IllegalArgmentException?

svenmeier:
Neither IllegalStateException nor IllegalArgumentException seems right to me. Probably it's better to just throw WicketRuntimeException.

martin-g:
`IllegalStateException` looks good to me. The developer expects that the page is in stateless state but it is not!
What about introducing a custom exception and pass it to the `#fail()` method ? By default it will be thrown but the application can override the method and log it or do whatever it finds suitable.

`if (!condition) { fail(new CustomException("A good message")) }`
and 
````java
protected void fail(CustomException cx) {
  throw cx;
}
```
or 
````java
protected void fail(CustomException cx) {
  LOG.warn("The page is not really stateful: {}", cx.getMessage(), cx);
}
```

kemsakurai:
Thanks for the suggestion.
I think you can create a `StatelessCheckFailureException` that inherits from WicketRunTimeException  
and let StateFullChecker throw a StatelessCheckFailureException.
I will try implementing it again.
This pull request is closed. 
Thanks.

PR #210, opened 2017-02-07
Wicket 4201 improved page provider

martin-g:
Creating a Pull Request so we can comment on the changes.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
@@ -236,7 +236,7 @@ public class PageProvider implements IPageProvider, IClusterable
 	 *         provider
 	 */
 	@Override
-	public final boolean isPageInstanceFresh()
+	public final boolean doesProvideNewPage()
 	{
 		if (!this.provision.wasResolved())
 		{
--- a/wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
@@ -20,7 +20,6 @@ import org.apache.wicket.Application;
 import org.apache.wicket.core.request.mapper.IPageSource;
 import org.apache.wicket.core.request.mapper.StalePageException;
 import org.apache.wicket.page.IPageManager;
-import org.apache.wicket.pageStore.IPageStore;
 import org.apache.wicket.protocol.http.PageExpiredException;
 import org.apache.wicket.request.IRequestHandler;
 import org.apache.wicket.request.IRequestMapper;
@@ -56,12 +55,11 @@ public class PageProvider implements IPageProvider, IClusterable
 
 	private IPageSource pageSource;
 
-	private transient IRequestablePage pageInstance;
-	private boolean pageInstanceIsFresh;
-
 	private Class<? extends IRequestablePage> pageClass;
 
 	private PageParameters pageParameters;
+
+	private Provision provision;
 
 	/**
 	 * Creates a new page provider object. Upon calling of {@link #getPageInstance()} this provider
@@ -153,9 +151,19 @@ public class PageProvider implements IPageProvider, IClusterable
 	{
 		Args.notNull(page, "page");
 
-		pageInstance = page;
+		provision = new Provision(page);
 		pageId = page.getPageId();
 		renderCount = page.getRenderCount();
+	}
+
+	private Provision getProvision()
+	{
+		if (provision == null)
+		{
+			provision = new Provision(getPageSource()).resolvePageInstance(pageId, pageClass,
+				pageParameters, renderCount);
+		}
+		return provision;
 	}
 
 	/**
@@ -164,16 +172,11 @@ public class PageProvider implements IPageProvider, IClusterable
 	@Override
 	public IRequestablePage getPageInstance()
 	{
-		if (pageInstance == null)
+		if (!getProvision().didResolvePage() && !getProvision().doesProvideNewPage())
 		{
-			resolvePageInstance(pageId, pageClass, pageParameters, renderCount);
-
-			if (pageInstance == null)
-			{
-				throw new PageExpiredException("Page with id '" + pageId + "' has expired.");
-			}
+			throw new PageExpiredException("Page with id '" + pageId + "' has expired.");
 		}
-		return pageInstance;
+		return getProvision().get();
 	}
 
 	/**
@@ -188,7 +191,7 @@ public class PageProvider implements IPageProvider, IClusterable
 		}
 		else if (isNewPageInstance() == false)
 		{
-			return pageInstance.getPageParameters();
+			return getProvision().get().getPageParameters();
 		}
 		else
 		{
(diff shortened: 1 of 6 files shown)


wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
```diff
@@ -54,15 +52,14 @@
 
 	private final Integer pageId;
 
-	private IPageSource pageSource;
-
-	private transient IRequestablePage pageInstance;
-	private boolean pageInstanceIsFresh;
+	private transient IPageSource pageSource;
```
martin-g: IMO using `transient` is just delaying the problem.
`PageProvider` should not be serializable (IClusterable).
If it is serialized and then deserialized then most probably some call with fail with NullPointerException because of those transient fields.

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
```diff
@@ -197,26 +194,49 @@ else if (isNewPageInstance() == false)
...
+	 * 
+	 * @return {@code true} iff the page instance held by this provider was instantiated by the
+	 *         provider
+	 */
+	@Override
+	public final boolean doesProvideNewPage()
+	{
+		if (provision == null)
+		{
+			throw new IllegalStateException("Page instance not yet resolved");
```
martin-g: This doesn't look nice.
At https://issues.apache.org/jira/browse/WICKET-4201?focusedCommentId=15853108&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15853108 you say 
 ` removed the PageExpiredException from PageProvider since it breaks the class coherence. It's designed to provide pages and related metadata (such as if the page was expired included), not to change the application flow on application exceptions.` but this runtime exception does exactly this.
IMO it would be better to return a three state result: `NEW`, `OLD` and `UNKNOWN`

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
```diff
@@ -349,6 +306,11 @@ public void detach()
 	 */
 	public void setPageSource(IPageSource pageSource)
 	{
+		if (provision != null)
+		{
+			throw new IllegalStateException(
+				"A provision was already been done. The provider can be forcefully detached or a new one needs to be used to provide using this page source.");
```
martin-g: `"A provision has already been done. The provider can be forcefully detached or a new one has to be created be be able to use this page source."` ?

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
```diff
@@ -388,49 +350,95 @@ public Integer getRenderCount()
...
+
+				failedToFindStoredPage = page == null;
+			}
+
+			return this;
+		}
+
+		void detach()
+		{
+			if (page != null)
```
martin-g: IMO this should just reset its own member fields.
Why do you think it should detach the page too ?

wicket-core/src/test/java/org/apache/wicket/core/request/mapper/MountedMapperTest.java
```diff
@@ -714,7 +714,7 @@ public boolean isNewPageInstance()
 	@Test
 	public void placeholderEncode4()
 	{
-		PageProvider provider = new PageProvider(new MockPage())
+		PageProvider provider = new PageProvider(MockPage.class)
```
martin-g: Why this change is needed ?
This is not the same as before.

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
pedrosans: Looking bad indeed, I fixed later during my work on the branch: https://github.com/apache/wicket/pull/210/commits/b64f2b4b2cb50a594cabb50a0c1eeca43485f6b9#diff-36b11db0d92a540368a080c520de953aL177

The three states looks overcomplicated. There's no place in Wicket core needing to test such flag for the value "OLD".

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
pedrosans: My bad, I meant to say: "to be able to **change** the page source". I will improve this exception message.

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
pedrosans: Good point, I just kept the old behaviour here.

wicket-core/src/test/java/org/apache/wicket/core/request/mapper/MountedMapperTest.java
pedrosans: Not the same and the change is needed. This test follows up by changing the page provider which will have no impact on the provider that already holds an instance.

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
martin-g: For me it is strange that a method with return type `boolean` returns either `true` or throws an exception.

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
martin-g: Well, if the old code detached it then let's keep it.
There is an open ticket in JIRA that sometimes a Page is detached twice in the same request cycle, that's why I've asked.

wicket-core/src/test/java/org/apache/wicket/core/request/mapper/MountedMapperTest.java
martin-g: I didn't understand you.
```java
/**
	 * WICKET-5247 page instantiated without required parameters won't be mapped
	 */
	@Test
	public void placeholderEncode4()
	{
		PageProvider provider = new PageProvider(new MockPage())
		{
			@Override
			public boolean isNewPageInstance()
			{
				return false;
			}
		};
		provider.setPageSource(context);
		IRequestHandler handler = new RenderPageRequestHandler(provider);
		Url url = placeholderEncoder.mapHandler(handler);
		assertNull(url);
	}
```
The test setups a PageProvider and uses it. There is no replacement after.
By making this change I think we don't test the same as what broke in WICKET-5247.

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
pedrosans: It can lead to inconsistent behaviour indeed, but not NPE since it will default to the application one. I will support the removal of IClusterable interface. From what I looked in the code, page provider is just a core's internal communicator of pages construction and fetch logic. It's used during the work to handle a page request and never shared on the session. I would bet that at some point a developer needed ou wanted a page provider as a page attribute and ended up marking it as serializable.

wicket-core/src/test/java/org/apache/wicket/core/request/mapper/MountedMapperTest.java
pedrosans: My bad again, I meant to say: "... test follows up by changing the page **source** ..."
The line: 
      provider.setPageSource(context);
was setting up a page source that was never used because the page provider already had a page instance (the one passed as a parameter in the page provider). 

I don't know why the test sets the page source, but either way a page with no parameters will be provided making it to test the same thing. Maybe we can just remove the custom page source.

wicket-core/src/main/java/org/apache/wicket/core/request/handler/PageProvider.java
pedrosans: oh, just just saw that this is a comment on the doesProvideNewPage method (github interface didn't show the method signature and I assumed we were talking about PageProvider#getPageInstance, where I indeed removed the illegal state exception. PageProvider#doesProvideNewPage can return true or false, but regardless I agreed that it shouldn't throw a runtime exception. Thanks for pointing it, will fix next.

martin-g:
@pedrosans I think this could be merged to `master`!

PR #214, opened 2017-02-18
experimenting with suspension of requestCycle via servlet 3 async

svenmeier:
Notes:
- enabled async-supported in web.xml for ComponentReferenceApplication
- LinkPage shows usage of RequestCycle#suspend()
- pageAccessSynchronizer uses the requestCycle instead of the thread
now, since two threads might process a page (PageAccessSynchronizerTest
still fails)
- ServletWebRequest must not call httpServletRequest#getContextPath(),
since it returns null during async processing

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/page/PageAccessSynchronizer.java
+++ b/wicket-core/src/main/java/org/apache/wicket/page/PageAccessSynchronizer.java
@@ -23,6 +23,7 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.function.Supplier;
 
 import org.apache.wicket.Application;
+import org.apache.wicket.request.cycle.RequestCycle;
 import org.apache.wicket.settings.ExceptionSettings.ThreadDumpStrategy;
 import org.apache.wicket.util.LazyInitializer;
 import org.apache.wicket.util.lang.Threads;
@@ -93,8 +94,8 @@ public class PageAccessSynchronizer implements Serializable
 	 */
 	public void lockPage(int pageId) throws CouldNotLockPageException
 	{
-		final Thread thread = Thread.currentThread();
-		final PageLock lock = new PageLock(pageId, thread);
+		final RequestCycle cycle = RequestCycle.get();
+		final PageLock lock = new PageLock(pageId, cycle);
 		final Time start = Time.now();
 
 		boolean locked = false;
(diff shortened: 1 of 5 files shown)


wicket-core/src/main/java/org/apache/wicket/page/PageAccessSynchronizer.java
```diff
@@ -93,8 +94,8 @@ public Duration getTimeout(int pageId)
 	 */
 	public void lockPage(int pageId) throws CouldNotLockPageException
 	{
-		final Thread thread = Thread.currentThread();
-		final PageLock lock = new PageLock(pageId, thread);
+		final RequestCycle cycle = RequestCycle.get();
```
martin-g: This is not a good idea.
Two threads should not work on the same page instance at the same time.
This will lead to concurrency related issues.
E.g. two Ajax requests will have their own Request Cycles but might work on the same page instance this way. Without AsyncContext being needed at all.

svenmeier: Not true!

As long as one RequestCycle holds the lock on a page, no other RequestCycle can get a lock. It doesn't make any difference whether we lock on the thread or the RequestCycle.

martin-g: How a RequestCycle holds a lock on a page ?
Where is this lock in the code ?

svenmeier: PageLock is the lock: as long as there is a PageLock instance, all access to the corresponding page has to wait.

Currently PageLock holds a reference to the thread, but the thread isn't used actually apart from identifying the lock. The RequestCycle can be used equally well for that.

martin-g: OK. I see!
But why do you need to change this code at all ? 
PageLock#getCycle() is not used anywhere in the PR.
The thread is very useful to identify problems with long running requests at the moment. The thread stack trace is telling me much more than a long (the request cycle start time).
I'd prefer to add the RequestCycle as yet another field to PageLock than removing the thread. But I don't really see the need to change PageAccessSynchronizer at all.

svenmeier: Sure, we could keep the thread reference. But when the request is suspended, a certain time *no* thread is keeping the lock (the page is still locked though) and from a certain point in time another thread would pick up the lock.

martin-g: In that timeframe the page should be unlocked. On resume the new thread should acquire a new lock or wait if it is not available at the moment.

svenmeier: No I don't think so: The request processing hasn't finished yet (e.g. rendering the page hasn't been started yet), so the lock on the page cannot be released.

klopfdreh: Any updates here?

svenmeier: Nothing from me - I didn't have any use for this yet, so I'm not sure how useful this is for components and how it should work exactly.

solomax:
I guess it can be dropped? :)

PR #215, opened 2017-02-18
Add hook method for exception handling to StatelessChecker

kemsakurai:
I implemented a hook method based on #208 feedback and add StatelessCheckFailureException.


martin-g:
Thank you!

PR #216, opened 2017-02-23
PropertyValidator does not work with dynamic validation groups

paulspeijers:
The Wicket property validator does not work with dynamic validation groups because of the following issues:

1. It can not cope with the case in which the Default validation group is provided explicitly
1. The initial validation groups determine whether required should be set on the component. But the PropertyValidator never unsets the flag.

When using a Wicket property validator you can provide a model for the validation groups:

```
IModel<Class<?>[]> validationGroups = Model.of(new Class[]{Default.class});
new PropertyValidator<>(validationGroups);
```

Let's say I have the following model object:

```
public class MyObject {

    @NotNull
    private String name;
    
    @NotNull(groups=AdditionalValidations.class)
    private String additional;
}
```

Then the @NotNull constraint on the name attribute is NOT performed when submitting the form.

This is caused by the fact that no validations are performed within FormComponent.validateValidators. On the following line both conditions evaluate to false:
`if (isNull == false || validator instanceof INullAcceptingValidator<?>)`

The trick to perform "required" validations is that the PropertyValidator sets the component to required within its onConfigure method:
`this.component.setRequired(true);`

However, this is only performed when isRequired() evaluates to true, which is NOT the case in my example. This is related to issue 1:
When the Default validation group is provided explicitly, isRequired returns the wrong result.

Issue 2 has to do with the required flag only being set once by the PropertyValidator:

Given the example model object above. Assume that when I create the form components I don't set them required explicitly (formComponent.setRequired(true), but I rather let it up to the PropertyValidator because I want to describe my validations on one place, the model object. So, when I initially create the PropertyValidator the "additional" model object attribute is NOT set to required within PropertyValidator.onConfigure, because the validation groups don't match. Then when I would dynamically change the validation groups:

`validationGroups.setObject(new Class[]{Default.class, AdditionalValidations.class});`

Then subsequently a required/NotNull validation should be performed for the "additional" attribute when submitting the form. However, it is not performed because within PropertyValidator.onConfigure it is only set to required once, and not updated at a later stage.


martin-g:
@paulspeijers Please close this PR and open a new one that uses the correct branches.
This PR tries to merge from `master` to `wicket-7.x` and this leads to hundreds of modified files.
Thank you!

PR #218, opened 2017-04-10
[WICKET-6354] string constants and references are added for all versi…

solomax:
…ons of jquery


sebfz1:
There is/was already a `VERSION_2` in `DynamicJQueryResourceReference`.
I think only one should be kept...

sebfz1:
My bad, you used it! ;)
Then opened question: why not centralize all in `JQueryResourceReference`

solomax:
It actually was my plan
But I was trying to keep changes minimal, hoping wicket team can polish it
I guess some test that ensures VERSION_* is updated with jquery update are required ....

PR #219, opened 2017-04-12
[WICKET-6355] It is now possible to set fileName to FileSystemResource

solomax:


wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -68,17 +68,19 @@ public FileSystemResource()
...
 	 * Creates a resource response based on the given attributes
 	 * 
 	 * @param path
 	 *            the path to create the resource response with
+	 * @param fileName
+	 *            fileName to set, path.getFileName() will be used in case null passed
 	 * @return the actual resource response x
 	 */
-	protected ResourceResponse createResourceResponse(Path path)
+	protected ResourceResponse createResourceResponse(Path path, String fileName)
```
martin-g: Do we really need this extra parameter here for the fileName ?
IMO by default it should do: `if (fileName != null) resourceResponse.setFileName(path.getFileName().toString());`
If the application wants to override it then do something like:
```java
ResourceResponse rr = super.createResourceResponse(path);
rr.setFileName("custom.name");
return rr;
```

Also I think passing `Attributes` as a parameter would be useful!

wicket-core/src/main/java/org/apache/wicket/resource/FileSystemResource.java
```diff
@@ -95,7 +95,9 @@ protected ResourceResponse createResourceResponse(Path path, String fileName)
 			resourceResponse.setContentType(getMimeType());
 			resourceResponse.setAcceptRange(ContentRangeType.BYTES);
 			resourceResponse.setContentLength(size);
-			resourceResponse.setFileName(fileName == null ? path.getFileName().toString() : fileName);
+			if (path != null && path.getFileName() != null) {
```
klopfdreh: I think we don't need to check *path != null* because it is checked before (and a WicketRuntimeException is thrown if it is null) - everything else looks good.

solomax: My bad, will fix

PR #222, opened 2017-06-22
fix Sesssion invalidate method dose not work after deserialisation

n1330:
Wicket 6.x fails Session.invalidateNow() after deserialisation (For example, when using Oracle Coherence).
Session.destroy() uses sessionStore. But it is null after deserialisation because sessionStore is transient.


martin-g:
https://issues.apache.org/jira/browse/WICKET-6409

martin-g:
Thank you, @n1330 !

PR #223, opened 2017-07-25
WICKET-6427 Fire an event once all ajax timers are registered

Jezza:
Fire Wicket.Event.Topic.AJAX_TIMERS_BOUND event once all Wicket.Timer.set() calls are executed

### DIFF
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
@@ -56,142 +56,6 @@
 	}
 
 	/**
-	 * Supporting code for getting mouse move and mouse up events from iframes.
-	 * The problem when dragging a div with an iframe is that when the mouse cursor
-	 * gets over an iframe, all mouse events are received by the iframe's document. (IE and FF)
-	 *
-	 * This code can recursively traverse all iframes in document and temporarily forward
-	 * events from their documents to parent document.
-	 */
-	Wicket.Iframe = {
-
-		/**
-		 * Returns the horizontal position of given element (in pixels).
-		 */
-		findPosX: function(e) {
-			if (e.offsetParent) {
-				var c = 0;
-				while (e) {
-					c += e.offsetLeft;
-					e = e.offsetParent;
-				}
-				return c;
-			} else if (e.x) {
-				return e.x;
-			} else {
-				return 0;
-			}
-		},
-
-		/**
-		 * Returns the vertical position of given element (in pixels).
-		 */
-		findPosY: function(e) {
-			if (e.offsetParent) {
-				var c = 0;
-				while (e) {
-					c += e.offsetTop;
-					e = e.offsetParent;
-				}
-				return c;
-			} else if (e.y) {
-				return e.y;
-			} else {
-				return 0;
-			}
-		},
-
-		/**
-		 * Forwards the events from iframe to the parent document (works recursively).
-		 * @param {Document} doc - document to which the events will be forwarded
-		 * @param {HTMLElement} iframe - source iframe
-		 * @param {Array} revertList - list to which altered iframes will be added
-		 */
-		forwardEvents: function(doc, iframe, revertList) {
-			try {
-				var idoc = iframe.contentWindow.document;
-				idoc.old_onmousemove = idoc.onmousemove;
-				idoc.onmousemove = function(evt) {
-					if (!evt) {
-						evt = iframe.contentWindow.event;
-					}
-					var e = {};
-
-					var dx = 0;
-					var dy = 0;
-					if (Wicket.Browser.isIELessThan11() || Wicket.Browser.isGecko()) {
-						dx = Wicket.Window.getScrollX();
-						dy = Wicket.Window.getScrollY();
-					}
-
-					e.clientX = evt.clientX + Wicket.Iframe.findPosX(iframe) - dx;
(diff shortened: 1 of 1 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/head/ResourceAggregator.java
```diff
@@ -353,6 +353,7 @@ private void renderCombinedEventScripts()
 		}
 		if (combinedScript.length() > 0)
 		{
+			combinedScript.append("\nWicket.Event.publish(Wicket.Event.Topic.AJAX_TIMERS_BOUND);");
```
martin-g: How do you know that any timers are in the response ?
It looks to me that this event is published no matter whether there are timer behaviors in use or not.

Jezza: It follows the same logic as the [AJAX_HANDLERS_BOUND](https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/markup/head/ResourceAggregator.java#L339) event.

Which is always fired if there's `OnDomReadyHeaderItem`s.

If the event name is confusing, then I'd recommend changing both of them to suit their actual function.

Jezza: The events themselves are closer to post-`load`s or post-`domready`s

martin-g: Right!
The names are confusing!
They should use `load` and `domready` in their names to be more clear.
What about `Wicket.Event.Topic.ONDOMREADY_EVENTS_BOUND` and `...ONLOAD_EVENTS_BOUND` ?

Jezza: I'd be happy to change it to that.
Should we offer some level of backwards compat, and alias `AJAX_HANDLERS_BOUND`?

martin-g: Yes, I think we should publish both for 8.x and remove `AJAX_HANDLERS_BOUND` for 9.x.
We can also log a warning when `AJAX_HANDLERS_BOUND` published to a subscriber.

Jezza: Also, I noticed that it doesn't work for header items added via the `AjaxRequestTarget`.
Should that be a concern?

Jezza: That actually might just be my code specifically. I'll have to take a look tomorrow.

Jezza: There we go, I pushed the latest state of the branch.

wicket-core/src/main/java/org/apache/wicket/markup/head/ResourceAggregator.java
```diff
@@ -336,7 +336,7 @@ private void renderCombinedEventScripts()
 		}
 		if (combinedScript.length() > 0)
 		{
-			combinedScript.append("\nWicket.Event.publish(Wicket.Event.Topic.AJAX_HANDLERS_BOUND);");
+			combinedScript.append("\nWicket.Event.publish(Wicket.Event.Topic.AJAX_ONDOMREADY_EVENTS_BOUND);");
```
martin-g: I think we should keep both here, to be backward compatible.

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-event-jquery.js
```diff
@@ -277,6 +277,10 @@
 			* @param topic {String} - the channel name for which all subscribers will be notified.
 			*/
 			publish: function (topic) {
+				if (topic === Topic.AJAX_HANDLERS_BOUND) {
+            		console.warn("As of Wicket 8, Topic.AJAX_HANDLERS_BOUND has been deprecated in favour of Topic.AJAX_ONDOMREADY_EVENTS_BOUND and will be removed in Wicket 9.")
+            		topic = Topic.AJAX_ONDOMREADY_EVENTS_BOUND;
+            	}
```
martin-g: To be backward compatible we will have to still publish this event together with the new one.
This check and warning should be moved to `#subscribe()`.

Jezza:
Pushed the latest changes.

Though that does spring a question to mind, the events are only published if there's `OnDomReady` or `OnLoad` events, so should we just always include them?

svenmeier:
>should we just always include them

Good question. IMHO it would make sense to always fire these 'new' events.

Jezza:
Alrighty, preliminary check with that in place:
It breaks 214 tests, so it's gonna take a bit to update all of the tests.

Jezza:
I've pushed the state so you can take a look as I'm working on updating the tests, so I wouldn't recommend accepting the PR just yet.

Jezza:
Should I take the opportunity to update some of the tests, such as `AttributeModifierComponentTest`.
It seems to use an old way of verifying rendered content.
I could migrate it over to the, seemingly, standardised `_expected.html` variant.

Jezza:
I'm moving my branches around, so I guess github spotted that it's no longer on my master branch.

I, personally, don't have any need of this ticket anymore, and honestly, there's a lot of work to do.
I think I got, maybe, 25% of the way through the tests.
I did have to change all of them, so I think unless someone wants it, I'll just let it lie.

PR #224, opened 2017-07-27
Fixed WicketTester to detect components in enclosure when doing isComponentOnAjaxResponse.

disblader:
When a component is in a wicket:enclosure and is then correctly re-rendered using ajax, WicketTester seemed to not be able to detect that the component was in the ajax response in the isComponentOnAjaxResponse method. 

I added additional functionality where isComponentOnAjaxResponse tries to find an enclosure whose child is the given component, and if one is found, it recurses into isComponentOnAjaxResponse passing the enclosure as an argument.

In order not to duplicate logic detecting when an enclosure's child is a given component, I made AjaxEnclosureListener's isControllerOfEnclosure public and static.

Accompanies a test verifying the behavior. I'm not sure what the convention is like for testing new features was, so currently there's only one test. Do let me know if additional tests are needed.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/util/tester/BaseWicketTester.java
+++ b/wicket-core/src/main/java/org/apache/wicket/util/tester/BaseWicketTester.java
@@ -2325,26 +2325,24 @@ public class BaseWicketTester
 		failMessage = "Component wasn't found in the AJAX response. " + componentInfo;
 		result = isTrue(failMessage, isComponentInAjaxResponse);
 
-		if (result.wasFailed()){
-			// Check if the component has been included as part of an enclosure render
-			Enclosure enclosure = getLastRenderedPage().visitChildren(Enclosure.class, (Enclosure enc, IVisit<Enclosure> visit) -> {
-				if (AjaxEnclosureListener.isControllerOfEnclosure(component, enc)){
-					visit.stop(enc);
-				}
-			});
-
-			if (enclosure != null){
-				failMessage = "Component's enclosure was not found in the AJAX response. " + enclosure;
-				boolean isEnclosureInAjaxResponse = !isComponentOnAjaxResponse(enclosure).wasFailed();
-				return isTrue(failMessage, isEnclosureInAjaxResponse);
-			} else {
-				return result;
-			}
-
-		} else {
+		if (!result.wasFailed()) {
 			return result;
 		}
 
+		// Check if the component has been included as part of an enclosure render
+		Enclosure enclosure = getLastRenderedPage().visitChildren(Enclosure.class, (Enclosure enc, IVisit<Enclosure> visit) -> {
+			if (AjaxEnclosureListener.isControllerOfEnclosure(component, enc)){
+				visit.stop(enc);
+			}
+		});
+
+		if (enclosure == null) {
+			return result;
+		}
+
+		failMessage = "Component's enclosure was not found in the AJAX response. " + enclosure;
+		boolean isEnclosureInAjaxResponse = !isComponentOnAjaxResponse(enclosure).wasFailed();
+		return isTrue(failMessage, isEnclosureInAjaxResponse);
 
 	}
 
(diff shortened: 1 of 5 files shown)


disblader:
Also, I used `Enclosure` as opposed to `InlineEnclosure` even thought `AjaxEnclosureListener` uses `InlineEnclosure` because I wasn't sure exactly what the difference was and just wanted to be safe. Should it be changed to `InlineEnclosure`?

wicket-core/src/main/java/org/apache/wicket/util/tester/BaseWicketTester.java
```diff
@@ -2321,7 +2323,23 @@ public Result isComponentOnAjaxResponse(final Component component)
 		boolean isComponentInAjaxResponse = ajaxResponse.matches("(?s).*<component id=\"" +
 			markupId + "\"[^>]*?>.*");
 		failMessage = "Component wasn't found in the AJAX response. " + componentInfo;
-		return isTrue(failMessage, isComponentInAjaxResponse);
+		result = isTrue(failMessage, isComponentInAjaxResponse);
+
+		// Check if the component has been included as part of an enclosure render
+		Enclosure enclosure = getLastRenderedPage().visitChildren(Enclosure.class, (Enclosure enc, IVisit<Enclosure> visit) -> {
```
martin-g: If `result` is true then there is no need to check inside Enclosure.

wicket-core/src/main/java/org/apache/wicket/util/tester/BaseWicketTester.java
```diff
@@ -2321,7 +2323,23 @@ public Result isComponentOnAjaxResponse(final Component component)
...
+			if (AjaxEnclosureListener.isControllerOfEnclosure(component, enc)){
+				visit.stop(enc);
+			}
+		});
+
+		if (enclosure != null){
+			failMessage = "Component's enclosure was not found in the AJAX response. " + enclosure.toString();
```
martin-g: No need to call `.toString()` explicitly.

martin-g:
`Enclosure` handles `<wicket:enclosure>`.
`InlineEnclosure` handles `<div wicket:enclosure="...">`

martin-g:
Thanks for the pull request!
We will create a ticket in JIRA for the changelog and merge it soon!

disblader:
Awesome! Thanks 👍

wicket-core/src/main/java/org/apache/wicket/util/tester/BaseWicketTester.java
```diff
@@ -2325,21 +2325,27 @@ public Result isComponentOnAjaxResponse(final Component component)
 		failMessage = "Component wasn't found in the AJAX response. " + componentInfo;
 		result = isTrue(failMessage, isComponentInAjaxResponse);
 
-		// Check if the component has been included as part of an enclosure render
-		Enclosure enclosure = getLastRenderedPage().visitChildren(Enclosure.class, (Enclosure enc, IVisit<Enclosure> visit) -> {
-			if (AjaxEnclosureListener.isControllerOfEnclosure(component, enc)){
-				visit.stop(enc);
+		if (result.wasFailed()){
```
bitstorm: To avoid if-else nesting I usually prefer to return immediately:

```
if (!result.wasFailed())
  {
    return result;
   }
```

disblader:
Oh no, accidently moved everything around with my IDE 😢

klopfdreh:
Just do a revert commit and apply the changes you originally wanted to do in a separate one 👍 

I think your IDE is also supporting this.

disblader:
I think I managed to revert the changes, the pull request should not contain only useful changes.

klopfdreh:
Thanks a lot for your effort. @martin-g I saw you did the review already. I created a JIRA ticket: https://issues.apache.org/jira/browse/WICKET-6434.

martin-g:
The PR has been merged to Wicket 8 (master).
@disblader Please close it because we do not own this repo.
Thank you!

PR #226, opened 2017-08-01
Allow to set the rel attribute with CssHeaderItem

klopfdreh:


klopfdreh:
Maybe we can create a ticket for this PR to have something for the version changes file

wicket-core/src/main/java/org/apache/wicket/markup/head/CssHeaderItem.java
```diff
@@ -169,6 +171,39 @@ public static CssReferenceHeaderItem forReference(ResourceReference reference,
 	}
 
 	/**
+	 * Creates a {@link CssReferenceHeaderItem} for the given reference.
+	 *
+	 * <strong>Warning</strong>: the conditional comments don't work when injected dynamically with
```
martin-g: Conditional comments are deprecated since a long time in IE (since 2012, I think).
I think we should not care about them.

martin-g:
Please add a ticket in JIRA. For the changelog. Thanks!

wicket-core/src/main/java/org/apache/wicket/markup/head/CssHeaderItem.java
klopfdreh: I just copied the constructor and added the rel parameter. Thats the reason why it is applied there, too. I would just keep it to be backward compatible - the arguments are all Strings, so I have to add a new one behind. 😞

klopfdreh:
Ok I create a JIRA ticket for this! 😃 

Thanks for reviewing @martin-g

klopfdreh:
Ticket: https://issues.apache.org/jira/browse/WICKET-6433

klopfdreh:
Thanks a lot @martin-g!!

PR #230, opened 2017-08-26
Replace filterPath pattern '/*' with empty string

fschrogl:
A filterPath pattern of '/*' is now replaced with an empty string. This matches the behavior of methods getFilterPathFromAnnotation() and getFilterPathFromConfig().

I'm not sure if it's a bug, but I noticed the different beahvior when initializing Wicket in a Spring Boot application. When using a @WebFilter("/*") annotation or the init parameter filterMappingUrlPattern="/*" Wicket works as expected, but when doing the same programmatically with wicketFilter.setFilterPath("/*") the path gets internally rewritten to "*/" and Wicket than only handles HTTP requests to the root path.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/protocol/http/WicketFilter.java
+++ b/wicket-core/src/main/java/org/apache/wicket/protocol/http/WicketFilter.java
@@ -710,6 +710,11 @@ public class WicketFilter implements Filter
 		}
 		if (filterPath != null)
 		{
+			if ("/*".equals(filterPath))
+			{
+			filterPath = "";
+			}
+			
 			filterPath = canonicaliseFilterPath(filterPath);
 
 			// We only need to determine it once. It'll not change.


martin-g:
bq. wicketFilter.setFilterPath("/") the path gets internally rewritten to "/"

Is something missing here ? The values are the same - `"/"`, so I do not understand what is rewritten.
Can you please provide your Spring Boot configuration related to WicketFilter that breaks your app ? I also have Spring Boot based app and I can try it locally to verify the problem..
Thanks!

fschrogl:
I tried to initialize Wicket using WicketFilter class and three different approaches (one after another). That's where I noticed the different behavior (at least that's how I understand it).

__(1) Using setFilterPath method__

    @Bean
    public FilterRegistrationBean<WicketFilter> initWicketFilterByGetter() {
      WicketFilter wicketFilter = new WicketFilter(new MyWicketApplication());
      // filterPath '/' works for all paths
      // filterPath '/*' only works for root path
      wicketFilter.setFilterPath("/");
      return new FilterRegistrationBean<>(wicketFilter);
    }

__(2) Using init params__

    @Bean
    public FilterRegistrationBean<WicketFilter> initWicketFilterByInitParam() {
      WicketFilter wicketFilter = new WicketFilter(new MyWicketApplication());
      FilterRegistrationBean<WicketFilter> wicketFilterRegistrationBean = new FilterRegistrationBean<>(wicketFilter);
      // Using '/*' works for all paths, using '/' leads to an exception
      wicketFilterRegistrationBean.addInitParameter(WicketFilter.FILTER_MAPPING_PARAM, "/*");
      return wicketFilterRegistrationBean;
    }

__(3) Using `@WebFilter` annotation on a custom filter and `@ServletComponentScan` for Spring Boot__

    // /* works, / doesn't work
    @WebFilter(value = "/*", initParams = @WebInitParam(name = "applicationClassName", value = "package.MyWicketApplication"))
    public static class MyWicketFilter extends WicketFilter {  }

In approach __(1)__ I have to use `/` while in approaches __(2)__ and __(3)__ I have to use `/*` to achieve the same outcome.

klopfdreh:
Just my thoughts:

/* - maps every path - root paths and sub paths to the specific filter
/ - maps only the root path but not every sub path

So from my understanding they aren't the same and shouldn't be mixed up.

https://stackoverflow.com/questions/4140448/difference-between-and-in-servlet-mapping-url-pattern

Maybe we should inspect this a bit more to achieve the right handling.

WDTY?

klopfdreh:
@martin-g what do you thing of extracting the lines:

https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/protocol/http/WicketFilter.java#L583

to 

https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/protocol/http/WicketFilter.java#L599

into a method and call this method within getFilterPath and getFilterPathFromConfig?

solomax:
I guess it can be dropped? :)

PR #232, opened 2017-09-04
Ajax inline enclosure fix

disblader:
A fix for the issue I described in WICKET-6459. 

I've made this for 7.x as that's the version relevant for me, but I'd like it to be propogate to 8.x as well, how would I do that? Should I create a new pull request trying to merge this branch into 8.x?


martin-g:
Thanks, @disblader !
We will cherry-pick it to 8.x once this is merged!

wicket-core/src/test/java/org/apache/wicket/markup/renderStrategy/ChildFirstHeaderRenderStrategyTest.java
```diff
@@ -51,6 +57,39 @@ public void test2() throws Exception
...
+	 * contributions reach the response
+	 *
+	 * WICKET-6459
+	 *
+	 */
+	@Test
+	public void testAjaxAndEnclosures() throws Exception
+	{
+
+		tester.startPage(SimplePage3.class);
```
martin-g: Let's rename the page name to `InlineEnclosureAjaxRenderPage`.
`SimplePage`s are legacy, but for new stuff we should try better! :-)

wicket-core/src/main/java/org/apache/wicket/markup/renderStrategy/AbstractHeaderRenderStrategy.java
```diff
@@ -116,7 +119,32 @@ protected void renderRootComponent(final HtmlHeaderContainer headerContainer,
 		final HeaderStreamState headerStreamState, final Component rootComponent)
 	{
 		headerContainer.renderHeaderTagBody(headerStreamState);
+
 		rootComponent.internalRenderHead(headerContainer);
+
+		// If the root component is an inline enclosure, then we force header render of its controller as well; normally
+		// this would not trigger because the controller is a sibling of the enclosure
+		if (rootComponent instanceof InlineEnclosure) {
```
martin-g: The code looks good to me!
Just let's extract it in its own method to reduce the complexity.
E.g. `renderInlineEnclosure( HtmlHeaderContainer)`

wicket-core/src/test/java/org/apache/wicket/markup/renderStrategy/ChildFirstHeaderRenderStrategyTest.java
disblader: I renamed it to `EnclosureAjaxRenderPage`, since it actually tests for both regular and inline enclosures

martin-g:
For some reason the PR doesn't auto-close itself.
@disblader Please close it since we do not have permissions.
Thank you!

disblader:
Will do!

PR #233, opened 2017-09-10
WICKET-6465 prevent unbound during storeTouchedPages only

svenmeier:
Why so complicated? We just want to prevent valueUnbound() from doing any harm while we're resetting the attribute during storeTouchedPages.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/page/PageStoreManager.java
+++ b/wicket-core/src/main/java/org/apache/wicket/page/PageStoreManager.java
@@ -24,7 +24,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.servlet.http.HttpSessionBindingEvent;
 import javax.servlet.http.HttpSessionBindingListener;
@@ -96,13 +95,18 @@ public class PageStoreManager extends AbstractPageManager
 
 		/**
 		 * A flag indicating whether this session entry is being re-set in the Session.
+		 * <p>
 		 * Web containers intercept {@link javax.servlet.http.HttpSession#setAttribute(String, Object)}
 		 * to detect changes and replicate the session. If the attribute has been already
 		 * bound in the session then it will be first unbound and then re-bound again.
 		 * This flag helps us to detect <em>update</em> operations and skip the default behavior
 		 * of {@link #valueUnbound(HttpSessionBindingEvent)}.
 		 */
-		private final AtomicBoolean storingTouchedPages = new AtomicBoolean(false);
+		private final ThreadLocal<Boolean> storingTouchedPages = new ThreadLocal<Boolean>() {
+			protected Boolean initialValue() {
+				return Boolean.FALSE;
+			};
+		};
 
 		/**
 		 * Construct.
--- a/wicket-core/src/main/java/org/apache/wicket/page/PageStoreManager.java
+++ b/wicket-core/src/main/java/org/apache/wicket/page/PageStoreManager.java
@@ -95,14 +95,14 @@ public class PageStoreManager extends AbstractPageManager
 		private transient List<Object> afterReadObject;
 
 		/**
-		 * A flag indicating whether this session entry has been re-set in the Session.
+		 * A flag indicating whether this session entry is being re-set in the Session.
 		 * Web containers intercept {@link javax.servlet.http.HttpSession#setAttribute(String, Object)}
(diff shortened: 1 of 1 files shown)


bitstorm:
We should consider method Session.destroy() to remove pages when session expires. See my comment at https://issues.apache.org/jira/browse/WICKET-6465

martin-g:
When the session expires or is invalidated via API call then `#valueUnbound()` is called and this clears the storages.
This is how it worked last time I worked in this area of the code.

papegaaij:
Indeed, Sesison.destroy() is only called under certain conditions, and in those conditions, valueUnbound() will also be called.

bitstorm:
I see. thank you for the clarification. Still, I'd like to evaluate more structural solutions, at least for Wicket 8. A good start point might be HttpSessionStore.SessionBindingListener#valueUnbound. 
It's already used to call Session#onInvalidate and we can add Session#clear() after it.

svenmeier:
Currently HttpSessionStore and PageStoreManager each use their own instance of HttpSessionBindingListener.
Yes, we might be able to unify these. But with a simple fix we don't introduce even more - possibly faulty - changes.

It was unfortunate that we didn't consider all possible callbacks from Servlet containers when this problem started with WICKET-6356.
IMHO we should go with the simplest fix: Prevent anything bad from happening, when re-setting the session attribute:
- clustering will work (because the attribute is re-set)
- it doesn't matter whether the container calls valueBound() or not
- if the container calls valueUnbound(), it does nothing bad during storeTouchedPages()

Andrea, what do you think?

BTW do we have to take care of concurrent access to the SessionEntry? I see that Martin used an AtomicBoolean, but how does that help if there are two request simultaneously storing touched pages?

bitstorm:
I'm fine with this quick fix, but I warmly suggest to improve it after it has been applied, at least in master branch.

About concurrent access afaik we do have to take care of concurrency. At the moment we don't have any guarantee that setSessionAttribute saves the entry for the last request submitted (for the same session).

papegaaij:
I don't see why the current implementation uses `AtomicBoolean`. The current implementation could just as well use a normal boolean. If we need to take concurrent access into account (and I think we do), we should probably use a `ThreadLocal`. Of course this assumes that the servlet container will call (un)bound from the same thread, but the current implementation already has that assumption. If a container decides to make the calls async, there is no way of telling if it will happen between the setting and clearing of the boolean.

bitstorm:
I agree about the use of a normal boolean, but I'm not sure I've understood your idea about ThreadLocal. What I would like to do is to prevent race conditions inside storeTouchedPages. I would do it using synchronized on entry object:

```
protected void storeTouchedPages(final List<IManageablePage> touchedPages)
{
	if (!touchedPages.isEmpty())
	{
		SessionEntry entry = getSessionEntry(true);
		synchronized (entry) 
		{
			entry.setSessionCache(touchedPages);
			for (IManageablePage page : touchedPages) 
			{
				// WICKET-5103 use the same sessionId as used in SessionEntry#getPage()
				pageStore.storePage(entry.sessionId, page);
			}
			entry.updating.set(true);
			setSessionAttribute(getAttributeName(), entry);
		}
	}
}
```
In this way two possible concurrent requests for the same session would execute storeTouchedPages one at a time.

papegaaij:
We don't care about concurrent calls to valueUnbound, that's fine. All code in that method is thread safe. What we don't want is multiple calls to storeTouchedPages and valueUnbound (session expiry) to interfere with each other. Synchronizing on the entry will not help as session expiry is probably done from a different thread.

bitstorm:
> We don't care about concurrent calls to valueUnbound, that's fine. All code in that method is thread safe. 

I agree

> What we don't want is multiple calls to storeTouchedPages and valueUnbound (session expiry) to interfere with each other. Synchronizing on the entry will not help as session expiry is probably done from a different thread.

Yes, synchronizing is meant to prevent problems for multiple invocations of storeTouchedPages for the same session, valueUnbound is unaffected. The boolean value should be enough to tell if valueUnbound is invoked as result of session expiration.

papegaaij:
No, a boolean won't work. Suppose the following happens: A thread calls storeTouchedPages, setting the boolean to true. At the same time, another thread unbinds the session, calling valueUnbound. This will leak the page store. We only want valueUnbound to be ignored when its a side effect of calling storeTouchedPages. This is local to the thread calling the method, hence ThreadLocal.

bitstorm:
Got it. That's way I hope for a future improvement of this boolean-based solution :-).

svenmeier:
Agreed, a ThreadLocal makes sense.

papegaaij:
Yes, this is what I had in mind. I'm ok to merge this in 7.x and 8.

bitstorm:
Don't forget to fix brackets indentation before commit

PR #238, opened 2017-10-10
WICKET-6479 keep window name

svenmeier:
If the page was not yet rendered into any window, just keep the window name as is, instead of changing it for each new page.


svenmeier:
With this change AjaxNewWindowNotifyingBehavior keeps the window's name if it is already present.

All successively rendered pages will be bound to an identical name and window.name will stay the same.
Thus going back to a previous page instance will not report a new window.

svenmeier:
With this change AjaxNewWindowNotifyingBehavior keeps the window's name if it is already present.

All successively rendered pages will be bound to an identical name and window.name will stay the same.
Thus going back to a previous page instance will not report a new window.

PR #239, opened 2017-10-20
WICKET-6055 non-blocking lazy loading volume II

svenmeier:
Building on #151 I've built this alternative implementation with the following advantages:

- lazy loading can now *start* on an Ajax request too
- much simpler without using the event bus
- the timer adjusts to the minimum of the preferred timeout of all LazyLoadPanels
- reworded methods and improved JavaDoc


svenmeier:
Hm, github screw up on me, this is the wrong pull-request :/.

PR #243, opened 2017-11-08
Added Czech localization for Wizard

vrozkovec:


svenmeier:
Could you provide an *.utf8.properties instead? Czech needs a lot of unicode escapes for ISO-8859-1 :P

vrozkovec:
Done

svenmeier:
Thanks.

PR #244, opened 2017-11-09
[WICKET-6491] ajax download seems to work under both IE11 and Edge

solomax:
I would appreciate if someone can review my changes

### DIFF
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/AjaxDownloadBehavior.java
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/AjaxDownloadBehavior.java
@@ -48,11 +48,11 @@ import com.github.openjson.JSONObject;
  * Download resources via Ajax.
  * <p>
  * Usage:
- *
+ * 
  * <pre>
  * final AjaxDownload download = new AjaxDownload(resource);
  * add(download);
- *
+ * 
  * add(new AjaxButton("download")
  * {
  * 	&#64;Override
@@ -66,7 +66,7 @@ import com.github.openjson.JSONObject;
  * <p>To set the name of the downloaded resource make use of
  * {@link org.apache.wicket.request.resource.ResourceStreamResource#setFileName(String)} or
  * {@link org.apache.wicket.request.resource.AbstractResource.ResourceResponse#setFileName(String)}</p>
- *
+ * 
  * @author svenmeier
  * @author Martin Grigorov
  * @author Maxim Solodovnik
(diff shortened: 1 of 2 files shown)


wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/wicket-ajaxdownload.js
```diff
@@ -108,29 +108,32 @@
 						var disposition = xhr.getResponseHeader("Content-Disposition");
 						if (disposition) {
 							var matches = /filename[^;=\n]*=(([""]).*?\2|[^;\n]*)/.exec(disposition);
-							if (matches != null && matches[1]) {
+							if (matches !== null && matches[1]) {
```
martin-g: Let's add this .js file to the ones checked by JSLint in `testing/js-tests/Gruntfile.js

solomax: Wasn't aware of this file
Just have checked `../../wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/wicket-ajaxdownload.js` is already added

martin-g: Strange that it didn't break the build due to `!=`.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/AjaxDownloadBehavior.java
```diff
@@ -323,10 +323,30 @@ private String getName()
 	}
 
 	/**
+	 * Controls if anti cache parameter should be added to the URL or not
+	 *
+	 * @return {@code true} to add the anti cache request parameter, {@code false} - otherwise
+	 */
+	protected boolean shouldAddAntiCacheParameter() {
```
svenmeier: What is this needed for? The resource controls the caching.

solomax: During my tests I found the issue: subsequent download was cached. I have implemented this anti-cache, same as in wicket Image.java

svenmeier: What is this good for? The resource can control caching - I don't see the necessity for an antiCache parameter (as ResourceLink doesn't need have it either).

solomax: I'll double-check (modify my tests) and remove it
I thought it would save some lines of code if anti-cache will be on by default

solomax:
@svenmeier I reverted my changes to AjaxDownloadBehavior.java, thanks for catching this :)

solomax:
Is it OK to merge this PR?

PR #246, opened 2017-11-19
WICKET-6499 Support Bean Validation 2.0

theigl:
This PR adds support for Bean Validation 2.0 to `wicket-bean-validation`.

It extends scanning of `@NotNull` annotations to `@NotEmpty`  and `@NotBlank` and supports registering further annotations that should be treated as not-null constraints to guard against future changes.

See https://issues.apache.org/jira/browse/WICKET-6499


wicket-bean-validation/src/main/java/org/apache/wicket/bean/validation/PropertyValidator.java
```diff
@@ -208,24 +206,28 @@ public void detach(Component component)
...
+			notNullAnnotations);
 
 		if (constraints.isEmpty())
 		{
 			return false;
 		}
 
-		HashSet<Class<?>> validatorGroups = new HashSet<Class<?>>();
+		HashSet<Class<?>> validatorGroups = new HashSet<>();
 		validatorGroups.addAll(Arrays.asList(getGroups()));
```
solomax: I would use `Set` here

wicket-bean-validation/src/main/java/org/apache/wicket/bean/validation/PropertyValidator.java
```diff
@@ -208,24 +206,28 @@ public void detach(Component component)
...
 
-		for (NotNull constraint : constraints)
+		for (Annotation constraint : constraints.keySet())
 		{
```
solomax: `entrySet()` can be used here to avoid `double search`

theigl:
@solomax: Thanks for the quick feedback. I adjusted the PR accordingly.

wicket-bean-validation/src/main/java/org/apache/wicket/bean/validation/PropertyValidator.java
```diff
@@ -208,24 +206,29 @@ public void detach(Component component)
...
-		HashSet<Class<?>> validatorGroups = new HashSet<Class<?>>();
+		Set<Class<?>> validatorGroups = new HashSet<>();
 		validatorGroups.addAll(Arrays.asList(getGroups()));
 
-		for (NotNull constraint : constraints)
+		for (Map.Entry<Annotation, ConstraintDescriptor<?>> entry : constraints.entrySet())
 		{
-			if (canApplyToDefaultGroup(constraint) && validatorGroups.isEmpty())
+			final Annotation annotation = entry.getKey();
+			ConstraintDescriptor<?> constraintDescriptor = constraints.get(annotation);
```
solomax: you can use `entry.getValue()` here

theigl: Sorry, totally overlooked that. Changed.

wicket-bean-validation/src/main/java/org/apache/wicket/bean/validation/PropertyValidator_fr.properties.xml
```diff
@@ -26,6 +26,8 @@
 	<entry key="javax.validation.constraints.Max.message">Le champ '${label}' doit être inférieur ou égal à ${value}</entry>
 	<entry key="javax.validation.constraints.Min.message">Le champ '${label}' doit être supérieur ou égal à ${value}</entry>
 	<entry key="javax.validation.constraints.NotNull.message">Le champ '${label}' ne doit pas être null</entry>
+	<entry key="javax.validation.constraints.NotBlank.message">Le champ '${label}' ne peut pas être vide</entry>
```
theigl: A couple of other annotations have been added to [BV 2.0](http://beanvalidation.org/2.0/) as well:

> new built-in constraints: @Email, @NotEmpty, @NotBlank, @Positive, @PositiveOrZero, @Negative, @NegativeOrZero, @PastOrPresent and @FutureOrPresent

@solomax: Should we add translations for these new constraints?

solomax: +1 for new labels, I'll add `_ru` localization ...

theigl: Added properties for English and French. My French is a bit rusty, but the translations should be ok.

theigl: We should probably copy the properties from the HV implementation:

https://github.com/hibernate/hibernate-validator/tree/master/engine/src/main/resources/org/hibernate/validator

solomax: It is APL licensed so, I guess, we can copy
I would better use existing ones, but not sure how is this possible :(

theigl: The existing ones could probably be loaded via a custom `IStringResourceLoader` but this is out of the scope of this PR.

@solomax: What else do you want me to do for this PR so it can be merged?

solomax: OK with me :)

svenmeier:
Actually I'm wondering whether we should just add another method to ITagModifier (possibly with better names):

```
public interface ITagModifier<T extends Annotation>
{
  void modify(FormComponent<?> component, T annotation);
	
  void modify(FormComponent<?> component, ComponentTag tag, T annotation);
}
```

This way we can use the identical lookup BeanValidationConfiguration#getTagModifier() to alter the component itself from #onConfigure() and the component tag from #onComponentTag().

All that's required for the new annotations is then:

```
register(Size.class, new SizeTagModifier());
register(NotNull.class, new RequiredTagModifier<NotNull>());
register(NotEmpty.class, new RequiredTagModifier<NotEmpty>());
register(NotBlank.class, new RequiredTagModifier<NotBlank>());
```

theigl:
@svenmeier: This makes a lot of sense. I'm not sure however how much time I'll find the next weeks to work on this.

I suggest the following approach:

1. I'll look into this this week and try to rework the code as you suggested.
2. If It takes too much time and I can't finish it, I vote for merging a very simple version of this PR that only applies the changes required for making `wicket-bean-validation` work with BV 2.0. I'll remove the API changes I introduced in `BeanValidationConfiguration` and hard-code the list of built-in annotations that imply a not-null constraint. This way, we do not introduce any new methods that we would later remove if your solution is implemented.

WDYT?

svenmeier:
That's a very good approach, thank you.

theigl:
I pushed the simplified implementation without API changes in https://github.com/apache/wicket/pull/246/commits/145e12e54c744b785f64ccf7acd5b8b0efda41e7.

wicket-bean-validation/src/main/java/org/apache/wicket/bean/validation/PropertyValidator.java
```diff
@@ -1,14 +1,12 @@
 package org.apache.wicket.bean.validation;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
+import java.lang.annotation.Annotation;
+import java.util.*;
```
martin-g: No `*` imports please

theigl: Changed

solomax:
I believe this can be closed
@theigl could you please close this PR?

PR #249, opened 2017-12-07
independent component renderer

svenmeier:
ComponentRenderer makes it easy to render components, but it still needs the thread to be attached to a WebApplication's ThreadContext: I've seen projects use WicketTester for that or rolling their own solution.

IMHO we could improve ComponentRender to officially support independent rendering of components.

WDYT?


solomax:
We are currently using WicketTester for this, I'll try this later tonight and will give huge +1 :))

wicket-core/src/main/java/org/apache/wicket/core/util/string/ComponentRenderer.java
```diff
@@ -16,35 +16,321 @@
...
+		public Set<BindListener> getBindListeners()
+		{
+			return null;
+		}
+	}
+
 	/**
 	 * Collects the html generated by the rendering of a page.
+	 * <p>
+	 * Important note: Must be called on a thread already processing a {@link WebApplication}!
```
martin-g: What does it mean "a thread processing an application" ?
Processing a request is clear, but application is not.

solomax: `Application.get()` will fail in case current thread doesn't owns Application ....

Why `Application.get()` is being used here, and not `this.application` ?

svenmeier: @solomax this is the 'old' static method that reuses whatever application is bound to the thread
@martin-g what about 'a thread bound to a wicket ThreadContext'

solomax: ups `static`, sorry :(

martin-g: > what about 'a thread bound to a wicket ThreadContext'

Yes, this is more clear, to me at least!

PR #250, opened 2017-12-14
WICKET-6503 render cleanup and ajax

svenmeier:
This is a combined effort to:
- align Ajax requests with the way components are prepared in normal request, i.e. beforeRender is called on *all* components, before any of them is rendered
- unify rendering of components separately as part of a page only vs. a whole page
- simplify handling of feedbacks  ... this was a real mess :(
- improve performance by prevent unnecessary traversal of the component hierarchy
- cleanup component's internal methods and flags

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -17,11 +17,11 @@
 package org.apache.wicket;
 
 import java.io.Serializable;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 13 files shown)


svenmeier:
This does *not* solve (yet) one aspect of WICKET-6503: should components be allowed to be added to an AjaxRequestTarget *after* components have already been rendered ("ART is already frozen").

svenmeier:
All test pass, there is a single change in FeedbackRenderTest only:

If IFeedbacks are *nested* (e.g. id1 -> id2 -> id3), *they* are prepared for render top-down (i.e. id1 then id2 then id3). This matches they way other components are rendered.

Previously the order was the other way around (i.e. id3 then id2 then id1), which was wrong IMHO.

PR #253, opened 2018-01-02
WICKET-6498 deferred javascript

svenmeier:
If addEventListener is used instead of jQuery  "domready" and "load" events, and OnLoadHeaderItem instead of JavaScriptHeaderItem, all JavaScript resources can be deferred.

Open questions:
- does this solution work reliably on all browsers?
- who decides which JavaScript resources to defer and which not?
- how to handle cases where user code is still using a JavaScriptHeaderItem?


wicket-core/src/main/java/org/apache/wicket/markup/head/OnLoadHeaderItem.java
```diff
@@ -86,8 +86,8 @@ public void render(Response response)
 		CharSequence js = getJavaScript();
 		if (Strings.isEmpty(js) == false)
 		{
-			JavaScriptUtils.writeJavaScript(response, "Wicket.Event.add(window, \"load\", " +
-				"function(event) { " + js + ";});");
+			JavaScriptUtils.writeJavaScript(response,
+				"var f = function() {" + js + ";};\nif ('loaded' === document.readyState) f(); else window.addEventListener('load', f);");
```
klopfdreh: Is loaded the correct state for sure?

https://developer.mozilla.org/de/docs/Web/API/Document/readyState

here "complete" is mentioned. I also remember that there were some other pitfalls with the readyState for some browsers.

This german site mentions that "loaded" is only available in IE.

https://wiki.selfhtml.org/wiki/JavaScript/DOM/Document/readyState

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
```diff
@@ -106,8 +107,7 @@ public void renderHead(final Component component, final IHeaderResponse response
 		RequestCycle requestCycle = component.getRequestCycle();
 		Url baseUrl = requestCycle.getUrlRenderer().getBaseUrl();
 		CharSequence ajaxBaseUrl = Strings.escapeMarkup(baseUrl.toString());
-		response.render(JavaScriptHeaderItem.forScript("Wicket.Ajax.baseUrl=\"" + ajaxBaseUrl
-			+ "\";", "wicket-ajax-base-url"));
```
klopfdreh: Mh I didn't look into the implementation but is wicket-ajax-base-url not used to be applied to the OnLoadHeaderItem, too. I could imagine that the information gets lost in this case.

wicket-core/src/main/java/org/apache/wicket/resource/CoreLibrariesContributor.java
```diff
@@ -63,8 +64,7 @@ public static void contributeAjax(final Application application, final IHeaderRe
 		if (debugSettings.isAjaxDebugModeEnabled())
 		{
 			response.render(JavaScriptHeaderItem.forReference(jsLibrarySettings.getWicketAjaxDebugReference()));
-			response.render(JavaScriptHeaderItem.forScript("Wicket.Ajax.DebugWindow.enabled=true;",
-				"wicket-ajax-debug-enable"));
```
klopfdreh: Same as wicket-ajax-base-url

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
klopfdreh: Ah this is only for the id attribute - so ignore this comment.

wicket-core/src/main/java/org/apache/wicket/resource/CoreLibrariesContributor.java
klopfdreh: Also ignore this. :-)

wicket-core/src/main/java/org/apache/wicket/markup/head/OnLoadHeaderItem.java
klopfdreh: Here I suggest to change it to „complete“

svenmeier: You're right, "complete" it should be.

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
solomax: this missing `id` attribute breaks `org.apache.wicket.util.tester.BaseWicketTester.getWicketAjaxBaseUrlEncodedInLastResponse()` ....
Will set it back if there are no better solutions ...

svenmeier: It's not missing, it's just on a different location (see below), where getWicketAjaxBaseUrlEncodedInLastResponse() doesn't find it - we have to change that method accordingly.

solomax: previous code set the id attribute, so the check was trivial (node.attr === 'wicket-ajax-base-url'). Current implementation adds text content. which have to be parsed, maybe it worth to set attribute back to avoid parsing the text?

wicket-core/src/main/java/org/apache/wicket/markup/head/OnDomReadyHeaderItem.java
```diff
@@ -86,8 +86,8 @@ public void render(Response response)
 		CharSequence js = getJavaScript();
 		if (Strings.isEmpty(js) == false)
 		{
-			JavaScriptUtils.writeJavaScript(response, "Wicket.Event.add(window, \"domready\", " +
-				"function(event) { " + js + ";});");
+			JavaScriptUtils.writeJavaScript(response,
+				"var f = function() {" + js + ";};\nif ('loading' !== document.readyState) f(); else document.addEventListener('DOMContentLoaded', f);");
```
solomax: I would like to change this code to be:
```
{ //scope
    const f = function() {" + js + ";}; ....
}
```

This way multiple `OnDomReadyHeader` items will work as expected in deferred mode
WDYT?

svenmeier: That does no harm +1

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
svenmeier: Setting the baseUrl has to be deferred until onLoad, otherwise the deferred wicket-ajax-jquery.js hasn't been loaded yet and Wicket.Ajax is not defined.

wicket-core/src/main/java/org/apache/wicket/markup/head/ResourceAggregator.java
```diff
@@ -337,24 +338,27 @@ private void renderCombinedEventScripts()
 		if (combinedScript.length() > 0)
 		{
 			combinedScript.append("\nWicket.Event.publish(Wicket.Event.Topic.AJAX_HANDLERS_BOUND);");
-			getRealResponse().render(
-				OnDomReadyHeaderItem.forScript(combinedScript.append('\n').toString()));
+			getRealResponse().render(OnDomReadyHeaderItem.forScript(combinedScript.append('\n')));
 		}
 
 		combinedScript.setLength(0);
+		String lastId = null;
```
solomax: Would appreciate review on this commit, especially this part ....

svenmeier: That doesn't seem to be reliable: what if multiple items are given an id?

IMHO we shouldn't changed code just to please test code. Couldn't BaseWicketTester just parse the url from the whole JavaScript via a regular expression?

> `.*Wicket\.Ajax\.baseUrl="(.*)";.*`

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/FilteringHeaderResponse.java
```diff
@@ -263,14 +262,12 @@ public void markRendered(Object object)
 			}
 		};
 
-		ResourceAggregator resourceAggregator = new ResourceAggregator(headerRenderer);
-
 		for (HeaderItem curItem : resp)
 		{
-			resourceAggregator.render(curItem);
+			headerRenderer.render(curItem);
```
solomax: Unfortunately with this change `FilteringHeaderResponseTest` fails due to JS dependencies are not being rendered :(((
@svenmeier Could you please take a look at this?

svenmeier: Thanks Maxim, I'll take a look.

solomax:
Thanks a lot @svenmeier!
All tests are green now :)

Shall we merge?

svenmeier:
Nope, not yet please. I have a new idea I want to try out.

solomax:
Sure :) no rush, just would like to check if everyone happy with this PR :)

svenmeier:
Many thanks Maxim for your work on this!

But with #258 we have a much simpler solution now, that should please those that are sceptical about the changes presented here.

If no one objects, I'll close this pull request.

svenmeier:
@solomax I'm sorry for all the hard work you put into fixing the tests, but #258 is much better than this initial solution. Closing.

solomax:
NP, I'll check the tests again  :)

svenmeier:
There aren't many test to check - the new pull request does not change the current Ajax implementation at all :p. IMHO that's a huge advantage.

PR #254, opened 2018-01-02
provide more information on read only LambdaModel setObject exception

andruhon:
It is sometimes hard to find which model is misused as ReadOnly model when it should be a ReadWrite one, because the UnsupportedOperationException doesn't provide enough information (some model of some component doesn't support setObject):
```
java.lang.UnsupportedOperationException: setObject(Object) not supported
    at org.apache.wicket.model.LambdaModel.setObject(LambdaModel.java:53)
    at org.apache.wicket.Component.setDefaultModelObject(Component.java:3122)
    at org.apache.wicket.markup.html.form.FormComponent.setModelObject(FormComponent.java:1579)
    at org.apache.wicket.markup.html.form.FormComponent.updateModel(FormComponent.java:1097)
```

this PR will make the unsupported stack trace to contain more detailed information like this:
`Caused by: java.lang.UnsupportedOperationException: setObject(Object) on Model:classname=[org.apache.wicket.model.PropertyModel]:nestedModel=[Model:classname=[org.apache.wicket.model.Model]:object=[com.company.blanket.BlanketOrderBean@14c8757d]]:expression=[organisationShipTo] is not supported`

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/model/LambdaModel.java
+++ b/wicket-core/src/main/java/org/apache/wicket/model/LambdaModel.java
@@ -156,6 +156,11 @@ public abstract class LambdaModel<T> implements IModel<T>
 			}
 
 			@Override
+			public void setObject(R r) {
+				throw new UnsupportedOperationException("setObject(Object) on " + target + " is not supported");
+			}
+
+			@Override
 			public void detach()
 			{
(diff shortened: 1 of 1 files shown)


martin-g:
Thank you, @andruhon !

PR #255, opened 2018-01-11
Allow multipart submission of non modified FormData with Wicket.Ajax

andruhon:
Wicket is currently does fancy processing of the data and it's not possible to avoid it. This commit allows to submit non processed data, which won't be processed.

This change, for example, will make possible to collect files from the drag and drop event and submit these files as FormData.

A drop file handler example:
```JavaScript
        dropArea.addEventListener("drop", dropHandler);
        function dropHandler(ev) {            
            ev.preventDefault();
            Wicket.Log.info("DragAndDropFileUpload drop " + dropAreaId);
            var dt = ev.dataTransfer;
            var formData = new FormData();
            var files = [];
            var i;
            if (dt.items) {
                // DataTransferItemList
                for (i = 0; i < dt.items.length; i++) {
                    if (dt.items[i].kind == "file") {
                        var f = dt.items[i].getAsFile();
                        Wicket.Log.info("... file[" + i + "].name = " + f.name);
                        files.push(f);
                    }
                }
            } else {
                // DataTransfer
                for (i = 0; i < dt.files.length; i++) {
                    Wicket.Log.info("... file[" + i + "].name = " + dt.files[i].name);
                    files.push(dt.files[i]);
                }
            }

            if (files.length > 0) {
                // formData.append(attrs['f']+"_hf_0","");
                files.forEach(function (f) {
                    formData.append(fileUpload.getAttribute("name"), f)
                });

                // Wicket.Ajax.ajax will simply submit a multipart form into the iframe, which we don't want
                // var xhr = new XMLHttpRequest();
                // xhr.open('POST', attrs['u'] + "&wicket-ajax=true&wicket-ajax-baseurl=" + Wicket.Form.encode(Wicket.Ajax.baseUrl || '.'), true);
                // xhr.send(formData);
                // xhr.addEventListener("load", function() {console.log(arguments)})
                Wicket.Ajax.ajax({
                    u: attrs.u,
                    m: 'POST',
                    mp: true,
                    npd: formData
                });

            } else {
                Wicket.Log.error("No files to upload " + dropAreaId);
            }
        }
```

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -54,7 +54,10 @@
 		isUndef,
 		replaceAll,
 		htmlToDomDocument,
-		nodeListToArray;
+		nodeListToArray,
+		isMultipartAjaxSupported;
+
+	isMultipartAjaxSupported = 'FileReader' in window && 'FormData' in window;
 
 	isUndef = function (target) {
 		return (typeof(target) === 'undefined' || target === null);
@@ -643,7 +646,7 @@
 
 			we.publish(topic.AJAX_CALL_PRECONDITION, attrs);
 
-			if (attrs.mp) { // multipart form. jQuery.ajax() doesn't help here ...
+			if (attrs.mp && !isMultipartAjaxSupported) { // multipart form. jQuery.ajax() doesn't help here ...
 				var ret = self.submitMultipartForm(context);
 				return ret;
 			}
@@ -666,7 +669,16 @@
 			}
 
 			// convert to URL encoded string
-			data = jQuery.param(data);
+			if (attrs.mp && isMultipartAjaxSupported) {
+				// Convert to FormData
+				data = data.reduce(function (f, v) {
+					f.append(v.name, v.value);
+					return f;
+				}, new FormData());
+			} else {
+				// Convert to URL encoded string
+				data = jQuery.param(data);
+			}
 
 			// execute the request
 			var jqXHR = jQuery.ajax({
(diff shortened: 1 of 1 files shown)


andruhon:
A little bit of demo showing what I was trying to achieve:
https://github.com/andruhon/WicketDragAndDropFileAjaxUpload

svenmeier:
Many thanks!

It seems you had to jump through a lot of hoops to make this work with Wicket's Ajax code - 
maybe we can change it to make this easier.
Why not drop the (old) iframe workaround and always use FormData for multipart requests?

I'll give it a try this afternoon.

andruhon:
@svenmeier I would agree that this legacy workaround is not a good thing for Wicket 8. IE 9 share is averagely less than 1% nowdays and keeps dropping. Those who need a legacy support of IE 6 and 7 can use Wicket 6. Thank You. Just let me know If I can help you with this.

andruhon:
@svenmeier I've tried a better approach without modifying the js function interface, it's just using multipart form data when possible, so works in IE9 as it used to work before.

See it in the branch:
https://github.com/andruhon/wicket/tree/try-multipart

Also updated the demo to use this snapshot:
https://github.com/andruhon/WicketDragAndDropFileAjaxUpload/tree/try-multipart
tested in Chrome, FF, IE 11 and in emulated IE9-10

If it looks like decent approach I can update my PR.

svenmeier:
Yes, this is what I came up with yesterday too.
We have to support file inputs in forms too, for that I changed serializeInput() to handle "file" types.

I'd like to drop support for the old iframe solution in submitMultipartForm(). I'll create an issue, let's see what other devs are thinking of this.

andruhon:
@svenmeier I'd say that submitMultipartForm could stay until the IE9 is closer to 0% in stats.

klopfdreh:
Hi,

is this PR reflection a subset of changes of: https://github.com/apache/wicket/pull/256?

If yes we might consider to close this one and continue it in only one PR.

Thanks a lot for your PR @andruhon 👍

PR #256, opened 2018-01-12
WICKET-6517 ajax multipart

svenmeier:
With this change multipart Ajax requests *are* send via Ajax.


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -658,23 +653,36 @@
...
+				}
+			}
 
 			// execute the request
 			var jqXHR = jQuery.ajax({
 				url: attrs.u,
 				type: attrs.m,
 				context: self,
+				processData: wwwFormUrlEncoded,
+				contentType: wwwFormUrlEncoded,
```
solomax: According to the http://api.jquery.com/jquery.ajax/
`contentType` can be string (should be `real` content type) or boolean
Are you sure `undefined` is OK here?

svenmeier: With undefined jQuery will just use the default, i.e. a www-form-url-encoded request.

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -643,11 +643,6 @@
 
 			we.publish(topic.AJAX_CALL_PRECONDITION, attrs);
 
-			if (attrs.mp) { // multipart form. jQuery.ajax() doesn't help here ...
```
andruhon: This one will also change the set of supported browsers: Won't work in IE 9 any more.

andruhon:
Just updated the demo which proves that Drag and Drop file upload now works with this branch.
https://github.com/andruhon/WicketDragAndDropFileAjaxUpload
tested in IE10-11, Chrome and FF

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
solomax: This is noted in bug description here: https://issues.apache.org/jira/browse/WICKET-6517

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxFileUploadBehavior.java
```diff
@@ -0,0 +1,156 @@
...
+
+	public void setMaxSize(Bytes maxSize)
+	{
+		Args.notNull(maxSize, "maxSize");
+		this.maxSize = maxSize;
+	}
+
+	public Bytes getFileMaxSize()
+	{
+		return fileMaxSize;
```
andruhon: I think these `getFileMaxSize` and `getSize` should either be abstract or have a default value, otherwise it's currently can lead to
```
java.lang.IllegalArgumentException: Argument 'maxSize' may not be null.
        at org.apache.wicket.util.lang.Args.notNull(Args.java:41) ~[wicket-util-8.0.0-SNAPSHOT.jar:8.0.0-SNAPSHOT]
        at org.apache.wicket.protocol.http.servlet.MultipartServletWebRequest.setMaxSize(MultipartServletWebRequest.java:121) ~[wicket-core-8.0.0-SNAPSHOT.jar:8.0.0-SNAPSHOT]
```

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxFileUploadBehavior.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.wicket.request.resource.ResourceReference;
+import org.apache.wicket.util.lang.Args;
+import org.apache.wicket.util.lang.Bytes;
+
+/**
+ * Uploads files from a drop event.
+ * 
+ * @author svenmeier
+ */
+public class AjaxFileUploadBehavior extends AjaxEventBehavior
```
andruhon: I think the name of this particular class should somehow point that this one handles drop event, because we might need to ajax upload with some other cause event (such as `change` of file select field). So most of functionality can go into some superclass, probably with `AjaxFileUploadBehavior` name and two subclasses this drag-n-drop one and another one for file input change.

svenmeier: I've changed it to AjaxFileDropBehavior.
I don't think we can and should provide behaviors for all corner cases - people will probably have to build their own implementation as needed. This simple class is just a great way to show the importance of the changes to wicket-ajax-jquery.js

andruhon: I agree, just wanted to emphasise that class name should somehow refer to drag and drop.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/AjaxFileUploadBehavior.java
```diff
@@ -0,0 +1,156 @@
...
+				{
+					fileUploads.add(new FileUpload(item));
+				}
+			}
+			
+			onFileUpload(target, fileUploads);
+		}
+		catch (final FileUploadException fux)
+		{
+			onError(fux);
```
andruhon: I'm pretty sure that target should also be passed into this onError, to make it possible for us to react somehow (show some feedback)

andruhon:
I'm pretty sure that target should also be passed into AjaxFileDropBehavior#onError, to make it possible for us to react somehow (show some feedback)

andruhon:
Another interesting neuance: I didn't find a way to cancel a request before it was actually sent,
Imagine I do drag and drop of multiple files and jquery uploads them one by one. Normally with jQuery it is possible to return false from `beforeSend` handle to cancel the request, however it's not a case with wicket's `bsh`. I could do something like this:
```JavaScript
            Wicket.Ajax.ajax({
                u: attrs.u,
                m: 'POST',
                mp: true,
                bsh: [
                    function (attributes, jqXHR, settings) {
                        settings.xhr = function () {
                            try {
                                if (uploadContext.cancelled) {
                                    // cancel upload
                                    return null;
                                }
                                if (uploadContext.item) {
                                    /* Add progress indicator */
                                    xhr.upload.addEventListener("progress", function (e) {
                                        var progress = Math.round(e.loaded / e.total * 100);
                                        onUploadProgress(uploadContext, attrs, progress);
                                    });
                                }
                                return xhr;
                            } catch (e) {
                            }
                        }
                    }
                ],
```

However it will cause the jQuery.send to explode with `Cannot read property 'open' of null`

I'm not sure what the best way to do that, but probably `_executeHandlers` could return false if any of handlers returns false, something like this:
```JavaScript
		/**
		 * A helper function that executes an array of handlers (before, success, failure)
		 *
		 * @param handlers {Array[Function]} - the handlers to execute
		 * @private
		 */
		_executeHandlers: function (handlers) {
			if (jQuery.isArray(handlers)) {

				// cut the handlers argument
				var args = Array.prototype.slice.call(arguments).slice(1);

				// assumes that the Ajax attributes is always the first argument
				var attrs = args[0];
				var that = this._getTarget(attrs);

				for (var i = 0; i < handlers.length; i++) {
					var handler = handlers[i];
					var r;
					if (jQuery.isFunction(handler)) {
						r = handler.apply(that, args);
					} else {
						r = new Function(handler).apply(that, args);
					}
					if (r === false) {
						return false;
					}
				}
			}
		},
 //...
```
and
```JavaScript
			// execute the request
			var jqXHR = jQuery.ajax({
				url: url,
				type: attrs.m,
				context: self,
				processData: wwwFormUrlEncoded,
				contentType: wwwFormUrlEncoded,
				
				beforeSend: function (jqXHR, settings) {
					if(self._executeHandlers(attrs.bsh, attrs, jqXHR, settings)===false) {
						return false;
					}
					we.publish(topic.AJAX_CALL_BEFORE_SEND, attrs, jqXHR, settings);

					if (attrs.i) {
						// show the indicator
						Wicket.DOM.showIncrementally(attrs.i);
					}
				},
// ...
```

martin-g:
Wicket Ajax has preconditions (`pre`) for canceling.

andruhon:
@martin-g I think it's a bit different: as far as I understand preconditions checked before the `jQuery.ajax` is called, but what I say happens after but before jQuery did actually start this request. I think I should create a separate issue with a little demo for this.

martin-g:
@andruhon I understand what you say. I just explain how it works now.
So preconditions are too soon for your use case ?

andruhon:
I created an issue in JIRA with beforeSend issue mentioned above https://issues.apache.org/jira/browse/WICKET-6532

PR #257, opened 2018-01-13
[WICKET-6518] Log4j is replaced with slf4j-simple

solomax:


klopfdreh:
Much better! Now only the relevant things are available and if the user wants to go for logback he just can add it, but doesn't need to remove not used xmls if logback isn't used. 👍

svenmeier:
+1

PR #258, opened 2018-01-22
Wicket 6498 deferred javascript made simple

svenmeier:
This is a much simpler solution than #253: Applications can just use a JavaScriptDeferHeaderResponse and that's it.

Application has to be changed slightly in its handling of IHeaderResponseDecorators: there's always a decorator that applies a ResourceAggregator - this enables us to chain a header response *behind* the ResourceAggregator.
This simplifies JavaScriptFilteredIntoFooterHeadResponse too, since it no longer has to worry about dependencies.


wicket-core/src/test/java/org/apache/wicket/markup/head/filter/FilteringHeaderResponseTest.java
```diff
@@ -74,4 +75,22 @@ protected Response getRealResponse()
...
-}
+
+	/**
+	 * WICKET-6498 all JavaScript resources have an "defer" attribute, all other JavaScript is
+	 * inside a {@code document.addEventListener('DOMContentLoaded', function() {}; } hook.
+	 */
+	@Test
+	public void deferred() throws Exception
+	{
+		tester.getApplication().setHeaderResponseDecorator(new IHeaderResponseDecorator()
```
solomax: this can be simplified to be
```
tester.getApplication().setHeaderResponseDecorator(response -> new ResourceAggregator(new JavaScriptDeferHeaderResponse(response)));
```

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/AjaxApplication.java
```diff
@@ -44,6 +46,11 @@ protected void init()
 
 		getDebugSettings().setAjaxDebugModeEnabled(true);
 		
+		setHeaderResponseDecorator(response -> {
```
solomax: this can be simplified to be
```
tester.getApplication().setHeaderResponseDecorator(response -> new ResourceAggregator(new JavaScriptDeferHeaderResponse(response)));
```

solomax:
I have `AjaxEventBehavior("click")` added to the component, and got exception in browser console:
```
Cannot bind a listener for event "click" on element "toggle58" because the element is not in the DOM
```

svenmeier:
Seems to work in wicket-examples. What's the difference?

solomax:
Will try to create quickstart
Might take some time .....

solomax:
@svenmeier 
here is incomplete investigation: I can see same errors using latest 8.0.0-SNAPSHOT from master branch :(
I have checked commit-by-commit and this behavior caused by this https://github.com/apache/wicket/commit/c27a57b31a0e519ef15dfe08b425d85f9a24f685 commit
I'll try to continue investigating using diffs tomorrow
Will appreciate any help here

svenmeier:
Hm, that would be an error on my other building site :P

I have no clue what might causing this atm.

solomax:
Hello @svenmeier 
Finally I was able to create quick-start https://github.com/solomax/repeater-with-events, 
The issue is reproducible using latest wicket-core from master branch
The issue is NOT reproducible in case this https://github.com/solomax/repeater-with-events/blob/master/src/main/java/org/apache/solomax/MainPanel.java#L34 line is commented out `setAuto(true);`

Hope you can understand what "magic" was broken :)))

svenmeier:
Anything related to setAuto() must be magic.

svenmeier:
Actually the problem is just your MainPanel that pretends to be auto, but actually is not: its beforeRender() and that of the contained DataView is call twice. I'll fix this.

svenmeier:
This was an issue with WICKET_6503, which is now fixed on master. Thanks Maxim!

So back to deferred JavaScript :)

solomax:
Thanks @svenmeier , `Auto mode` seems to work :)
I have updated https://github.com/solomax/repeater-with-events with case that fails in deferred mode ....
I'll try to check this from my side

solomax:
It seems I should use `git rebase` while merging ....

svenmeier:
Ah, that one is easy: Your header item is defining a function "wbAction". With deferred items this function is now defined inside the 'DOMContentLoaded' callback only.

This is one of the corner cases that can be solved in client only:

```
	private static StringBuilder getNamedFunctionStr(String name, AbstractDefaultAjaxBehavior b, CallbackParameter... extraParameters) {
		StringBuilder sb = new StringBuilder();
		sb.append("window.").append(name).append(" = function(");
		boolean first = true;
		for (CallbackParameter curExtraParameter : extraParameters) {
			if (curExtraParameter.getFunctionParameterName() != null) {
				if (first) {
					first = false;
				} else {
					sb.append(',');
				}
				sb.append(curExtraParameter.getFunctionParameterName());
			}
		}
		sb.append(") {\n");
		sb.append(b.getCallbackFunctionBody(extraParameters));
		sb.append("}\n;");
		return sb;
	}
```

solomax:
Thanks @svenmeier, my bad, I should investigate deeper.

+1 for merging this PR, everything works as expected :)))

svenmeier:
Merged into master

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/JavaScriptDeferHeaderResponse.java
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public JavaScriptDeferHeaderResponse(IHeaderResponse response)
+	{
+		super(response);
+	}
+	
+	@Override
+	public void render(HeaderItem item)
+	{
+		if (item instanceof IWrappedHeaderItem) {
```
martin-g: Usually it is better to use `while` instead of `if` for such checks

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/JavaScriptDeferHeaderResponse.java
```diff
@@ -0,0 +1,141 @@
...
+		super.render(item);
+	}
+
+	/**
+	 * A specialization that uses native "DOMContentLoaded" events without dependency to external JavaScript.
+	 * <p>
+	 * For Ajax requests we utilize the fact, that {@link PartialPageUpdate} renders {@link #getJavaScript()} only,
+	 * executing the JavaScript directly without any event registration.
+	 */
+	private class NativeOnDomContentLoadedHeaderItem extends OnDomReadyHeaderItem
```
martin-g: As far as I can see this class can be `static`

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/JavaScriptDeferHeaderResponse.java
```diff
@@ -0,0 +1,141 @@
...
+		}
+	}
+	
+	/**
+	 * A specialization that uses native "load" events without dependency to external JavaScript 
+	 * <p>
+	 * For Ajax requests we utilize the fact, that {@link PartialPageUpdate} renders {@link #getJavaScript()} only,
+	 * executing the JavaScript directly without any event registration.
+	 */
+	private class NativeOnLoadHeaderItem extends OnLoadHeaderItem
```
martin-g: `static` ?!

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/JavaScriptDeferHeaderResponse.java
kbachl: may I ask why while is better?

martin-g: Because it will unwrap deeply wrapped items (IWrappedHeaderItem inside IWrappedHeaderItem, inside another IWrappedHeaderItem, ...)

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/JavaScriptDeferHeaderResponse.java
svenmeier: Sure.

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/JavaScriptDeferHeaderResponse.java
svenmeier: I've changed it to 'while', this unwraps recursively nested header items.

PR #260, opened 2018-01-26
[WICKET-6523] unique suffix is added if more than 1 timer behaviour i…

solomax:
…s added for the component

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/AbstractAjaxTimerBehavior.java
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/AbstractAjaxTimerBehavior.java
@@ -16,6 +16,8 @@
  */
 package org.apache.wicket.ajax;
 
+import java.util.List;
+
 import org.apache.wicket.Component;
 import org.apache.wicket.Page;
 import org.apache.wicket.core.request.handler.IPartialPageRequestHandler;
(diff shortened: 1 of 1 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/AbstractAjaxTimerBehavior.java
```diff
@@ -92,6 +95,35 @@ public void renderHead(Component component, IHeaderResponse response)
...
+	{
+		StringBuilder jsId = new StringBuilder(getComponent().getMarkupId());
+		List<AbstractAjaxTimerBehavior> list = getComponent().getBehaviors(AbstractAjaxTimerBehavior.class);
+		if (list.size() != 1)
+		{
+			for (int i = 0; i < list.size(); ++i)
+			{
+				if (list.get(i) == this)
+				{
+					jsId.append('_').append(i);
```
martin-g: Maybe it will be better to use UUID or an id from the session (like the page ids).
Using the behavior index might break when the application removes some behavior and adds another.

If behavior index is used then you may want to use `break` here.

solomax: I thought UUIDs are not being used cause it is long String.
This was my first idea. If there would be no better ideas I'll update PR

PR #261, opened 2018-01-26
Wicket 6523 ajax timers

svenmeier:
Another Ajax timers improvement.


solomax:
The code looks good to me
Our main application also seems to work as expected with this version

+1

PR #262, opened 2018-02-04
WICKET-6526 check HTTP method for all form submissions

duesenklipper:
this moves the HTTP method check from onFormSubmitted to
onFormSubmitted(submitter) so that every form submission
performs this check, instead of only non-ajax requests.
also adds tests that verify this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,6 +184,28 @@
 				</plugins>
 			</build>
 		</profile>
+		<profile>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -184,6 +184,28 @@
 				</plugins>
 			</build>
 		</profile>
+		<profile>
```
martin-g: This is not needed.

PR #265, opened 2018-02-05
WICKET-6528 Don't block user when component not on Page using AJAX

dashorst:
Re-instate the behavior that was in versions < 8.0 that a component
added to an AjaxRequestTarget but is not part of the page, would be
ignored (or failed to refresh with an error in the JavaScript console).

So instead of always throwing an error, we only do so in DEVELOPMENT
mode causing an error page, and in DEPLOYMENT mode we only log the
fault, but don't break the application hard for the user: the component
refresh is ignored, just as it would've been in versions prior to 8.

Fixes WICKET-6528


dashorst:
Is merged with commit https://github.com/apache/wicket/commit/3be55cf5a24982cfca400225aabb6c4296859a7d

PR #266, opened 2018-02-05
WICKET-6530: fix possible race-condition for session invalidation

papegaaij:


martin-g:
This should be backported to 7.x/6.x, no ?

papegaaij:
It has already been backported. The patch applied mostly cleanly on 7.x and 6.x.

PR #267, opened 2018-02-12
 Fix to use locale of the session 

try0:
Fix to output validation message according to locale even when format pattern contains "E".

Sorry if my English is wrong.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/validation/validator/DateValidator.java
+++ b/wicket-core/src/main/java/org/apache/wicket/validation/validator/DateValidator.java
@@ -18,9 +18,7 @@ package org.apache.wicket.validation.validator;
 
 import java.text.SimpleDateFormat;
 import java.util.Date;
-import java.util.Locale;
 
-import org.apache.wicket.protocol.http.WebSession;
 import org.apache.wicket.validation.IValidatable;
 import org.apache.wicket.validation.IValidationError;
 import org.apache.wicket.validation.ValidationError;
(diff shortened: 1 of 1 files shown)


wicket-core/src/main/java/org/apache/wicket/validation/validator/DateValidator.java
```diff
@@ -186,7 +188,15 @@ protected IValidationError decorate(IValidationError error, IValidatable<Date> v
 			// format variables if format has been specified
 			if (format != null)
 			{
-				SimpleDateFormat sdf = new SimpleDateFormat(format);
+				WebSession session = WebSession.get();
```
martin-g: First check that there is a Session with `if (Session.exists()) { ... }` because `Session.get()` will create a new one.

martin-g: There is no need of `Web`Session. Just `Session.get()` should be enough to get the locale.

try0:
Thank you very much for the review.

I fixed to check if session exists and changed WebSession to Session.

klopfdreh:
Hi @try0 
The fix is integrated, now:

master - https://github.com/apache/wicket/commit/6c91b495738c597a9c89f624d6f675cd6eb1f609
wicket-7.x - https://github.com/apache/wicket/commit/7ba197c68508f2e15abf5e4bdda046cd4fa14087

Would you be so kind and close this PR?

kind regards and thanks.

try0:
Hi klopfdreh

Thank you for integrating.
Okay, I see. 

I'm happy to have been able to contribute to Wicket.

martin-g:
Thank you, @try0 !

try0:
＠martin-g
Thank you so much.
You’re welcome.

PR #268, opened 2018-02-15
Remove ArrayListStack in favour of ArrayDeque

Jezza:
I found it as I was working on #223, thought I might as well take a break, and fix this small thing.


wicket-core/src/main/java/org/apache/wicket/DequeueContext.java
```diff
@@ -226,9 +227,9 @@ private DequeueTagAction canDequeueTag(ComponentTag open)
 		}
 
 		DequeueTagAction action;
-		for (int i = containers.size() - 1; i >= 0; i--)
+		for (MarkupContainer container : containers)
```
klopfdreh: Only small side note: Is the order of the for loop important in this case?

Before:
* Last container was taken, then one before last and so on...

Now:
* First is taken, then the second and so on...

The reason why I ask is that the loop was build to explicit iterate in the opposite direction.

wicket-core/src/main/java/org/apache/wicket/DequeueContext.java
```diff
@@ -286,9 +287,8 @@ public MarkupContainer popContainer()
 	 */
 	public Component findComponentToDequeue(ComponentTag tag)
 	{
-		for (int j = containers.size() - 1; j >= 0; j--)
+		for (MarkupContainer container : containers)
```
klopfdreh: See review of line 230.

wicket-core/src/main/java/org/apache/wicket/markup/parser/TagStack.java
```diff
@@ -166,12 +167,13 @@ public ComponentTag getNotClosedTag()
 		// If there's still a non-simple tag left, it's an error
 		if (stack.size() > 0)
 		{
-			for (int i = 0; i < stack.size(); i++)
+			Iterator<ComponentTag> it = stack.descendingIterator();
```
klopfdreh: Maybe we can use this iterator functionality also for line 230 / line 290.

wicket-core/src/main/java/org/apache/wicket/DequeueContext.java
Jezza: I preserved all of the iteration behaviour.

The `ArrayDeque` iterates "backwards" already, hence why there's a `descendingIterator()` for a "forward" iteration.

klopfdreh: Ah ok, thanks a lot for the explanation. Great changes 👍

Jezza: No problem! :)

wicket-util/src/main/java/org/apache/wicket/util/collections/ArrayListStack.java
```diff
@@ -1,139 +0,0 @@
-/*
```
martin-g: I'd prefer if this class is just marked with `@Deprecated` in Wicket 8.x and be removed with Wicket 9.x.

Jezza:
Yeah, that's a better idea.
Deprecated `ArrayListStack`, and completely removed a subclass of it.
It wasn't used, and it was package-private, so I threw that out too.

martin-g:
Merged to master!

martin-g:
And to `wicket-8.x`.

For some reason GitHub didn't detect the change and didn't close automatically the PR.
@Jezza Please close this PR. Thank you!

Jezza:
Glad I could help!

PR #269, opened 2018-03-21
[WICKET-6544] mobile browser detection is improved

solomax:


wicket-core/src/test/java/org/apache/wicket/protocol/http/request/WebClientInfoTest.java
```diff
@@ -680,6 +708,35 @@ public void konqueror()
...
+		final String userAgent = "Mozilla/5.0 (iPhone; CPU iPhone OS 11_2_6 like Mac OS X) AppleWebKit/604.5.6 (KHTML, like Gecko) FxiOS/10.6b8836 Mobile/15D100 Safari/604.5.6";
+		WebClientInfo webClientInfo = new WebClientInfo(requestCycleMock, userAgent);
+
+		assertThat(userAgent, webClientInfo.getProperties().getBrowserVersionMajor(),
+			is(equalTo(10)));
+		assertThat(userAgent, webClientInfo.getProperties().getBrowserVersionMinor(),
+			is(equalTo(6)));
+		assertThat(userAgent, webClientInfo.getProperties().isBrowserMozillaFirefox(),
+			is(equalTo(true)));
+		assertThat(userAgent, webClientInfo.getProperties().isBrowserMozilla(),
```
solomax: This part is something I don;t really like in this PR :(

martin-g:
With the addition of these new mobile browsers user agent strings the code becomes more complex.
Maybe we should think in direction how to make use of external libraries to do this checks, e.g. https://github.com/pieroxy/java-user-agent-detection or https://github.com/HaraldWalker/user-agent-utils

solomax:
And/or maybe browser list and supported patterns can be cleaned up to support more recent versions only?

kbachl:
drop this "feature" completely - it just wont work reliable anyway

klopfdreh:
I also looked a bit around and stumbled over Apache DeviceMap which has been moved into the Attic and from the Attic (http://attic.apache.org/projects/devicemap.html) I followed the link to https://github.com/nielsbasjes/yauaa which seems to be maintained.

@kbachl - I understand your suggestion to remove it because of reliability, but I would keep the functionality with the hint that there might be some not supported browsers which is better than no detection.

klopfdreh:
I just tested a bit around and I would suggest the following changes:

* Drop the support for minor version in exchange for a full version String (11.0, 3.5.8, 4.0b4a)
* Change the implementation to yauaa - I also played around a bit and found some not correctly resolved version numbers (the browsers are all detected correctly) https://github.com/nielsbasjes/yauaa/issues/75
* Change the UserAgent enum to not handle excludes, but only browser names

If the issue is fixed I can provide a PR.

svenmeier:
Guys I get it, there are many good solutions to choose from, but why should we make Wicket dependent on anyone of them?

https://github.com/nielsbasjes/yauaa looks good, but do users really need assistance for this?

`UserAgent  userAgent = analyzer.parse(new WebClientInfo(requestCycle).getUserAgent());`

klopfdreh:
See my suggestion over here: https://github.com/apache/wicket/pull/269

klopfdreh:
Close this one in favour of the mentioned one? (https://github.com/apache/wicket/pull/275)

solomax:
OK, let's close this one :)

PR #273, opened 2018-03-30
WICKET-6321 Support Integrity and Crossorigin attributes

tremel:
WICKET-6321 Support Integrity and Crossorigin attributes for JavaScriptUrlReferenceHeaderItem

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
@@ -17,6 +17,7 @@
 package org.apache.wicket.core.util.string;
 
 import org.apache.wicket.request.Response;
+import org.apache.wicket.request.resource.IntegrityAttributed;
 import org.apache.wicket.util.string.Strings;
 
 
@@ -152,6 +153,30 @@ public class JavaScriptUtils
 	public static void writeJavaScriptUrl(final Response response, final CharSequence url,
 		final String id, boolean defer, String charset, boolean async)
 	{
+                writeJavaScriptUrl(response, url, id, defer, charset, async, null);
+	}
+	/**
+	 * Write a reference to a javascript file to the response object
+	 * 
+	 * @param response
+	 *            The HTTP response
+	 * @param url
+	 *            The javascript file URL
+	 * @param id
+	 *            Unique identifier of element
+	 * @param defer
+	 *            specifies that the execution of a script should be deferred (delayed) until after
+	 *            the page has been loaded.
+	 * @param charset
+	 *            a non null value specifies the charset attribute of the script tag
+	 * @param async
+	 *            specifies that the script can be loaded asynchronously by the browser
+         * @param integrity
+         *            integrity attributes
+	 */
+	public static void writeJavaScriptUrl(final Response response, final CharSequence url,
+		final String id, boolean defer, String charset, boolean async, final IntegrityAttributed integrity)
+	{
 		response.write("<script type=\"text/javascript\" ");
 		if (id != null)
 		{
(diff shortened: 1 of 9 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/head/CssHeaderItem.java
```diff
@@ -200,7 +207,13 @@ public static CssReferenceHeaderItem forReference(ResourceReference reference,
 	public static CssReferenceHeaderItem forReference(ResourceReference reference,
 		PageParameters pageParameters, String media, String condition, String rel)
 	{
-		return new CssReferenceHeaderItem(reference, pageParameters, media, condition, rel);
+                final CssReferenceHeaderItem cssReferenceHeaderItem = new CssReferenceHeaderItem(reference, pageParameters, media, condition, rel);
+                if(reference instanceof IntegrityAttributed) {
```
solomax: whitespaces: space after `if`, line break before `{`
and the whole PR

klopfdreh:
Hi,

we are using a CORS enum for media tags already, maybe this is a good moment to use this class in there, too.

@svenmeier I just told that if there are more places I remember your suggestion 😄 

WDYT?

svenmeier:
@klopfdreh indeed, now is the time to reconsider the former decision  :+1:

wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
```diff
@@ -151,6 +152,30 @@ public static void writeJavaScriptUrl(final Response response, final CharSequenc
...
+	 *            the page has been loaded.
+	 * @param charset
+	 *            a non null value specifies the charset attribute of the script tag
+	 * @param async
+	 *            specifies that the script can be loaded asynchronously by the browser
+         * @param integrity
+         *            integrity attributes
+	 */
+	public static void writeJavaScriptUrl(final Response response, final CharSequence url,
+		final String id, boolean defer, String charset, boolean async, final IntegrityAttributed integrity)
```
svenmeier: Perhaps it's better to switch to an attribute map instead? The method had too much parameters already.

wicket-core/src/main/java/org/apache/wicket/markup/head/JavaScriptHeaderItem.java
```diff
@@ -201,7 +202,13 @@ public static JavaScriptReferenceHeaderItem forReference(ResourceReference refer
 	public static JavaScriptReferenceHeaderItem forReference(ResourceReference reference,
 		PageParameters pageParameters, String id, boolean defer, String charset)
 	{
-		return new JavaScriptReferenceHeaderItem(reference, pageParameters, id, defer, charset, null);
+                final JavaScriptReferenceHeaderItem javaScriptReferenceHeaderItem = new JavaScriptReferenceHeaderItem(reference, pageParameters, id, defer, charset, null);
+                if(reference instanceof IntegrityAttributed) {
```
svenmeier: Does a reference really need to be able to decide about integrity attributes? It can't change charset or defer either, why are integrity attributes treated differently?

tremel:
For a usage example and test I made
[CORS.zip](https://github.com/apache/wicket/files/1874596/CORS.zip), a mofified quickstart showing a bootstrap 4 page with bootstrap css and jquery, popper and bootstrap js with CORS.

svenmeier:
@tremel thanks for your usage example.

It seems to me we're mixing at least three topics in this pull request:
- JavaScriptReferenceHeaderItem needs integrity and and crossOrigin ... fine
- JavaScriptUtils#writeJavaScriptUrl() is getting to many parameters, maybe an attribute map can help here?
- references cannot alter the attributes of their headerItem (besides the url of course): defer, async, charset and now integrity and crossOrigin

IMHO we should solve the last issue separately and uniformly, i.e. not IntegrityAttributed but maybe a broader interface for all attributes or just a callback that allows the reference to alter the headerItem.
For now you could use an IHeaderResponseDecorator to set the attributes.

kbachl:
@svenmeier maybe a bit off topic but what about version itself? jQuery 2 vs 3? - AFAIK wicket 8 should work with jQuery 3.3.1 out of the box?

svenmeier:
@kbachl all jQuery 3.x should work fine. If you think Wicket should switch to v3 as default, you an open anothe issue. For me v2 is fine though.

kbachl:
@svenmeier thx for the info - I originally meant we should mention this maybe in the docs and also provide  the library of jquery 3.3.1 with wicket 8 together?

andruhon:
#376 should open a relatively easy way for integrity and crossorigin once merged

svenmeier:
Indeed.

PR #275, opened 2018-04-12
[WICKET-6544] mobile browser detection is improved (with YAUAA)

klopfdreh:
Hi all,

here my 5 cent regarding the browser detection with https://github.com/nielsbasjes/yauaa

you have to load the diff of the test to see the real benefit.

kind regards

Tobias


wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
```diff
@@ -46,14 +46,20 @@
...
+	 * {@link org.apache.wicket.protocol.http.ClientProperties#isNavigatorJavaEnabled()} property,
+	 * which is not set until an actual reply from a browser (e.g. using {@link BrowserInfoPage} is
+	 * set.
 	 */
 	private final String userAgent;
 
 	/** Client properties object. */
 	private final ClientProperties properties;
 
+	private final static UserAgentAnalyzer UAA = UserAgentAnalyzer.newBuilder()
```
svenmeier: UserAgentAnalyzer isn't reentrant. Since it is expensive to create one, instances have to be pooled.

wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
```diff
@@ -193,165 +198,117 @@ protected String getRemoteAddr(RequestCycle requestCycle)
...
-		setEdgeProperties();
-		setSafariProperties();
+		nl.basjes.parse.useragent.UserAgent parsedUserAgent = UAA.parse(getUserAgent());
+		setInternetExplorerProperties(parsedUserAgent);
+		setKonquerorProperties(parsedUserAgent);
+		setMozillaProperties(parsedUserAgent);
+		setOperaProperties(parsedUserAgent);
+		setChromeProperties(parsedUserAgent);
+		setEdgeProperties(parsedUserAgent);
+		setSafariProperties(parsedUserAgent);
```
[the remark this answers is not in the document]
klopfdreh: Hi @solomax - yes indeed - it was due to the refactoring and those methods were present before - I just refactored them.

wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
klopfdreh: The initialization is expensive, but the UserAgentAnalyzer is thread safe so it can be reused. Maybe we are able to use MetaDataKey

klopfdreh:
Following changes have been made:

* The init() method has been changed into gatherExtendedInfo() which is only be executed if org.apache.wicket.settings.RequestCycleSettings.setGatherExtendedBrowserInfo(boolean) is set to true
* MetaDataKey has been used to initialize YAUAA
* All browser specific methods have been moved into detectBrowserProperties
* gatherExtendedInfo / detectBrowserProperties are protected / public so that the user may override them

wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
```diff
@@ -147,6 +149,93 @@ private String getUserAgentStringLc()
...
+		// Opera
+		properties.setBrowserOpera(UserAgent.OPERA.getUaStrings().contains(userAgentName));
+
+		// Internet Explorer
+		properties.setBrowserInternetExplorer(
+			UserAgent.INTERNET_EXPLORER.getUaStrings().contains(userAgentName));
+
+		// FireFox
+		boolean isFireFox = UserAgent.FIREFOX.getUaStrings()
+			.contains(parsedUserAgent.getValue("AgentName"));
```
solomax: `userAgentName`

wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
```diff
@@ -147,6 +149,93 @@ private String getUserAgentStringLc()
...
+		if (isFireFox)
+		{
+			properties.setBrowserMozillaFirefox(true);
+			properties.setBrowserMozilla(true);
+		}
+		else
+		{
+			properties.setBrowserMozilla(
+				UserAgent.MOZILLA.getUaStrings().contains(parsedUserAgent.getValue("AgentName")));
```
solomax: `userAgentName`

wicket-core/src/test/java/org/apache/wicket/protocol/http/request/WebClientInfoTest.java
```diff
@@ -59,9 +58,52 @@ public void before()
 	}
 	
 	/**
-	 * Test IE 6.x user-agent strings
+	 * Test check check all user agents
 	 */
 	@Test
+	public void testBrowsersAndVersionsOfUserAgents() {
+		// Internet Explorers
```
solomax: I would rather have all these as individual tests, to be able to see which one was failed
And run all of them regardless failure state of others

klopfdreh: This might slow down the test execution because of the init of YAUAA - try it out.

klopfdreh: You will see if something fails in the test results btw.

solomax: You can move init to `@BeforeClass` .....

klopfdreh: Yes that is true, but I think if you change the code / the browser detection you are going to investigate why the test is failing and fix the integration as long as every test is back working again.

@BeforeClass is not working, because WicketTestCase and at this stage no Application is attached to the main thread.

klopfdreh: May you have any other good idea to keep the tests individual.

Just send a patch to me if something would fit the requirement and I am going to integrate it.

solomax: Will try to provide patch later tonight
Have to do day time job right now :(

klopfdreh: Hehe me too 😄 - no hurry, because I think this is going to be released in a further version.

wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
svenmeier: UserAgentAnalyzer#parse(UserAgent) is synchronized, please check its code. This is even mentioned on the project home page.

klopfdreh: Have a look at the recent changes. A lot of stuff changed already and I think it is ok like it is now (except the tests)

wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
```diff
@@ -143,215 +150,133 @@ private String getUserAgentStringLc()
...
-	 * sets the safari specific properties
-	 */
-	private void setSafariProperties()
-	{
-		properties.setBrowserSafari(UserAgent.SAFARI.matches(getUserAgent()));
-
-		if (properties.isBrowserSafari())
+		else
 		{
-			String userAgent = getUserAgentStringLc();
```
solomax: `getUserAgentStringLc` is private and not used anymore

wicket-core/src/test/java/org/apache/wicket/protocol/http/request/WebClientInfoTest.java
solomax: Sorry for delay, here is the code works for me:

```
	private static ThreadLocal<UserAgentAnalyzer> localAnalyzer = new ThreadLocal<UserAgentAnalyzer>();

	@Before
	public void before()
	{
		requestCycleMock = mock(RequestCycle.class);

		webRequest = mock(ServletWebRequest.class);
		when(requestCycleMock.getRequest()).thenReturn(webRequest);

		servletRequest = mock(HttpServletRequest.class);
		when(webRequest.getContainerRequest()).thenReturn(servletRequest);

		if (localAnalyzer.get() == null) {
			WebClientInfo webClientInfo = new WebClientInfo(requestCycleMock, "test");
			webClientInfo.gatherExtendedInfo();
			localAnalyzer.set(Application.get().getMetaData(WebClientInfo.UAA_META_DATA_KEY));
		} else {
			Application.get().setMetaData(WebClientInfo.UAA_META_DATA_KEY, localAnalyzer.get());
		}
	}
	
	@AfterClass
	public static void staticAfter()
	{
		localAnalyzer.set(null);
	}
```

Here are the measurment results:
```
Combined tests
real	0m24.123s
user	1m30.748s
sys	0m1.804s


Individual tests (after above changes were applied)
real	0m22.844s
user	1m23.948s
sys	0m1.832s

```

klopfdreh:
What about merging this? Or do we proceed with this PR after the release @bitstorm ?

svenmeier:
My concerns still hold :(

Most importantly: with that single UserAgentAnalyzer instance you've introduced a bottleneck for all concurrent request that need to gatherExtendedInfo().

wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
```diff
@@ -133,24 +140,99 @@ public final String getUserAgent()
 	}
 
 	/**
-	 * returns the user agent string (lower case).
+	 * Initializes the {@link WebClientInfo} user agent detection. This can be overridden to choose
+	 * a different detection as YAUAA (https://github.com/nielsbasjes/yauaa) - if you do so, you
+	 * might exclude the maven dependency from your project in favor of a different framework.
+	 */
```
svenmeier: No, you can't exclude the dependency - with UserAgentAnalyzer in the signature of #detectBrowserProperties(UserAgentAnalyzer) you will get a NoClassDefFoundError if the dependency is not present.

klopfdreh: Ok, then keep the dependency and just override it. We have to adjust the documentation here.

klopfdreh:
To create UserAgentAnalyzer takes rather a long time. Maybe we are able to find a way to resolve the bottleneck, but my concerns to implement the browser detection in such a bad way like it was before also still hold.

Also when I see the benchmarks I would consider to just inform about the bottleneck and keep it that way, or to turn it gatherExtendedInfo() off if 0.011ms are to much.

nielsbasjes:
@klopfdreh 
A few points about the slow startup.
When you explicitly specify which fields you really need the system will completely 'not load' the rules that have no impact. Especially if you do not need the DeviceBrand/DeviceName the impact will be quite large. You can simply pass something like this .withField("AgentName") .withField("DeviceClass") .withField("AgentVersion") and it will have a big impact on both startup times and run times for a not yet cached parse.

I had a look at your project and with this patch the startup time drops to less than 1.5 seconds.

```
diff --git pom.xml pom.xml
index 00cc8f8..7910bd9 100644
--- pom.xml
+++ pom.xml
@@ -146,7 +146,7 @@
                        <dependency>
                            <groupId>nl.basjes.parse.useragent</groupId>
                            <artifactId>yauaa</artifactId>
-                           <version>4.2</version>
+                           <version>4.5</version>
                        </dependency>
                        <dependency>
                                <groupId>com.google.code.findbugs</groupId>
diff --git wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
index 5a02a3d..1e488e9 100644
--- wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
+++ wicket-core/src/main/java/org/apache/wicket/protocol/http/request/WebClientInfo.java
@@ -151,6 +151,9 @@ public class WebClientInfo extends ClientInfo
                {
                        userAgentAnalyzer = UserAgentAnalyzer.newBuilder()
                                .hideMatcherLoadStats()
+                               .dropTests()
+                               .withField("AgentName")
+                               .withField("AgentVersion")
                                .withCache(25000)
                                .build();
                        Application.get().setMetaData(UAA_META_DATA_KEY, userAgentAnalyzer);
@@ -217,7 +220,7 @@ public class WebClientInfo extends ClientInfo
         */
        protected void setBrowserVersion(nl.basjes.parse.useragent.UserAgent parsedUserAgent)
        {
-               String value = parsedUserAgent.get("AgentVersion").getValue();
+               String value = parsedUserAgent.getValue("AgentVersion");
                if (!"Hacker".equals(value))
                {
                        properties.setBrowserVersion(value);
```

solomax:
I believe we can add browser detection as wicketstuff module and/or as confluence example
Should be easy and useful

klopfdreh:
Due to the vote that Wicket is going to drop the support of Browser detection I am closing this branch right now.

@solomax - yes why not a wicketstuff module. 👍 

/closed

solomax:
@klopfdreh can you create wicketstuff module?

klopfdreh:
Hey @solomax - sadly I am totally busy right now, but I try my best to change this.

PR #276, opened 2018-04-19
WICKET-6550 : Unify all metadata capable objects.

Jezza:
Introduce a IMetadataContext that contains the current metadata
 implementation.
This should allow more generic code to be written for all of the objects
 currently capable of storing and handling metadata.
Ideally, this could be expanded with useful default methods.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Application.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Application.java
@@ -138,7 +138,7 @@ import org.slf4j.LoggerFactory;
  * @see org.apache.wicket.protocol.http.WebApplication
  * @author Jonathan Locke
  */
-public abstract class Application implements UnboundListener, IEventSink
+public abstract class Application implements UnboundListener, IEventSink, IMetadataContext<Object, Application>
 {
 	/** Configuration constant for the 2 types */
 	public static final String CONFIGURATION = "configuration";
@@ -405,6 +405,7 @@ public abstract class Application implements UnboundListener, IEventSink
 	 * @return The metadata
 	 * @see MetaDataKey
 	 */
+	@Override
 	public final synchronized <T> T getMetaData(final MetaDataKey<T> key)
 	{
 		return key.get(metaData);
@@ -518,7 +519,8 @@ public abstract class Application implements UnboundListener, IEventSink
 	 * @throws IllegalArgumentException
 	 * @see MetaDataKey
 	 */
-	public final synchronized <T> Application setMetaData(final MetaDataKey<T> key, final Object object)
+	@Override
+	public synchronized final <T> Application setMetaData(final MetaDataKey<T> key, final T object)
 	{
 		metaData = key.set(metaData, object);
 		return this;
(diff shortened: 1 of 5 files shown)


wicket-core/src/main/java/org/apache/wicket/IMetadataContext.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.wicket;
+
+/**
+ * @author Jezza
+ */
+public interface IMetadataContext<B, R> {
```
Jezza: I don't like the names of the generics, nor the R parameter.
It was only added because the current implementation returns the object the method is enclosed in, so right now, it just conveys that type information.  
Ideally, I would want to remove it.

One solution would be just to return the IMetadataContext<B> itself, but you wouldn't be able to do much with the return result, short of just setting more metadata.

wicket-core/src/main/java/org/apache/wicket/Application.java
```diff
@@ -518,9 +519,13 @@ public void logResponseTarget(final IRequestHandler requestTarget)
 	 * @throws IllegalArgumentException
 	 * @see MetaDataKey
 	 */
-	public final synchronized <T> Application setMetaData(final MetaDataKey<T> key, final Object object)
+	@Override
+	public final <T> Application setMetaData(final MetaDataKey<T> key, final T object)
```
Jezza: I had to change the signature of this method.
Ideally, it should have been that way anyway.
Everything still compiles, which means we weren't putting values in keys that weren't expecting the value.

I fear that some users might be doing exactly that though, so this whole IMetadataContext might have to be put off for a bigger version.

wicket-core/src/main/java/org/apache/wicket/IMetadataContext.java
Jezza: I also have doubts about the interface name.
If anyone has a better suggestion, I'd be happy to swap it out.

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -422,6 +422,7 @@ public Locale getLocale()
 	 * @return The metadata
 	 * @see MetaDataKey
 	 */
+	@Override
 	public synchronized final <M extends Serializable> M getMetaData(final MetaDataKey<M> key)
```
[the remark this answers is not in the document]
Jezza: There's no practical difference with implementation, but I had to move it because it was clashing with the signature from the interface.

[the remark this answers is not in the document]
Jezza: Ah, got it.
Yeah, stupid mistake.
I think initially I mistook the error for the sync modifier, but it was the `T object` parameter.
Reverted that weird change.

martin-g:
Cool!

Please just add Javadoc about the `tparam`s of IMetadataContext!
LGTM!

Jezza:
Added some javadoc with some barebones explanation.

Jezza:
I just realised I should probably do `IMetadataContext<B, R extends IMetadataContext<B, R>>`

Jezza:
I went ahead and changed it.

PR #278, opened 2018-05-04
Changed order of elements in the DataTable's HTML

vrozkovec:
It should not matter, but it does for example when generating PDFs with wkhtmltopdf tool - table footer gets displayed at the top, this change fixes that.


svenmeier:
IIRC the current ordering (<thead><tfoot><tbody>) was chosen deliberately to conform to the HTML standard, but it seems this has changed with HTML5:

https://stackoverflow.com/questions/18901950/thead-tfoot-and-tbody-order-in-html5?rq=1

So we might choose to change the default markup, please open an issue and/or a discussion/vote for this. For now you can just use your custom markup in a custom subclass.

PR #279, opened 2018-05-28
Allow meta tags to be contributed during AJAX request

dhs3000:
Meta tags that are added during an AJAX request seem to be ignored. Their header contribution is not processed as far as I could see. The tag is added on server-side but there was no error message in the browser and the tag does not appear.

After adding a `processMeta` section in `Wicket.Head.Contributor.processContribution` the new meta tag appears.


solomax:
Could you please fix spaces to be tabs?

dhs3000:
I replaced spaces with tabs

PR #283, opened 2018-07-09
Wicket-6563 page store implementation

svenmeier:
Basically I propose to
*unify IPageStore with IDataStore
*allow all IPageStore implementations to use Request/Session data (see DiskPageStore, GroupingPageStore and CryptingPageStore) as needed
*trim PageStoreManager/AbstractPageManager to a very simple facade with a chain of stores, each offering different solutions
*make it easier to add new/alternative page store implementations

I appreciate your feedback.


klopfdreh:
A lot of changes there and we should carefully test this, but I like the idea to unify PageStore / DataStore! 👍

svenmeier:
I've rebased, sqashed and forced-pushed for easier review.

wicket-core/src/main/java/org/apache/wicket/pageStore/DelegatingPageStore.java
```diff
@@ -1,100 +1,76 @@
...
+	 * Versioning is supported depending on the delegate.
+	 */
+	@Override
+	public boolean supportsVersioning()
+	{
+		return delegate.supportsVersioning();
+	}
+	
+	@Override
+	public void addPage(IPageContext context, IManageablePage page) {
```
bitstorm: Just a minor issue with curly brackets indentation which should be on a new line.

wicket-core/src/main/java/org/apache/wicket/pageStore/disk/NestedFolders.java
```diff
@@ -0,0 +1,123 @@
...
+	/**
+	 * Get a nested folder for the given name.
+	 * 
+	 * @param name name 
+	 * @param create
+	 * @return
+	 */
+	public File get(String name, final boolean create) {
+		name = name.replace('*', '_');
+		name = name.replace('/', '_');
```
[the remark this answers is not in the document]
svenmeier: I've moved that code from the old DiskDataStore implementation into this utility class. I think you're right, backslash should probably be escaped too.

wicket-core/src/main/java/org/apache/wicket/pageStore/DelegatingPageStore.java
svenmeier: You're right, I'll do a reformat once we're ready to move this implemenation to master.

wicket-core/src/main/java/org/apache/wicket/DefaultPageManagerProvider.java
```diff
@@ -167,6 +168,10 @@ protected IPageStore newPersistentStore()
 		Bytes maxSizePerSession = storeSettings.getMaxSizePerSession();
 		File fileStoreFolder = storeSettings.getFileStoreFolder();
 
-		return new DiskPageStore(application.getName(), fileStoreFolder, maxSizePerSession, getSerializer());
+		if (storeSettings.isEncrypted()) {
```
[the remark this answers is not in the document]
klopfdreh: Well the delegation could be done by a default implementation (for example as long as you return a store to delegate to). But I think you are right - the store configuration is rather static and because of this the stack is more verbose and not worth the effort to make it that configurable.

martin-g:
I think it would be a good idea to create a new branch at https://github.com/wicketstuff/core/ and migrate datastores-parent project to the new APIs. Just to validate the new approach.
These data stores are rather simple but this is all we have.

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+					stream.close();
+					ois.close();
+				}
+			}
+			catch (Exception e)
+			{
+				log.error("Couldn't load DiskPageStore index from file " + index + ".", e);
+			}
+		}
+		Files.remove(index);
```
martin-g: why is it removed ?

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+	}
+
+	/**
+	 * 
+	 */
```
martin-g: This empty javadoc could be removed.

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+	private void saveIndex()
+	{
+		File storeFolder = folders.getBase();
+		if (storeFolder.exists())
```
martin-g: Should we try to create it if it does not exist ?

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+		{
+			File index = new File(storeFolder, INDEX_FILE_NAME);
+			Files.remove(index);
+			try
+			{
+				OutputStream stream = new FileOutputStream(index);
+				ObjectOutputStream oos = new ObjectOutputStream(stream);
```
martin-g: nit: try-with-resources

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+			}
+			catch (Exception e)
+			{
+				log.error("Couldn't write DiskPageStore index to file " + index + ".", e);
+			}
+		}
+	}
+
+	@Override
+	public Set<String> getContextIdentifiers()
```
martin-g: Is this the session ids ? I think it is. Should we use the same language consistently ? It is a bit confusing once to see `context` and other times `session`

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+		{
+			pageStore.folders.remove(sessionIdentifier);
+
+			sessionIdentifier = null;
+		}
+	}
+
+	/**
+	 * Returns the file name for specified session. If the session folder (folder that contains the
+	 * file) does not exist and createSessionFolder is true, the folder will be created.
```
martin-g: What if the folder does not exist and `createSessionFolder` is `false` ? AFAIS nothing will break but do we want to say something about it ?

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+	 * 
+	 * @param sessionIdentifier
+	 * @param createSessionFolder
+	 * @return file name for pagemap
```
martin-g: pagemap term is not used since Wicket 1.4

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+	/**
+	 * Attribute held in session.
+	 */
+	static class SessionAttribute implements Serializable, HttpSessionBindingListener
+	{
+
+		private final String applicationName;
+
+		/**
+		 * The identifier of the session, must not be equal to {@link Session#getId()}, e.g. when
```
martin-g: `must` or `could` ?

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+		 * the container changes the id after authorization.
+		 */
+		public final String identifier;
```
martin-g: again, I think `sessionId` or `initialSessionId` would be clear

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
```diff
@@ -0,0 +1,569 @@
...
+import org.apache.wicket.util.lang.Args;
+import org.apache.wicket.util.lang.Bytes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A storage of pages in files.
+ * <p>
+ * While {@link DiskPageStore} uses a single file per session, this implementation stores each page
+ * in its own file. This improves on a {@link DiskPageStore disadvantage of DiskPageStore} surfacing
```
martin-g: Maybe now is a good time to rename `DiskPageStore` to `SingleFilePageStore`, and this one to `FilePerPagePageStore` ?

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
```diff
@@ -0,0 +1,569 @@
...
+ */
+public class FilePageStore implements IPersistentPageStore
+{
+	private static final String ATTRIBUTE_PAGE_TYPE = "user.wicket_page_type";
+
+	private static final String FILE_SUFFIX = ".data";
+
+	private static final Logger log = LoggerFactory.getLogger(FilePageStore.class);
+
+	private static final String KEY = "wicket:FilePageStore";
```
martin-g: do not forget to rename this key if the class is renamed as I suggested above. Or better use `"wicket:" + FilePageStore.class.getSimpleName()`

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
```diff
@@ -0,0 +1,569 @@
...
+
+		byte[] data = readFile(attribute.identifier, id);
+		if (data == null) {
+			return null;
+		}
+		
+		IManageablePage page;
+		if (serializer == null)
+		{
+			page = new SerializedPage(id, "unknown", data);
```
martin-g: What is `unknown` here ? The session id ? What could be the consequences of using it ?

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
```diff
@@ -0,0 +1,569 @@
...
+		if (page instanceof SerializedPage)
+		{
+			type = ((SerializedPage)page).getPageType();
+			data = ((SerializedPage)page).getData();
+		}
+		else
+		{
+			if (serializer == null)
+			{
+				throw new WicketRuntimeException("DiskPageStore not configured for serialization");
```
martin-g: DiskPageStore

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
```diff
@@ -0,0 +1,569 @@
...
+
+		byte[] data = null;
+
+		try
+		{
+			FileChannel channel = FileChannel.open(file.toPath());
+			try
+			{
+				int size = (int)channel.size();
+				MappedByteBuffer buf = channel.map(MapMode.READ_ONLY, 0, size);
```
martin-g: Why do we read it off-heap in this page store impl ? Do we want to do the same in DiskPageStore ?

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
```diff
@@ -0,0 +1,569 @@
...
+	/**
+	 * Attribute held in session.
+	 */
+	static class SessionAttribute implements Serializable, HttpSessionBindingListener
+	{
+
+		private final String applicationName;
+
+		/**
+		 * The identifier of the session, must not be equal to {@link Session#getId()}, e.g. when
```
martin-g: `must` vs. `could`

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
```diff
@@ -0,0 +1,569 @@
...
+		}
+	}
+
+	public class LastModifiedComparator implements Comparator<File>
+	{
+
+		@Override
+		public int compare(File f1, File f2)
+		{
+			return (int)(f2.lastModified() - f1.lastModified());
```
martin-g: better use `Long.compare(l1, l2)`

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
```diff
@@ -0,0 +1,569 @@
...
+	}
+
+	@Override
+	public Set<String> getContextIdentifiers()
+	{
+		Set<String> identifiers = new HashSet<>();
+
+		for (File folder : folders.getAll())
+		{
+			identifiers.add(folder.getName());
```
martin-g: It seems we store all files in one flat folder ?! This caused problems before and a logic to split by prefix has been introduced.

svenmeier:
> I think it would be a good idea to create a new branch at https://github.com/wicketstuff/core/ and migrate datastores-parent project to the new APIs. Just to validate the new approach.
> These data stores are rather simple but this is all we have.

Good idea, I'm on it +1

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
svenmeier: +1

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
svenmeier: +1

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
svenmeier: It "could" indeed.

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -290,7 +291,7 @@ public final void clear()
 	{
 		if (isTemporary() == false)
 		{
-			getPageManager().clear();
+			getPageManager().removeAllPages();
```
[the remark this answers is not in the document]
svenmeier: +1

wicket-core/src/main/java/org/apache/wicket/pageStore/CryptingPageStore.java
```diff
@@ -0,0 +1,153 @@
...
+		return true;
+	}
+
+	/**
+	 * Supports asynchronous add if the delegate supports it.
+	 */
+	@Override
+	public boolean canBeAsynchronous(IPageContext context)
+	{
+		getSessionData(context);
```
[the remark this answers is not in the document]
svenmeier: I've added a short explanation, IPageStore#canBeAsynchronous() has more details.

wicket-core/src/main/java/org/apache/wicket/pageStore/CryptingPageStore.java
```diff
@@ -0,0 +1,153 @@
...
+	 * Create a new {@link ICrypter} for the given context.
+	 */
+	protected ICrypter newCrypter(IPageContext context) {
+		return new DefaultCrypter();
+	}
+
+	@Override
+	public IManageablePage getPage(IPageContext context, int id)
+	{
+		IManageablePage page = super.getPage(context, id);
```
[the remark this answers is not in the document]
svenmeier: I've changed it to use getDelegate().

wicket-core/src/main/java/org/apache/wicket/pageStore/DefaultPageContext.java
```diff
@@ -23,79 +23,85 @@
...
+public class DefaultPageContext implements IPageContext
 {
-	private static final MetaDataKey<Object> requestCycleMetaDataKey = new MetaDataKey<Object>()
-	{
-		private static final long serialVersionUID = 1L;
-	};
+	private Session session;
 
+	public DefaultPageContext() {
+		this.session = Session.get();
```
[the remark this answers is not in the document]
svenmeier: Sure, we don't need that member.

wicket-core/src/main/java/org/apache/wicket/pageStore/DefaultPageContext.java
```diff
@@ -23,79 +23,85 @@
...
 	@Override
-	public String getSessionId()
+	public <T extends Serializable> T setSessionData(MetaDataKey<T> key, T value)
 	{
-		return Session.get().getId();
+		synchronized (session)
+		{
+			T oldValue = session.getMetaData(key);
+			if (oldValue != null) {
+				return oldValue;
```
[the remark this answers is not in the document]
svenmeier: IPageContext#setSessionData() explains it in detail: this method acts like putIfAbsent().

Many page stores have to keep something in the session, but it should not be set concurrently from different requests.

Maybe we should rename it?

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+public class DiskPageStore implements IPersistentPageStore
+{
+	private static final Logger log = LoggerFactory.getLogger(DiskPageStore.class);
+
+	private static final String KEY = "wicket:DiskPageStore";
+
+	private static final String INDEX_FILE_NAME = "DiskPageStoreIndex";
+
+	/**
+	 * A cache that holds all page stores.
```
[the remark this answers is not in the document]
svenmeier: Indeed.

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+		DISK_STORES.remove(applicationName);
+
+		log.debug("Destroyed.");
+	}
+
+	@Override
+	public boolean canBeAsynchronous(IPageContext context)
+	{
+		// session attribute must be added here *before* any asynchronous calls
+		// when session is no longer available
```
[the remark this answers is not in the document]
svenmeier: AsynchronousPageStore uses a task, it's named PendingAdd now:
It does some magic so that the code being executed on another thread can still access the IPageContext, e.g. the stored session attributes *although* the session is actually no longer accessible from there.

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+		{
+			removeDiskData(diskData);
+		}
+	}
+
+	protected void removeDiskData(DiskData diskData)
+	{
+		synchronized (diskDatas)
+		{
+			diskDatas.remove(diskData.sessionIdentifier);
```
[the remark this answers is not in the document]
svenmeier: Yes, the wording is difficult here: The sessionId isn't always equal to the 'session identifier' (see dicussion above), the former might change (e.g. the container changes it), but the latter must stay the same so that data stored in a separate storage (disk, files, cassandra) can still be identified.

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+			}
+			else
+			{
+				if (serializer == null)
+				{
+					throw new WicketRuntimeException(
+						"DiskPageStore not configured for serialization");
+				}
+				data = serializer.serialize(page);
+				type = page.getClass().getName();
```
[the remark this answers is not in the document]
svenmeier: +1

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
```diff
@@ -0,0 +1,717 @@
...
+	protected DiskData getDiskData(final IPageContext context, final boolean create)
+	{
+		SessionAttribute attribute = getSessionAttribute(context, create);
+
+		if (!create && attribute == null)
+		{
+			return null;
+		}
+
+		return getDiskData(attribute.identifier, create);
```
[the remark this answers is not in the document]
svenmeier: Same as above: The identifier must not be identical to the session-id.

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
svenmeier: Has been that way in DiskDataStore forever.

Note that the index is only valid when the server is shut down. During operation the index is updated in-memory only, to be written to disk not until the next shutdown.

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
svenmeier: At this moment it is created if at least one page was written to disk.

That code comes from the former DiskDataStore, and I've moved all code the folder handling to the NestedFolders utilty,

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
svenmeier: +1

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
svenmeier: Actually that method is called with createSessionFolder=true only (even before this pull-request) so I removed the parameter.

wicket-core/src/main/java/org/apache/wicket/pageStore/DiskPageStore.java
svenmeier: As above, it doesn't have to.

wicket-core/src/main/java/org/apache/wicket/pageStore/FilePageStore.java
svenmeier: Yeah, though about that too. May we can do that later, as it will make any review even more difficult to follow.
I'm thinking about deprecating DiskPageStore anyway, since it has some major drawbacks:
- the index doesn't survive hard server crashes
- hard to support transparently moving session to a fallback server
- pages being purged to early because of alternating requests between browser tabs

wicket-devutils/src/main/java/org/apache/wicket/devutils/debugbar/InspectorDebugPanel.java
```diff
@@ -73,7 +74,22 @@ protected ResourceReference getImageResourceReference()
...
+			@Override
+			public String getObject()
+			{
+				Page enclosingPage = getPage();
+				long pageSize = WicketObjects.sizeof(enclosingPage);
+				Bytes pageSizeInBytes = (pageSize > -1 ? Bytes.bytes(pageSize) : null);
+				String pageSizeAsString = pageSizeInBytes != null ? pageSizeInBytes.toString()
+					: "unknown";
+
+				return "Page: " + pageSizeAsString;
```
martin-g: `Page size: ` ? Otherwise it may render `Page: unknown` and it will be a little confusing

wicket-devutils/src/main/java/org/apache/wicket/devutils/diskstore/browser/PersistedPagesProvider.java
```diff
@@ -0,0 +1,133 @@
...
+	@Override
+	public Iterator<? extends IPersistedPage> iterator(long first, long count)
+	{
+		List<IPersistedPage> pages = getPages();
+
+		if (getSort() != null) {
+			Collections.sort(pages, new SortComparator());
+		}
+
+		return pages.iterator();
```
martin-g: `pages.subList(first, count).iterator()`

wicket-core/src/main/java/org/apache/wicket/pageStore/InMemoryPageStore.java
```diff
@@ -0,0 +1,252 @@
...
+		MemoryData data = datas.get(sessionIdentifier);
+		if (data == null)
+		{
+			return new ArrayList<>();
+		}
+
+		synchronized (data)
+		{
+			return StreamSupport.stream(data.spliterator(), false)
+				.map(page -> new PersistedPage(page.getPageId(), page instanceof SerializedPage ? ((SerializedPage)page).getPageType() : page.getClass().getName(), getSize(page)))
```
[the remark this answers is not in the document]
svenmeier: I agree for consistency.

wicket-core/src/main/java/org/apache/wicket/pageStore/InSessionPageStore.java
```diff
@@ -0,0 +1,266 @@
...
+	 * @param serializer
+	 *            for serialization of pages if session gets persisted
+	 */
+	public InSessionPageStore(IPageStore delegate, int maxPages, ISerializer serializer)
+	{
+		super(delegate);
+
+		this.serializer = Args.notNull(serializer, "serializer");
+
+		this.maxPages = maxPages;
```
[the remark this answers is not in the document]
svenmeier: +1

wicket-core/src/main/java/org/apache/wicket/pageStore/SerializedPage.java
```diff
@@ -0,0 +1,88 @@
...
+	private final int pageId;
+	
+	private final String pageType;
+
+	private final byte[] data;
+
+	public SerializedPage(int pageId, String pageType, byte[] data)
+	{
+		this.pageId = pageId;
+		this.pageType = pageType;
```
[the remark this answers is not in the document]
svenmeier: It can, I've update the javadoc.

wicket-core/src/main/java/org/apache/wicket/pageStore/disk/NestedFolders.java
```diff
@@ -0,0 +1,123 @@
...
+
+	public Set<File> getAll()
+	{
+		Set<File> files = new HashSet<>();
+		
+		if (base.exists()) {
+			for (File low : base.listFiles()) {
+				for (File high: low.listFiles()) {
+					for (File file : high.listFiles()) {
+						files.add(file);
```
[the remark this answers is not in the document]
martin-g: The javadoc of `File#listFiles()` says that it may return `null` in some cases.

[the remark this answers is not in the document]
svenmeier: Ah yes, I'll add a check.

wicket-devutils/src/main/java/org/apache/wicket/devutils/debugbar/InspectorDebugPanel.java
svenmeier: Actually I think this was better before where we just showed a "Inspector" label - serializing the containing page might not something we want to do on all pages. I'll revert this change.

wicket-core/src/main/java/org/apache/wicket/pageStore/InSessionPageStore.java
```diff
@@ -0,0 +1,266 @@
...
+{
+
+	private static final MetaDataKey<SessionData> KEY = new MetaDataKey<SessionData>()
+	{
+		private static final long serialVersionUID = 1L;
+	};
+
+	private final ISerializer serializer;
+
+	private int maxPages;
```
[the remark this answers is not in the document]
svenmeier: No, there's no size based eviction currently. I'll see to move that code to the page store.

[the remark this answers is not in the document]
svenmeier: IEvictionStrategy was tightly coupled to HttpSessionDataStore - I've added size based eviction to InMemoryPageStore and InSessionPageStore.

svenmeier:
> `GroupingPageStore` ... When one should use it ?

One of my projects makes heavy use of multi-tabbed usage of its application, i.e. editing of items is done each in its own tab.
Regretfully long-term usage of one tab evicts pages in other tabs. This can be prevented with GroupingPageStore and AjaxNewWindowNotifyingBehavior, so that each tab gets its own pool of pages.

Since then I discovered that DiskPageStore was part of the problem too (i.e. alternating Ajax requests from different tabs evict pages even quicker, see DiskPageStore's javaDoc ), so GroupingPageStore isn't so much useful anymore.

svenmeier:
I've updated wicketstuff-datastore on branch https://github.com/wicketstuff/core/tree/WICKET-6563-page-stores

solomax:
wicketstuff build fail: https://travis-ci.org/wicketstuff/core/jobs/485357669

martin-g:
It fails because it depends on WICKET-6563

bitstorm:
Piiing! :-)

PR #285, opened 2018-07-23
Improve some logging statements

lzh3636:
There are two changes made in this pull request to improve some logging statements:

1. There is a possible copy and paste error in the log messages (The logging statement was copied from an old place to a new place, but the message wasn't changed to adapt to the function of the new place) which may cause confusion when operators are reading the log messages. 

**org.apache.wicket.pageStore.AsynchronousPageStore.storePage
org.apache.wicket.pageStore.AsynchronousDataStore.storeData**
In the above two classes, there are two identical logging statements:
`log.debug("Storing synchronously page with id '{}' in session '{}'", id, sessionId);`
While in AsynchronousDataStore.storeData the code is dealing with **data** rather than **page**, so I changed the message from page to data.

2. In **org.apache.wicket.core.util.lang.PropertyResolver.findSetter**, different exception types are caught, but the logging messages are the same and there's no stack trace. So people cannot know what exception happened here and cannot distinguish if the exception type is NoSuchMethodException or just general Exception. Simply adding a full stack trace is able to improve it.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/core/util/lang/PropertyResolver.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/util/lang/PropertyResolver.java
@@ -1066,11 +1066,11 @@ public final class PropertyResolver
 						}
 					}
 				}
-				log.debug("Cannot find setter corresponding to " + getMethod);
+				log.debug("Cannot find setter corresponding to " + getMethod, e);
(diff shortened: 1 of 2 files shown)


klopfdreh:
This PR has been merged and can be closed.

Thanks a lot for the changes!

PR #290, opened 2018-09-06
Fix org.junit as resolution:=optional

SubrahmanyamSistha:
wicket-core and the wicket-util bundle import junit packages, those should be resolution:=optional

getting issue using wicket library as bundles

refer issue : 
http://apache-wicket.1842946.n4.nabble.com/wicket-core-and-util-OSGI-wiring-problem-with-junit-framework-with-test-environment-td4680840.html


solomax:
@SubrahmanyamSistha it seems you are trying to merge fix for wicket-9 to wicket-8 branch ...
I believe this PR need to be closed and corrected

PR #292, opened 2018-09-14
WICKET-6587 make CheckBoxSelector extensible

duesenklipper:


svenmeier:
Why not add a lookup to CheckBoxSelector instead:

```
public CheckBoxSelector(String id, CheckBox... boxes)
{
	super(id);

	connectedCheckBoxes = boxes;
}

/**
 * Override if you need want the boxes to be collected dynamically.
 */
protected List<Check> getCheckBoxes() {
    return Arrays.asList(connectedCheckBoxes);
}

@Override
protected CharSequence getFindCheckboxesFunction()
{
	return String.format("Wicket.CheckboxSelector.getCheckboxesFunction(%s)",
		buildMarkupIdJSArrayLiteral(getCheckBoxes()));
}
```
No need to hold references to all checkboxes, with the possibility of leakage when one is removed.

duesenklipper:
That is a possibility, yes, though in your example you do hold on to the connectedCheckBoxes. My main idea was to keep it simple, by providing addCheckBox() users can just add checkboxes in a loop, e.g. in ListView#populateItem.

Any checkboxes that are removed from the component hierarchy are cleared from connectedCheckBoxes by the attached behavior, so there shouldn't be a leak.

But I'm not heavily invested in this design, if you prefer a lookup method, I can do that.

svenmeier:
Ah, I didn't notice your cleanup behavior. That's an improvement to the old solution already +1.

I just think that a dynamic lookup with an overriden #getCheckBoxes() would be more flexible. Why not do both: Add the cleanup all checkboxes passed in the constructor *and* let users override the lookup method if they want/need to.

PR #294, opened 2018-09-26
WICKET-5552 fix modal mousedown and make modal more extendable

andruhon:
Demo to reproduce the issue https://github.com/andruhon/wicket-modal-mousedown-issue

Demo using 9.0.0-SNAPSHOT from maven local to demonstrate the fix:
https://github.com/andruhon/wicket-modal-mousedown-issue/tree/fix-WICKET-5552

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -2285,9 +2285,10 @@
 			 * @param {Element} element - element clicking on which the drag should begin
 			 * @param {Function} onDragBegin - handler called at the begin on dragging - passed element as first parameter
 			 * @param {Function} onDragEnd - handler called at the end of dragging - passed element as first parameter
-			 * @param {Function} onDrag - handler called during dragging - passed element and mouse deltas
+			 * @param {Object} settings - extra settings:
+			 * 				stopDragOnCssSelector - css selector to ignore mouseDownHandler event handler
 			 */
-			init: function(element, onDragBegin, onDragEnd, onDrag) {
+			init: function(element, onDragBegin, onDragEnd, onDrag, settings) {
 
 				if (typeof(onDragBegin) === "undefined") {
 					onDragBegin = jQuery.noop;
@@ -2299,6 +2300,10 @@
 
 				if (typeof(onDrag) === "undefined") {
 					onDrag = jQuery.noop;
+				}
+
+				if (typeof(settings) === "object" && settings.stopDragOnCssSelector) {
+					element.wicketStopDragOnSelector = settings.stopDragOnCssSelector;
 				}
 
 				element.wicketOnDragBegin = onDragBegin;
@@ -2314,6 +2319,14 @@
 				e = Wicket.Event.fix(e);
 
 				var element = this;
+
+				if (element.wicketStopDragOnSelector) {
+					var target = $(e.target);
+					if (target.is(element.wicketStopDragOnSelector) || target.closest(element.wicketStopDragOnSelector).size()) {
+						/* ignore the event instead of swallowing everything else */
+						return true;
+					}
+				}
 
 				Wicket.Event.stop(e);
 
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
@@ -284,6 +284,10 @@
 				width: 600,  /* initial width */
 				height: 300, /* may be null for non-iframe, non-resizable window (automatic height) */
 
+				headerHeight: 40,
+
+				overflow: "Auto",
+
 				resizable: true,
 
 				widthUnit: "px", /* valid only if not resizable */
@@ -308,6 +312,8 @@
 					this.close();
 					return false;
 				}, this), /* called when close button is clicked */
+
+				afterInit: function() { },
 
 				onClose: function() { }, /* called when window is closed */
 
@@ -405,7 +411,7 @@
 		 * Binds the handler to the drag event on given element.
 		 */
 		bind: function(element, handler) {
-			Wicket.Drag.init(element, Wicket.bind(this.onBegin, this), Wicket.bind(this.onEnd, this), Wicket.bind(handler, this));
+			Wicket.Drag.init(element, Wicket.bind(this.onBegin, this), Wicket.bind(this.onEnd, this), Wicket.bind(handler, this), {"stopDragOnCssSelector": ".wicket-ignore-drag"});
 		},
 
 		/**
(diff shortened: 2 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -2301,6 +2302,10 @@
 					onDrag = jQuery.noop;
 				}
 
+				if (typeof(settings) !== "undefined" && settings.stopDragOnCssSelector) {
```
solomax: According to the code `settings` should be `object`
Maybe it would be better to change this line to be
`typeof(settings) === 'object'` ?

andruhon: True. I'll change it to === 'object'. Will check that it still works and push.

andruhon:
Updated with typeof object check.

andruhon:
Is there anything else I should rework here?

svenmeier:
Why so complicated - I've fixed the issue by removing the mouse blocker from modal.js and restricting any dragging on the element itself:

```
diff --git a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
index 64852de..8d94b36 100644
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -2485,6 +2485,10 @@
 
 			mouseDownHandler: function (e) {
 				e = Wicket.Event.fix(e);
+				
+				if (e.target !== this) {
+					return;
+				}
 
 				var element = this;
 
diff --git a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
index f683119..5a0f122 100644
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
@@ -1189,7 +1189,7 @@
 					"<div class=\"w_left\" id='"+idLeft+"'>"+
 						"<div class=\"w_right_1\">"+
 							"<div class=\"w_right\" id='"+idRight+"'>"+
-								"<div class=\"w_content_1\" onmousedown=\"Wicket.Event.stop(event);\">"+
+								"<div class=\"w_content_1\" >"+
 									"<div class=\"w_caption\"  id=\""+idCaption+"\">"+
 										"<a class=\"w_close\" style=\"z-index:1\" href=\"#\"></a>"+
 										"<h3 id=\""+idCaptionText+"\" class=\"w_captionText\"></h3>"+
```
Can you try that instead?

andruhon:
@svenmeier sure, it was the first thing I tried. Unfortunately it prevents the w_captionText of the popup from working as a drag area, which I think is undesired effect.

Potentially can be done as something like `e.target !== this && e.target.className.indexOf('w_captionText') == -1`, but it makes the mouseDownHandler know about css classes of modal so I think the approach with some general stopper class is more universal.

svenmeier:
Ha, that's really some crufty code :)

Just noticed that you filed the issue for Wicket 6.x - note that we will only apply security fixes for that version.

andruhon:
@svenmeier it still exists in 7, 8 and in 9 snapshot. Are there jiras for appropriate versions? Didn't find them.

andruhon:
Shoud I create a new PR without a ticket name or rewording commit is enough?

svenmeier:
I'm still not happy about the change :/.

Note that for Wicket 9 I want to move all Drag-handling from wicket-ajax to modal.js - we don't need it really in the former and for modal we'll have it easier to adjust it.

For now we could allow the start handler to veto the drag:

```
diff --git a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
index 18ab6bd..e71b4e8 100644
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -2476,14 +2476,16 @@
 
 				var element = this;
 
+				if (element.wicketOnDragBegin(element, e) === false) {
+					return;
+				}
+
 				Wicket.Event.stop(e);
 
 				if (e.preventDefault) {
 					e.preventDefault();
 				}
 
-				element.wicketOnDragBegin(element);
-
 				element.lastMouseX = e.clientX;
 				element.lastMouseY = e.clientY;
 
diff --git a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
index b00ceec..6d63807 100644
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
@@ -901,7 +901,12 @@
 		/**
 		 * Called when dragging has started.
 		 */
-		onBegin: function(object) {
+		onBegin: function(element, event) {
+			// ignore anything inside the content
+			if (jQuery(event.target).closest('.w_content_2').size()	) {
+				return false;
+			}
+			
 			if (this.isIframe() && (Wicket.Browser.isGecko() || Wicket.Browser.isIELessThan11() || Wicket.Browser.isSafari())) {
 				this.revertList = [];
 				Wicket.Iframe.documentFix(document, this.revertList);
@@ -1183,7 +1188,7 @@
 					"<div class=\"w_left\" id='"+idLeft+"'>"+
 						"<div class=\"w_right_1\">"+
 							"<div class=\"w_right\" id='"+idRight+"'>"+
-								"<div class=\"w_content_1\" onmousedown=\"Wicket.Event.stop(event);\">"+
+								"<div class=\"w_content_1\">"+
 									"<div class=\"w_caption\"  id=\""+idCaption+"\">"+
 										"<a class=\"w_close\" style=\"z-index:1\" href=\"#\"></a>"+
 										"<h3 id=\""+idCaptionText+"\" class=\"w_captionText\"></h3>"+

```

andruhon:
@svenmeier Do you plan to completely remove the drag handling from wicket-ajax or just to allow the modal.js control it by itself?

svenmeier:
It is used by ModalWindow and the Ajax debug console only - if we can agree to remove the latter (has been discussed a few years already), we can just move the code to modal.js
But that's a separate discussion.

svenmeier:
I've fixed WICKET-5552, please open a new pull request if you think ModalWindow is in need of further extensibility.

andruhon:
Thanks for ace845c412d6ea3117c6f366d9bd91b06c6270da. I'll create another one with extensibility improvements soon.

PR #295, opened 2018-10-12
WICKET-6586 Broken JavaScript due to fix charsetName in JavaScriptPac…

martin-g:
…kageResource

Introduce PackageResource#getProcessingEncoding() that could be overridden by the application to use a custom encoding to read a resource from the class path.
Move the compressing logic from Css/JavaScript-PackageResource to PackageResource class to DRY it.


svenmeier:
+1

PR #296, opened 2018-10-15
BufferedWebResponse's Cleanup

Jezza:
As I was debugging an issue I was having, I saw this class being a bit outdated.
Thought I'd just give it a bit of face lift.
The main goal was to simplify the 10-15 classes in there that was just used for a single method invocation.
Ideally, I've pushed all of that functionality into a single class, and the lambda for a given action.
We still need the class because of the action type, but now there's no need for the 10+ classes.

Open to suggestions for any more improvements.
For example, I don't know if it was a wise choice to use the ActionType as the constructor for the action.
It makes it a bit nicer to read over constructing an Action manually.

I was also tempted to run my formatter over it, but I thought against it.
I actually don't know what template rules wicket follows, if any.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/protocol/http/BufferedWebResponse.java
+++ b/wicket-core/src/main/java/org/apache/wicket/protocol/http/BufferedWebResponse.java
@@ -22,6 +22,7 @@ import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


bitstorm:
@Jezza Code rules can be found in wicket-core which contains file EclipseCodeFormat.xml. Unfortunately this file is not documented anywhere yet.

Jezza:
@bitstorm Ah, must have missed it.
I'll do my best to stick to it.

PR #300, opened 2018-10-24
[WICKET-6603] Asypc page/data store destroyed without hanging

solomax:


solomax:
nice catch :)
I'll update PR, thanks!

solomax:
It would be impossible to call `pageSavingThread.join();` to ensure Thread is stopped
Would it be OK?

svenmeier:
I like to minimize the count of member variables:

```
@Override
public void destroy()
{
	final Thread pageSavingThread = this.pageSavingThread;
	this.pageSavingThread = null;
	if (pageSavingThread.isAlive())
	{
		try
		{
			pageSavingThread.join();

@Override
public void storePage(String sessionId, IManageablePage page)
{
	if (pageSavingThread == null)
	{
		return;
	}
```

Your boolean works fine too.

PR #340, opened 2018-11-22
Improve formatting of example snippet

jmewes:


jmewes:
Currently this snippet is hardly readable in the documentation:

before

With this commit is should become better:

after_png

bitstorm:
thank you @jmewes !

paulbors:
Should we not consider a code style machine cleanup for the entire project?

martin-g:
@paulbors Any improvements are very welcome!

solomax:
I would not change `tab` with 2 spaces for whole project
Even this markup can have `tab` and will be readable if CSS will be in separate file

my 2c

martin-g:
Hi @paulbors ,

I am not sure this is a good idea. There is no big of a problem with the current formatting. Such big re-formattings usually lead to problems later with merging/cherry-picking commits from one branch to the other.

paulbors:
I tend to agree as this proj is widely used and a bug can easily creep in even with the extensive test coverage. If it wasn't as popular it would have made more sense. The kind of good problems to have :)

However, the question still remains: should the comitee vote on adopting a single uniform code style and enforce it going forward? Over time the code base can be refactored to catch up.

PR #342, opened 2018-12-01
Wicket 6616 always render new pages

svenmeier:
Question: Should Wicket always prepare new pages at least once before any component in it is accessed?

For stateless pages
  * until 8.1.0
    * components can be accessed **unprepared** iff they are already present in the page hierarchy
    * otherwise the page is prepared for render
    * any component added in onInitialize() or onBeforeRender() can be accessed **prepared**
  * since 8.2.0 (WICKET-6608)
    * components can be accessed **unprepared** iff they are already present in the page hierarchy
    * otherwise the page is initialized
    * any component added in onInitialize()  (e.g. through queueing) can be accessed **unprepared**
    * otherwise the the page is prepared for render
    * any component added in onBeforeRender() can be accessed **prepared**

IMHO all these cases should allow access to **prepared** components only - otherwise onConfigure() is not yet called, and the component's enabled/visibility-state is not as the developer would expect it.


bitstorm:
@svenmeier  i agree with the overall goal of accessing to render components only. There is however a huge drawback with this solution as now onConfigure is executed before form validation or event handling. Components like FeddbackPanel can not work properly anymore. 
To overcome this limit with stateless pages we can consider to use a brand new page instance for rendering phase. To achieve this result we can for example detach PageComponentProvider after we have invoked the required listener.
I will commit this improvement as soon as possible.

bitstorm:
@svenmeier sorry, what I said wasn't correct. By calling onDetach we don't discard the current page instance but we force configuration phase to be performed again before the actual rendering phase, so no information is lost. 
I still think this can be an improvement because users expect onConfigure to be called before the actual rendering and after event/submission handling.

bitstorm:
Sounds good to me

svenmeier:
This needs a thorough discussion of a solution for 9.x.

PR #345, opened 2018-12-10
Change ClientSideImageMap to use shape 'poly' instead of non-conforming 'polygon'

toby1984:
See https://www.w3.org/TR/html52/semantics-embedded-content.html#the-area-element


svenmeier:
Thanks!

PR #346, opened 2019-01-16
WICKET-6630 Delete (temp) file before writing to it

laurenshop:
As of commons-fileupload 1.4, FileUtils.moveFile() is used to write the uploaded contents to the provided file. This implementation does not allow the destination to be present.


wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUpload.java
```diff
@@ -231,6 +231,10 @@ public long getSize()
 	 */
 	public void writeTo(final File file) throws Exception
 	{
+		if (file.isFile() && file.exists())
```
martin-g: `File#isFile()` implies that it exists. No need of the second condition.

PR #352, opened 2019-04-05
WICKET-6649 Store the applicationName, sessionId and key in WebSocket…

martin-g:
… Text/BinaryMessage and in IWebSocketConnection

Downport of 12b1fb227564c22f4887dea23610be7cb8c5d977 for Wicket 8.x.
It manages to trick maven-clirr-plugin (binary compatibility checker) by using dummy values for Application, session id and IKey

### DIFF
--- a/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/message/BinaryMessage.java
+++ b/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/message/BinaryMessage.java
@@ -16,29 +16,57 @@
  */
 package org.apache.wicket.protocol.ws.api.message;
 
+import org.apache.wicket.Application;
+import org.apache.wicket.Session;
+import org.apache.wicket.mock.MockApplication;
+import org.apache.wicket.protocol.ws.api.registry.IKey;
+
 /**
  * A {@link IWebSocketMessage message} with binary data
  *
  * @since 6.0
  */
-public class BinaryMessage implements IWebSocketMessage
+public class BinaryMessage extends AbstractClientMessage
 {
 	private final byte[] data;
 	private final int offset;
 	private final int length;
 
 	/**
-	 * Constructor.
+	 * Not used by Wicket since 8.5.0!
 	 *
 	 * @param data
 	 *      the binary message from the client
 	 * @param offset
 	 *      the offset to read from
 	 * @param length
-	 *      how much data to read
+	 * 		the length of the data to read
 	 */
+	@Deprecated
 	public BinaryMessage(byte[] data, int offset, int length)
 	{
+		this(new MockApplication(), "", new IKey() {}, data, offset, length);
+	}
+
+	/**
+	 * Constructor.
+	 *
+	 * @param application
+	 *      the Wicket application
+	 * @param sessionId
+	 *      the id of the http session
+	 * @param key
+	 *      the page id or resource name
+	 * @param data
+	 *      the binary message from the client
+	 * @param offset
+	 *      the offset to read from
+	 * @param length
+	 * 		the length of the data to read
+	 */
+	public BinaryMessage(Application application, String sessionId, IKey key, byte[] data, int offset, int length)
+	{
+		super(application, sessionId, key);
 		this.data = data;
 		this.offset = offset;
 		this.length = length;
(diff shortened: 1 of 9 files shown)


wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/message/BinaryMessage.java
```diff
@@ -16,29 +16,57 @@
...
 	 * @param offset
 	 *      the offset to read from
 	 * @param length
-	 *      how much data to read
+	 * 		the length of the data to read
 	 */
+	@Deprecated
 	public BinaryMessage(byte[] data, int offset, int length)
 	{
+		this(new MockApplication(), "", new IKey() {}, data, offset, length);
```
martin-g: Is this acceptable ?

It is needed to be able to downport the improvement made in Wicket 9.x.
This constructor is used only by Wicket. I do not expect the user application to have made use of it, but I cannot be fully certain.

solomax: I see no issues here :)

wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/message/TextMessage.java
```diff
@@ -16,19 +16,46 @@
...
+	/**
+	 * Not used by Wicket since 8.5.0!
+	 *
+	 * @param text
+	 * 		the message sent from the client
+	 */
+	@Deprecated
+	public TextMessage(CharSequence text)
 	{
+		this(new MockApplication(), "", new IKey() {}, text);
```
reiern70: I guess you can't use Application.get() because can't be sure there is thread local Application attached to thread? Would it make sense to do this with some "static"  

```
ty {
 return Application.get()  
} catch() {
  return new MockApplication():
}
```

martin-g: Wicket uses the new constructors at `org.apache.wicket.protocol.ws.api.AbstractWebSocketProcessor#onMessage()` and here there is no Application ThreadLocal for sure. But let's assume that there are Application and Session thread locals - the IKey would be still unknown, and this makes all attempts useless. It would only work if the developer overrides the getter methods of Text/BinaryMessage, but then it would be better to migrate to the new constructors.

reiern70: ok

martin-g:
Any more opinions ?
Should we merge this PR or you think it might break user applications ?

solomax:
Hello @martin-g,
The original request was to simplify message sending to some "subset" of websocket connections
Not sure how this code change will help ...

martin-g:
I understood the request as: `make the Application, Session and IKey available in TextMessage, BinaryMessage and IWebSocketConnection, so the application can use them when deciding to whom to send response message`.
@reiern70 did I understand you correctly ?

reiern70:
@martin-g 

As requester of this feature if there is any risk of breaking existing application I would prefer to use some "workaround". Also the new version of our application is already migrated to Wicket 8. And probably Wicket 9 will be out before we get it into production... So, we will almost certainly be able to use correct implementation on Wicket 9.

Thanks again!

reiern70:
> I understood the request as: `make the Application, Session and IKey available in TextMessage, BinaryMessage and IWebSocketConnection, so the application can use them when deciding to whom to send response message`.
> @reiern70 did I understand you correctly ?

Yes you did.

martin-g:
> As requester of this feature if there is any risk of breaking existing application I would prefer to use some "workaround". Also the new version of our application is already migrated to Wicket 8. And probably Wicket 9 will be out before we get it into production... So, we will almost certainly be able to use correct implementation on Wicket 9.

In that case I will close this PR!
Thank you!

reiern70:
> Hello @martin-g,
> The original request was to simplify message sending to some "subset" of websocket connections
> Not sure how this code change will help ...

Hi,

If IKey of message is same as IKey of connection (page that sent message), then I can get all connections and exclude the one that have same IKey of message. That was my use case.

@martin-g 

Is this reasoning correct? If, so your changes suffice IMHO.

martin-g:
That's correct!

PR #354, opened 2019-04-20
ReplaceAll() to replace() for non-regex patterns

bd2019us:
[WICKET-6657](https://issues.apache.org/jira/browse/WICKET-6657) Removed performance overhead cost associated with replaceAll() when a non-regex pattern is used

### DIFF
--- a/wicket-util/src/main/java/org/apache/wicket/util/lang/Args.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/lang/Args.java
@@ -170,7 +170,7 @@ public class Args
 	 */
 	static String format(String msg, final Object... params)
 	{
-		msg = msg.replaceAll("\\{\\}", "%s");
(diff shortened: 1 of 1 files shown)


wicket-util/src/main/java/org/apache/wicket/util/lang/Args.java
```diff
@@ -170,7 +170,7 @@ public static boolean isFalse(final boolean argument, final String msg, final Ob
 	 */
 	static String format(String msg, final Object... params)
 	{
-		msg = msg.replaceAll("\\{\\}", "%s");
+		msg = msg.replace("\\{\\}", "%s");
```
martin-g: Is the behavior the same ?
I think we need to remove the `\\` now.

martin-g: Adding tests if there are no such will be good!

PR #355, opened 2019-04-29
WICKET-6660 do not trim PasswordTextField input

d3ns0n:
override FormComponent.shouldTrimInput() in PasswordTextField to prevent passwords from being trimmed


svenmeier:
@d3ns0n please close this pull request if you're fine with Wicket continuing to trim passwords - IMHO it's a matter of taste so I'd prefer to keep.

PR #356, opened 2019-05-02
Wicket 6662

bitstorm:
see discussion at https://issues.apache.org/jira/browse/WICKET-6662


martin-g:
I've glanced over several files and the changes look good!
I will make a proper review in the coming days.

But I'd prefer to keep the old classes as deprecated for Wicket 9.x and remove them in Wicket 10.

bitstorm:
@martin-g normally I would do the same thing, but to me these are more classes for internal uses.  I don't think many people use them directly with Wicket. Once we remove them from Wicket code base they are completely useless. But it's not a problem if we decide to keep them in Wicket 9.0.

martin-g:
@bitstorm My reasoning is: If the classes are public then they have to be deprecated before being removed. There is no way to tell how many applications use any of those classes.
There are no any drawbacks in keeping them as deprecated for 1 major release. The size of the jars will be slightly bigger - no one cares about such difference. We won't support those classes since they are deemed to be removed.
We keep them around just for users' convenience.

bitstorm:
@martin-g done! I've reintroduced and deprecated package org.apache.wicket.util.time

wicket-core/src/main/java/org/apache/wicket/request/resource/DynamicImageResource.java
```diff
@@ -19,13 +19,12 @@
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-
+import java.sql.Time;
```
martin-g: this import looks wrong and unused.

wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
```diff
@@ -284,14 +283,14 @@ void ajaxRedirectSetsNoCachingHeaders()
 		tester.startPage(new Wicket3921());
 
 		tester.clickLink("updatePage");
-		assertEquals(Time.START_OF_UNIX_TIME.toRfc1123TimestampString(),
+		assertEquals(Instants.toRFC7231Format(Instant.EPOCH),
```
martin-g: Why change the RFC ?

wicket-core/src/test/java/org/apache/wicket/page/PageAccessSynchronizerTest.java
```diff
@@ -98,13 +106,13 @@ public void run()
 		T2 t2 = new T2();
 		t2.setName("t2");
 		t1.start();
-		Duration.milliseconds(100).sleep();
+		Thread.sleep(100);
```
martin-g: nit: a more direct translation would be: `TimeUnit.MILLISECONDS.sleep(100)`

wicket-jmx/src/main/java/org/apache/wicket/jmx/wrapper/RequestCycleSettings.java
```diff
@@ -117,9 +117,9 @@ public void setResponseRequestEncoding(final String responseRequestEncoding)
 	 * @see org.apache.wicket.jmx.RequestCycleSettingsMBean#setTimeout(java.lang.String)
 	 */
 	@Override
-	public void setTimeout(final String timeout)
+	public void setTimeout(final Duration timeout)
```
martin-g: I am not sure JMX will properly construct a Duration from the String input.

wicket-util/src/main/java/org/apache/wicket/util/string/StringValue.java
```diff
@@ -548,7 +552,14 @@ public final Double toDoubleObject() throws StringValueConversionException
...
 	public final Duration toDuration() throws StringValueConversionException
 	{
-		return Duration.valueOf(text, locale);
+		try
+		{
+			return Duration.parse(text);
+		}
+		catch (Exception e)
+		{
+			throw new StringValueConversionException("xxx", e);
```
martin-g: `xxx`

wicket-jmx/src/main/java/org/apache/wicket/jmx/wrapper/RequestCycleSettings.java
bitstorm: @martin-g  I think you've lost me here :-). The old version of this method took in input a String and used a custom 'human-friendly' format to build a duration. For example something like '1.5 days' would have been converted into 36 hours. I've simply removed this textual format in order to use directly a standard Duration class.

martin-g: JMX works with a text-based protocol, like HTTP.
E.g. a user opens a UI and enters the values for the exposed settings. 
If the type of the property is a String or a primitive then all is fine (e.g. JMX will just use `Integer.parseInt(input)`). 
But if the of the property is of some complex type then someone has to help JMX to convert the String input to the target type.
I doubt there is a default converter for `java.time.Duration`. 

You can test it by starting a Wicket app and then connect to the JVM via JConsole or similar tool (e.g. wicketstuff-jmx-panel) and try to set a new value for `RequestCycleSettings#timeout`

wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
bitstorm: It seems the right RFC for Date formats is RFC7231:
https://tools.ietf.org/html/rfc7231#section-7.1.1.1

But I admit to be a little confused, especially in respect of what RFC1123 actually says about Date formats.

bitstorm:
@martin-g I would proceed merging this PR If there are no other change requests.

martin-g:
Go ahead!

PR #359, opened 2019-05-09
WICKET-6667 Wicket js cleanup

svenmeier:
Before moving to a new technology I'd rather clean up the current JavaScript first.

I've took the effort to separate all changes into separate commits, so anyone interested can assess the impact of that change more easily.

I'm looking at http://youmightnotneedjquery.com/ now.


wicket-core/src/test/js/all.html
```diff
@@ -181,8 +181,6 @@
 
 		<div id="usedAsContextWicket5025"></div>
 
-		<input type="text" id="inputChangeInput">
```
martin-g: why this is no more needed ? I think it is OK to keep this test for the `input` event

wicket-core/src/test/js/dom.js
```diff
@@ -112,38 +112,6 @@ jQuery(document).ready(function() {
 		equal( el.style.display, '', "Wicket.DOM.show should set .style.display to ''." );
 	});
 
-	test("toggleClass() - single CSS class", function() {
```
martin-g: why those tests are removed ?

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/modal/ModalContent1Page.html
```diff
@@ -20,9 +20,5 @@
 			<div wicket:id="modal"></div>
 			<a wicket:id="open">Open another modal dialog</a>
 		</p>
-		<p>
-			<div>An example of a component that uses header contributions</div>
-			<div wicket:id="dateTimeField" />
```
martin-g: how this is related to the JS cleanup ?

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
```diff
@@ -1143,23 +1085,9 @@
 
 				e.style.zIndex = Wicket.Window.Mask.zIndex;
```
andruhon: I believe that dodgy HTML in getMarkup at LL981-1041 can also be heavily simplified. I think it was done that complex to make it drop shadow made from images, which is unnecessary in 2019, because support for browsers without `box-shadow` has dropped a while ago. I can do this refactoring later to make it look nicer as well.

wicket-core/src/test/js/all.html
martin-g: Ignore this. It is better to remove some code.

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/modal/ModalContent1Page.html
martin-g: I saw the commit message. I am OK to do this change in the examples.

wicket-core/src/test/js/dom.js
svenmeier: Because I propose to remove that method.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
svenmeier: modal.js is a beast on its own. IMHO we shouldn't put any effort in it, but deprecate the whole thing instead.

wicket-core/src/test/js/dom.js
martin-g: Let's deprecate it first.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
andruhon: You're not going to deprecate the org.apache.wicket.extensions.ajax.markup.html.modal.ModalWindow itself? I think it's good to refactor it with some new JS or maybe to prepare some replacement with similar interface.

wicket-core/src/test/js/dom.js
svenmeier: We can add @deprecated to those methods in Wicket-8.x, but I'm afraid nobody reads JSDoc anyway. Or log an error instead?

martin-g: Both `@deprecated` and a warning.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
martin-g: Lately there are no many new tickets for ModalWindow but there are several old ones in JIRA which no one cared to work on, me included.
I'm +1 to deprecate it but keep it around. The deprecation is just to tell the users "We don't want to maintain it anymore". But I think we should keep it for the ones which use it without problems. Migrating to another modal (Bootstrap or anything) is an adventure on its own.

svenmeier: ModalWindow is broken beyond repair - not only its JS, but the Wicket internals are broken too, e.g. the need to wrap a Form around it.

svenmeier: +1 we'll keep it around until someone has better ideas.

ivaynberg: We have written a replacement for wicket modal that is ADA compliant and more flexible. I can rip out our dependencies and commit it. It does use jquery...i guess that part will have to be rewritten if we are migrating away.

andruhon: That's a common issue of wicket. I love wicket, it's brilliant, but most of wicket's JS/HTML/CSS looks like 2005, and it has a bad impact on framework decisions people make, because the first thing people check is demos and resulting HTML, unfortunately, looks like a complete rubbish. I really want to do my best to improve this.

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -36,66 +36,28 @@
 		return;
 	}
 
-	/**
```
andruhon: By the way, just for my education. What's the purpose of the `/*global DOMParser: true, ActiveXObject: true, console: true */` in javascript files?

martin-g: Those are **local** hints for JSHint. The global ones are defined at https://github.com/apache/wicket/blob/886f11de02782feaddc3349e1d7ee599922516c3/testing/wicket-js-tests/Gruntfile.js#L71-L98

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
martin-g: > We have written a replacement for wicket modal that is ADA compliant and more flexible. I can rip out our dependencies and commit it. It does use jquery...i guess that part will have to be rewritten if we are migrating away.

That'd be great if the Java API is still the same, or at least similar to the current one.
If it is a totally new implementation then I suggest to discuss it at dev@.

martin-g: > That's a common issue of wicket. I love wicket, it's brilliant, but most of wicket's JS/HTML/CSS looks like 2005

Are you talking about the HTML generated by Wicket(-core/-extensions) components ? Or about wicket-examples ?

Because Wicket generates whatever the application tells it to generate.
Wicket-examples received some styling updates for 8.0.0 but any help to improve it further is very welcome!

wicket-core/src/test/js/dom.js
martin-g: Now I think we can deprecate it in 8.x and remove it for 9.0.0.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
andruhon: I mean wicket-ajax-jquery particularly. It's a 3000 lines file in the source, it has three window.eval, the wicket modal, mentioned above. Demos are also not great.

martin-g: > I mean wicket-ajax-jquery particularly. It's a 3000 lines file in the source, it has three window.eval,

Managers do not care about this, do they ? I think only geeks care about this.

svenmeier:
Just a note that amd.html/.js fails currently. I'm unable to fix it though: as soon as I remove "WicketDebugWindow" from the requires, qunit isn't found. Strange.

martin-g:
> Just a note that amd.html/.js fails currently. I'm unable to fix it though: as soon as I remove "WicketDebugWindow" from the requires, qunit isn't found. Strange.

I've just fixed it!

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/res/modal.js
andruhon: Well, true. But it won't even be included in the list by technical team. At least that's what happened to myself in the past, I considered wicket but then was scared away by demos and wicket-ajax-jquery, so I didn't include it into the list. Also necessity to include 'unsafe-eval' into csp might drop chances of approval to zero if there's a security team present.

bitstorm: > ... I considered wicket but then was scared away by demos

@andruhon I guess your are referring to our [examples site](http://examples8x.wicket.apache.org/index.html). These are mainly technical examples for teaching purpose and yes, they are not the best good-looking examples of Wicket applications. 
There are however a couple of community sites that host more advanced examples using popular web frameworks like [Bootstrap](https://wb7.teliclab.info), [JQuery UI and Kendo UI](http://www.7thweb.net/wicket-jquery-ui). 
Maybe these two should be more visible on our official site (http://wicket.apache.org/).

andruhon: All I wanted to tell is that wicket core's JS needs some attention, wicket-ajax in particulars, and I'm volunteering to discuss and contribute.

svenmeier:
I'll create a Jira issue and merge.

svenmeier:
Thanks to all!

andreikondratev:
I injected this JS into our wicket 7.10 application and send it to our build server to see how Selenium UI tests run.
passed: 1915 tests, failed 8. All 8 are failed due to inputchange removal (I didn't replace the behaviour class)

andruhon:
I injected this JS into our wicket 7.10 application and send it to our build server to see how Selenium UI tests run.
passed: 1915 tests, failed 8. All 8 are failed due to inputchange removal (I didn't replace the behaviour class)

martin-g:
Hi Andrew,

You will have to update your tests to trigger "input" event instead of
"inputchange" event.
Maybe we should provide support for inputchange to be mapped to input in
Java in 9.x, as we were doing for "on" prefix (onclick)?!

andreikondratev:
Mi Martin! 

I think it's redundant to add this support. They failed just because OnChangeAjaxBehavior hasn't been replaced.

andruhon:
I think this behaviour was broken before wicket cleanup http://examples8x.wicket.apache.org/ajax/on-change-ajax-behavior

Very similar thing is broken in wicket 8 demo

andruhon:
Ah, sorry. The demo works, it just a bit misleading without looking at the code. Created a PR to make it more obvious: #364

PR #360, opened 2019-05-10
update Dockerfile to point to tomcat:8.5-jre11

andreikondratev:
because a new build target is 11


martin-g:
Thank you, @andreikondratev !

solomax:
Maybe it worth to use tomcat9?

martin-g:
:+1:

PR #361, opened 2019-05-10
WICKET-6666 initial checkin of new ModalDialog

ivaynberg:
I dont think this is ready to be merged, its just here to start a conversation...

### DIFF
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog-skin.css
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog-skin.css
@@ -1,0 +1,24 @@
+body.modal-dialog-open {
+	
+}
+
+body.modal-dialog-no-scroll {
+}
+
+.modal-dialog-overlay {
+	background: rgba(0, 0, 0, 0.2);
+}
+
+.modal-dialog {
+	max-width: 800px;
+	background: white;
+	box-shadow: 0 0 60px 10px rgba(0, 0, 0, 0.7);
+}
+
+.modal-dialog-scroll-area {
+	overflow: auto;
+}
+
+.modal-dialog-content {
+	padding: 20px;
+}
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.css
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.css
@@ -1,0 +1,43 @@
+body.modal-dialog-open {
+	
+}
+
+body.modal-dialog-no-scroll {
+	overflow: hidden;
+	height: 100%;
+	width: 100%;
+}
+
+body.modal-dialog-open-ios {
+	position: fixed;
+}
+
+.modal-dialog-overlay {
+	position: fixed;
+	top: 0;
+	left: 0;
+	width: 100%;
+	height: 100%;
+	margin: 0;
+	padding: 0;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	flex-direction: column;
+	z-index: 1000000;
+}
+
+.modal-dialog {
+	position: relative;
+	width: 100%;
+	margin: auto;
+	overflow: hidden;
+	z-index: 1000010;
+}
+
+.modal-dialog-scroll-area {
+	overflow: auto;
+}
+
+.modal-dialog-content {
+}
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
@@ -112,6 +112,10 @@ public class ModalDialog extends Panel
 
 		this.options = options;
 
+
+		// the use of prepend here is purposeful. it makes sure that any javascript contributed that
+		// affects elements inside the modal is executed after said elements have been
+		// placed in their correct dom location
 		target.prependJavaScript(getOpenJavascript());
 		return this;
 	}
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
@@ -1,0 +1,240 @@
+package org.apache.wicket.extensions.ajax.markup.html.modal;
+
+import java.io.Serializable;
+
+import org.apache.wicket.Application;
+import org.apache.wicket.ajax.AjaxRequestTarget;
+import org.apache.wicket.markup.head.IHeaderResponse;
+import org.apache.wicket.markup.head.JavaScriptHeaderItem;
+import org.apache.wicket.markup.head.OnDomReadyHeaderItem;
+import org.apache.wicket.markup.html.WebMarkupContainer;
+import org.apache.wicket.markup.html.form.Form;
+import org.apache.wicket.markup.html.panel.EmptyPanel;
+import org.apache.wicket.markup.html.panel.Panel;
+
+import com.github.openjson.JSONStringer;
+
+/**
+ * Presents a modal dialog to the user. See open and close methods.
+ * 
+ * @author Igor Vaynberg (ivaynberg)
+ */
+public class ModalDialog extends Panel
+{
+
+	public static final String CONTENT_ID = "content";
+
+	private final WebMarkupContainer container;
+	private final WebMarkupContainer contentContainer;
+	private boolean open = false;
+	private transient boolean openedInThisRequest = false;
+	private Options options;
+
+	public ModalDialog(String id)
+	{
+		super(id);
+		setOutputMarkupId(true);
+
+		// Container controls the overall visibility of the modal form innards. In its initial state
+		// this is set to
+		// hidden so only the external tag renders. When the modal is openedInThisRequest it is
+		// first repainted with
+		// this shown to
+		// insert the markup into the dom, and then the modal javascript is invoked to rip out this
+		// tag out of the dom
+		// and make it modal.
+		container = new WebMarkupContainer("container");
+		container.setOutputMarkupId(true);
+		container.setVisible(false);
+		add(container);
+
+		// We need this here in case the modal itself is placed inside a form. If that is the case
+		// and the content
+		// contains a form that form's markup form tag will be rendered as div because as far as
+		// wicket is concerned it
+		// is part of the same dom as the page and nested forms are forbidden. By overriding
+		// isRootForm() to true we are
+		// forcing this form - which will be ripped out of the dom along with the content form if
+		// there is one to always
+		// render its tag as 'form' instead of 'div'.
+		var form = new Form<Void>("form")
+		{
+			@Override
+			public boolean isRootForm()
+			{
+				return true;
+			}
+		};
+		form.setOutputMarkupId(true);
+		container.add(form);
+
+		contentContainer = form;
+		contentContainer.add(new EmptyPanel(CONTENT_ID));
+	}
+
+	@Override
+	public void renderHead(IHeaderResponse response)
+	{
+		super.renderHead(response);
+		response.render(JavaScriptHeaderItem.forReference(ModalDialogReferences.JS));
+
+		// if the page is refreshed and the window was open in the previous request we need to
+		// re-open it
+		if (open == true && openedInThisRequest == false)
+		{
+			response.render(OnDomReadyHeaderItem.forScript(getOpenJavascript()));
+		}
+	}
+
+	public ModalDialog open(AjaxRequestTarget target, WebMarkupContainer content)
+	{
+		open(target, null, content);
+		return this;
+	}
+
+	public ModalDialog open(AjaxRequestTarget target, Options options, WebMarkupContainer content)
+	{
+
+		if (!content.getId().equals(CONTENT_ID))
+		{
+			throw new IllegalArgumentException(
+				"Content must have wicket id set to ModalDialog.CONTENT_ID");
+		}
+
+		contentContainer.replace(content);
+
+		container.setVisible(true);
+
+		open = true;
+		openedInThisRequest = true;
+
+		target.add(this);
+
+		this.options = options;
+
+		target.prependJavaScript(getOpenJavascript());
+		return this;
+	}
+
+	protected String getOpenJavascript()
+	{
+		Options options = this.options;
+
+		if (options == null)
+		{
+			options = new Options();
+		}
+
+		if (options.validate == null)
+		{
+			options.validate = Application.get().usesDevelopmentConfig();
+		}
+
+		String optionsJson = options.toJson();
+		String javascript = String.format("window.wicket.modal.open('%s', %s);",
+			container.getMarkupId(), optionsJson);
+		// wrapping in timeout removes the execution out of wicket's ajax update workflow making
+		// errors non-fatal and
+		// errors easier to debug due to simplified stack trace
+		javascript = String.format("window.setTimeout(function() { %s }, 0);", javascript);
+		return javascript;
+
+	}
+
+	public ModalDialog close(AjaxRequestTarget target)
+	{
+		open = false;
+		this.options = null;
+		String javascript = String.format("window.wicket.modal.close('%s');",
+			container.getMarkupId());
+		javascript = String.format("window.setTimeout(function() { %s }, 0);", javascript);
+		target.prependJavaScript(javascript);
+		container.setVisible(false);
+		contentContainer.replace(new EmptyPanel(CONTENT_ID));
+		target.add(this);
+		return this;
+	}
+
+	@Override
+	protected void onDetach()
+	{
+		super.onDetach();
+		openedInThisRequest = false;
+	}
+
+	public static class Options implements Serializable
+	{
+		private Boolean validate;
+		private String console;
+		private String maxWidth;
+		private String maxHeight;
+
+		public Boolean getValidate()
+		{
+			return validate;
+		}
+
+		public void setValidate(Boolean validate)
+		{
+			this.validate = validate;
+		}
+
+		public String getConsole()
+		{
+			return console;
+		}
+
+		public void setConsole(String console)
+		{
+			this.console = console;
+		}
+
+		public String getMaxWidth()
+		{
+			return maxWidth;
+		}
+
+		public void setMaxWidth(String maxWidth)
+		{
+			this.maxWidth = maxWidth;
+		}
+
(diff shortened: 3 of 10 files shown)


wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.js
```diff
@@ -0,0 +1,424 @@
...
+	};
+
+	var behaviors = [ appendBodyClassBehavior, returnFocusOnCloseBehavior, closeOnOverlayClickOrEscapeBehavior, addAriaAttributesBehavior,
+			trapFocusInsideModalBehavior, focusDefaultOnOpeningBehavior, sizingBehavior, secondaryCloseButtonBehavior ];
+
+	//
+	// Entry Methods
+	//
+
+	window.wicket = window.wicket || {};
```
andruhon: Why Wicket ns doesn't start from capital W here?

ivaynberg: no particular reason.

andruhon:
Another thing I really want it to do is to work nicely with the Select2 (maybe it already does). Select2 inside of the old ModalWindow was a pain for quite a while.

ivaynberg:
it already supports select2

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialogReferences.java
```diff
@@ -0,0 +1,38 @@
...
+ * References for {@link ModalDialog}
+ * 
+ * @author Igor Vaynberg (ivaynberg)
+ */
+public class ModalDialogReferences
+{
+	public static final ResourceReference CSS = new CssResourceReference(
+		ModalDialogReferences.class, "ModalDialog.css");
+
+	public static final ResourceReference CSS_SKIN = new CssResourceReference(
```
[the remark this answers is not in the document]
ivaynberg: the skin is the visual part of the css. in the previous modal both visual and layout css were rolled into a single file so it was harder to override what the modal looked like in applications. if you notice the dialog itself only includes the non-skin css, the skin css is included manually on the example page and serves as a starting point for applications to build their own

ivaynberg:
> I believe that some abstract component such as AbstractModalDialog should be created in core to represent dialogs, it could be squeezed in as a base class for existing ModalWindow and maybe for this new one, to make it easier to switch between implementations. Potentially this abstract component can have some very basic JS with enough hooks and callbacks so it can easily be extended with other wicket extensions or replaced completely.
> 
I doubt that is possible. Each dialog works in fundamentally different ways, so a base class makes little sense. The only reason why a separate class exists is to make migration easier, eventually ModalWindow will be removed.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog-skin.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.modal-dialog-overlay {
+	background: rgba(0, 0, 0, 0.2);
+}
+
+.modal-dialog {
+	max-width: 800px;
+	background: white;
+	box-shadow: 0 0 60px 10px rgba(0, 0, 0, 0.7);
```
andruhon: Just my opinion: I'd say 60px is too strong for default style. Maybe 0 0 20px 0 rgba(0,0,0,0.7); ?

andruhon:
Just tested the modal in Chrome, FF, Edge and IE11 all good. Looks a bit dodgy in IE11.

> I doubt that is possible. Each dialog works in fundamentally different ways, so a base class makes little sense. The only reason why a separate class exists is to make migration easier, eventually ModalWindow will be removed.

Maybe only a new one can extend such an abstract class.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.js
martin-g: We should change it to capital `W`. The idea here is to have one global entry in the JS DOM - `window.Wicket`

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.css
```diff
@@ -0,0 +1,43 @@
...
+	left: 0;
+	width: 100%;
+	height: 100%;
+	margin: 0;
+	padding: 0;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	flex-direction: column;
+	z-index: 1000000;
```
martin-g: should `z-index` be configurable ? AFAIK it is configurable in ModalWindow because of Select widgets.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.css
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.modal-dialog {
+	position: relative;
+	width: 100%;
+	margin: auto;
+	overflow: hidden;
+	z-index: 1000010;
```
martin-g: or may be this one should be configurable ? Or both ?!

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
```diff
@@ -0,0 +1,240 @@
...
+		// We need this here in case the modal itself is placed inside a form. If that is the case
+		// and the content
+		// contains a form that form's markup form tag will be rendered as div because as far as
+		// wicket is concerned it
+		// is part of the same dom as the page and nested forms are forbidden. By overriding
+		// isRootForm() to true we are
+		// forcing this form - which will be ripped out of the dom along with the content form if
+		// there is one to always
+		// render its tag as 'form' instead of 'div'.
+		var form = new Form<Void>("form")
```
martin-g: :champagne: I believe this is the first usage of `var` in Wicket ! :-)

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
```diff
@@ -0,0 +1,240 @@
...
+		// this shown to
+		// insert the markup into the dom, and then the modal javascript is invoked to rip out this
+		// tag out of the dom
+		// and make it modal.
+		container = new WebMarkupContainer("container");
+		container.setOutputMarkupId(true);
+		container.setVisible(false);
+		add(container);
+
+		// We need this here in case the modal itself is placed inside a form. If that is the case
```
martin-g: can we do this in `onInitialize()` and add it only if there is / is not an outer form ?

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
```diff
@@ -0,0 +1,240 @@
...
+		container.setVisible(true);
+
+		open = true;
+		openedInThisRequest = true;
+
+		target.add(this);
+
+		this.options = options;
+
+		target.prependJavaScript(getOpenJavascript());
```
martin-g: I would expect to use `append`JavaScript() here. I guess it works with `prepend` because of `setTimeout(..., 0)` below. Should we use `append` to be more clear ?

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.js
```diff
@@ -0,0 +1,424 @@
...
+	var CLOSE_SELECTOR = ".x-modal-close";
+	var SECONDARY_CLOSE_SELECTOR = ".x-modal-close-secondary";
+	
+	//
+	// UTILITY METHODS
+	//
+
+	/** Retreives id of the element, creates one if none */
+	var getOrCreateIdCounter = 0;
+	function getOrCreateId(element) {
```
martin-g: if `element` is a jQuery object then let's name it `$element` to make it more clear.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.js
```diff
@@ -0,0 +1,424 @@
...
+	/**
+	 * Behavior that memorizes the focussed element when dialog is opened, and
+	 * returns focus to it when dialog is closed
+	 */
+	var returnFocusOnCloseBehavior = {
+		open : function(overlay, element, data) {
+			data.opener = document.activeElement;
+			if (data.options.validate) {
+				if (!data.opener || $(data.opener).is("body")) {
+					data.options.console.error("Error saving focused element when opening the modal, it is either none or body: ",
```
martin-g: What is the idea behind `options.console` ? To be able to use either `window.console`, `Wicket.Log` or a third party one ?

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.js
```diff
@@ -0,0 +1,424 @@
...
+				overlay.on("click.modal-dialog", function (event) {
+					if ($(event.target).closest(CONTENT_SELECTOR).length === 0) {
+						// clicked outside modal's content
+						findAndClickCloseButton(element);
+					}
+				});
+			}
+			if (data.options.closeOnEscape) {
+				overlay.on("keydown", function (event) {
+					if (event.which == 27) {
```
martin-g: JSHint (once ModalDialog.js is added to testing/wicket-js-tests/Gruntfile.js) will complain here on `==`. We should use `===` instead.

svenmeier:
Thanks Igor, it's much easier to start with something instead of starting from scratch.

I would like to improve some general points:

- body>overlay>form>content
ModalDialog replicates what ModalWindow has done: its JS creates some wrapping HTML, moves the dialog's content inside of it and appends everything to the body. I think it would be simpler to leave the dialog where it is in the DOM: then we don't have to create markup from JS, no extra form is required and it works without JavaScript too (yes, call me old-school).

- behaviors
ModalDialog has its own pure-JS 'behavior' solution. We can make these Wicket-behaviors, so they are modular and reuasable: some projects will not need all of them, want to replace them with their own or add custom ones (e.g. resizing).

- styling
ModalDialog offers sort of a default skin: I'd like to make this a behavior with a base CSS selector (similar to 'themes' of Palette and AbstractTree) - this makes it possible to use different styles on a single page simultaneously (e.g. a simple warning dialog vs. an larger one with an input form).

- multiple/nested dialogs
Many elements are modified via CSS class selector, we'll have to improve that to support multiple dialogs.

- options.console
IMHO we can simply call Wicket.Log.* (see recent "js-cleanup" discussion)

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.css
ivaynberg: neither should be configurable from java imho, they will be overridden by application-specific css. eg we have this in our css:
```
.modal-dialog-overlay {
	z-index:3 !important;
	& ~ .select2-drop-mask {
		z-index:3;
	}
	& ~ .select2-drop {
		z-index:4; //make sure select2 dropdown is visible inside a modal
	}
}
```

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
ivaynberg: what would you gain? if you are not in a form you would have to add a webmarkupcontainer instead with overridden oncompoennttag that mutates the form tag to a div in markup. it seems like needless complexity to me...

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
ivaynberg: the ```prepend``` is purposeful. if there are any other javascript contributed that works with elements that are inside the modal ```prepend``` guarantees that said javascript will execute after the affected elements are in the right place in the dom. i will add javadoc explaining this...

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.js
ivaynberg: yes, and also to be able to suppress any warnings/errors in production build

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
martin-g: I have to double check but I think the order is:
- prepend JS
- replace DOM
- append JS

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.js
ivaynberg: ok

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
martin-g: Actually the question should be "Why do we need a Form here at all ?"
Why not let the application provide a Form inside the `content` container if it needs one or use the outer (if there is such) ? 
I am on my mobile now, so I cannot re-read ModalWindow's javadoc. But it would be good to simplify here as much as possible. AFAIR Wicket-Bootstrap's Modal component does not provide such mandatory Form and no one complained.

ivaynberg:
> Thanks Igor, it's much easier to start with something instead of starting from scratch.
> 
> I would like to improve some general points:
> 
> * body>overlay>form>content
>   ModalDialog replicates what ModalWindow has done: its JS creates some wrapping HTML, moves the dialog's content inside of it and appends everything to the body. I think it would be simpler to leave the dialog where it is in the DOM: then we don't have to create markup from JS, no extra form is required and it works without JavaScript too (yes, call me old-school).

that is what i started with, but i quickly ran into all sorts of zindex and other problems with more complex layouts. unfortunately there is no substitute for being the last child of body yet.

> * behaviors
>   ModalDialog has its own pure-JS 'behavior' solution. We can make these Wicket-behaviors, so they are modular and reuasable: some projects will not need all of them, want to replace them with their own or add custom ones (e.g. resizing).

im not sure these should be refactored into wicket Behaviors, these are only there to keep the code easy to understand since most things happen in open/close having everything there was messy. and they have lifecycles specific to the modal. 

if customization is required we could move the behaviors array into options and allow users to provide their own

> * styling
>   ModalDialog offers sort of a default skin: I'd like to make this a behavior with a base CSS selector (similar to 'themes' of Palette and AbstractTree) - this makes it possible to use different styles on a single page simultaneously (e.g. a simple warning dialog vs. an larger one with an input form).

i am not sure how necessary this is. unlike the old modal window, this one contains nothing in its body, so you can already create different title bars for warnings/errors/other dialogs by plopping a different panel into the content. 

the only thing that the css provides is the look and feel of the mask and borders which i dont think would change based on your usecase.

but, i do not feel strongly about it so feel free to tweak, the code is in a branch. my only word of caution is that a big reason why the old modal window was such a mess and so hard to maintain is because we put a ton of features into it that only a small percentage of people used, i would try to keep this one as thin as possible.

> * multiple/nested dialogs
>   Many elements are modified via CSS class selector, we'll have to improve that to support multiple dialogs.

most selectors are used downstream of the overlay element, since each overlay will be a direct child of body i dont think there will be collissions... maybe i am wrong

> * options.console
>   IMHO we can simply call Wicket.Log.* (see recent "js-cleanup" discussion)

feel free to make this change, we should also remove it from options then...

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.java
ivaynberg: suppose
* form(1) -> modal -> content -> form(2) -> submit
* because modal is a child of form(1) form(2) will render its tag as a div because html does not support nested forms
* then it is moved into its own direct child of body by modal js
* so now submit button does not have a form tag to work with

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/ModalDialog.js
```diff
@@ -235,7 +235,7 @@
 			}
 			if (data.options.closeOnEscape) {
 				overlay.on("keydown", function (event) {
-					if (event.which == 27) {
```
andruhon: By the way `which` is deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which, `key` is recommended instead https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

svenmeier:
Hi Igor,

> my only word of caution is that a big reason why the old modal window was such a mess and so hard to maintain is because we put a ton of features into it that only a small percentage of people used, i would try to keep this one as thin as possible.

I fully agree, most projects will have their special requirements anyway. This is why the new ModalDialog should rather be a "best practice" implementation instead of a "one fits all" solution.

Sven

svenmeier:
Regarding the location of the dialog in the DOM - in case of a dialog nested inside a form, a page refresh produces invalid markup with two nested form-tags.
Firefox removes the inner form-tag, so after moving the dialog markup to the body, its content is no longer wrapped inside a form-tag :(.

IMHO we should restrain from such tinkering - BTW Bootstrap leaves the dialog where it is in the DOM too.

andruhon:
Another concern is a presence of jQuery. It doesn't do much here now. As far as I remember Wicket has dropped support for IE9 and IE10 in v8. I think the only use for jQuery here is https://developer.mozilla.org/en-US/docs/Web/API/Element/closest if we still support the IE11 (can be polyfilled). closest works fine in MS Edge.

svenmeier:
>most selectors are used downstream of the overlay element

I've double checked, your right :)

svenmeier:
@ivaynberg any objections against continuing the new modal dialog implementation on #375 ?
If not, could you please close this pull request?

ivaynberg:
no objections

PR #364, opened 2019-05-17
Improve OnChangeAjaxBehavior demo

andruhon:
The demo was misleading. I thought it will just display what I typed in the label, I typed 555 and though it's broken.  Updated the demo to be more obvious.


wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/OnChangeAjaxBehaviorPage.java
```diff
@@ -32,6 +33,8 @@
 public class OnChangeAjaxBehaviorPage extends BasePage
 {
 
+	private final IModel<String> inputModel = Model.of("");
```
solomax: I believe this model should be explicitly `detach`ed after such refactoring

andruhon:
Hi Colleagues! What's up with this one?

PR #365, opened 2019-05-30
WICKET-6672: restore use of wicket Duration for Ajax timer

reiern70:


wicket-core/src/main/java/org/apache/wicket/ajax/AbstractAjaxTimerBehavior.java
```diff
@@ -50,11 +50,22 @@
 	 */
 	private String timerId;
 
+	/**
+	 * Construct.
+	 *
+	 * @param updateInterval
+	 *            {@link org.apache.wicket.util.time.Duration} between AJAX callbacks
+	 */
+	public AbstractAjaxTimerBehavior(final org.apache.wicket.util.time.Duration updateInterval)
```
solomax: I would add @Deprecated here

wicket-core/src/main/java/org/apache/wicket/ajax/AjaxSelfUpdatingTimerBehavior.java
```diff
@@ -33,11 +33,22 @@
 {
 	private static final long serialVersionUID = 1L;
 
+	/**
+	 * Construct.
+	 *
+	 * @param updateInterval
+	 *            {@link org.apache.wicket.util.time.Duration} between AJAX callbacks
+	 */
+	public AjaxSelfUpdatingTimerBehavior(final org.apache.wicket.util.time.Duration updateInterval)
```
solomax: I would add @Deprecated here

wicket-util/src/main/java/org/apache/wicket/util/time/Duration.java
```diff
@@ -1,555 +1,565 @@
-/*
```
solomax: Why this class was re-formatted?

wicket-core/src/main/java/org/apache/wicket/ajax/AjaxSelfUpdatingTimerBehavior.java
reiern70: Done!

wicket-util/src/main/java/org/apache/wicket/util/time/Duration.java
reiern70: Not following you. Class seems to have been removed.

reiern70:
@solomax Thanks for review. Added deprecated to constructors. Also rebased + squashed commits.

wicket-util/src/main/java/org/apache/wicket/util/time/Duration.java
solomax: @reiern70 it is reformatted, please load full diff and scroll to the middle :)

reiern70: The only difference I can spot is this...

image

And this is a method I have added.

solomax: The question is: "why the diff is so huge"?
I guess it might be caused by massive line ending change

reiern70: Hum... I remember my IDE complaining about that and I might have chosen the wrong option.

reiern70: image

One of master has CRLF. Which one is correct?

reiern70: Duration class have 

image

and  AjaxSelfUpdatingTimerBehavior

image

solomax: I don't know which one is correct :(
But re-formatting made diff absolutely unusable

reiern70: > I don't know which one is correct :(
> But re-formatting made diff absolutely unusable

My problem was I had

 git config --global core.autocrlf true

PR #368, opened 2019-06-05
Wicket 6673 priority ordering

svenmeier:
With this change all priority items are sorted by the depth of their base component in the component tree, i.e. parent-first but order of siblings is preserved.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/ResourceAggregator.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/ResourceAggregator.java
@@ -27,14 +27,12 @@ import java.util.Set;
 
 import org.apache.wicket.Application;
 import org.apache.wicket.Component;
-import org.apache.wicket.behavior.Behavior;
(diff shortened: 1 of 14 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/head/ResourceAggregator.java
```diff
@@ -202,21 +173,19 @@ public ResourceAggregator(IHeaderResponse real)
 	public void markRendered(Object object)
 	{
 		super.markRendered(object);
-		if (object instanceof Component || object instanceof Behavior)
```
martin-g: Why `Behavior` is no more tracked ?

svenmeier:
Previously both components and behaviors were tracked, to be able to keep PriorityHeaderItems inside each base in their original order.
With my proposal the order of PrioryHeaderItems only changes if one is on a shallower depth in the component tree. So there's no need to track behaviors any longer.
Note that I changed component to render the behaviors' headers in the base of their owning component, see my change there.

PR #370, opened 2019-06-13
[WICKET-6676] Fix for quickstart application not deploying to GlassFish

peedeeboy:
As per https://www.oracle.com/webfolder/technetwork/jsc/xml/ns/javaee/index.html, any `web.xml` files referencing `web_app_3_1.xsd` or `web_app_4_0.xsd` should use the  `http://xmlns.jcp.org/xml/ns/javaee/` namespace, not the older `http://java.sun.com/xml/ns/javaee/` namespace.

Servers that validate the `web.xml` file (i.e. GlassFish) will fail to deploy as XML validation will fail due to the namespace used in the `web.xml` not matching the namespace used in the referenced .xsd


martin-g:
Thank you, @peedeeboy !

PR #371, opened 2019-06-14
[WICKET-6676A] Correction for WICKET-6676

peedeeboy:
Bit embarrassing this... tested the 8.6.0-SNAPSHOT last night and realized there is a typo on Oracle's JavaEE XSD page, which I copied into my commit.

The namespace in `web.xml` should be `http://xmlns.jcp.org/xml/ns/javaee` (without the trailing slash) not `http://xmlns.jcp.org/xml/ns/javaee/`

Apologies for this - my bad for not double checking


martin-g:
Thanks!

PR #372, opened 2019-06-28
Identify javascript template literals to prevent stripping too much code

abogaart:


martin-g:
Thank you, @abogaart !

PR #374, opened 2019-07-10
WICKET-6682 add CSP nonce support: strategy approach

andruhon:
This PR adds a CSP nonce support to Wicket.

Just want to start a discussion. Certainly not ready to be merged yet.

PS. I didn't apply the formatting yet, because it has to be done in Eclipse which I do not use. Will do it later when something is agreed on implementation.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
@@ -19,6 +19,7 @@ package org.apache.wicket.core.util.string;
 import org.apache.wicket.request.Response;
 import org.apache.wicket.util.string.Strings;
 
+import static org.apache.wicket.markup.head.nonce.CspUtils.getNonceAttribute;
 
 /**
  * Provide some helpers to write javascript related tags to the response object.
@@ -152,7 +153,9 @@ public class JavaScriptUtils
 	public static void writeJavaScriptUrl(final Response response, final CharSequence url,
 		final String id, boolean defer, String charset, boolean async)
 	{
-		response.write("<script type=\"text/javascript\" ");
+		response.write("<script ");
+		getNonceAttribute().ifPresent(response::write);
+		response.write("type=\"text/javascript\" ");
 		if (id != null)
 		{
 			response.write("id=\"" + Strings.escapeMarkup(id) + "\" ");
@@ -227,7 +230,9 @@ public class JavaScriptUtils
 	 */
 	public static void writeOpenTag(final Response response, String id)
 	{
-		response.write("<script type=\"text/javascript\" ");
+		response.write("<script ");
+		getNonceAttribute().ifPresent(response::write);
+		response.write("type=\"text/javascript\" ");
 		if (id != null)
 		{
 			response.write("id=\"" + Strings.escapeMarkup(id) + "\"");
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/INonceStrategy.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/INonceStrategy.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.markup.head;
+
+/**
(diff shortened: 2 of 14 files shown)


svenmeier:
A simpler solution is a new DecoratingHeaderResponse that may set a new "nonce" member on JavaScriptContentHeaderItem.
Just like JavaScriptDeferHeaderResponse does it with "defer", no need to introduce a huge new infrastructure.

andruhon:
Thank you @svenmeier !  I'll have a look.

andruhon:
@svenmeier Right, I agree with what you say in #273. It would be nice to have some interface for extra header item attributes.

wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
```diff
@@ -152,7 +153,9 @@ public static void writeJavaScriptUrl(final Response response, final CharSequenc
 	public static void writeJavaScriptUrl(final Response response, final CharSequence url,
 		final String id, boolean defer, String charset, boolean async)
 	{
-		response.write("<script type=\"text/javascript\" ");
+		response.write("<script ");
+		getNonceAttribute().ifPresent(response::write);
+		response.write("type=\"text/javascript\" ");
```
martin-g: There is no need to split line 155 into 156 and 158. The `nonce` attribute could be appended after the `type` attribute. This will save us one call to `Response#write()`

wicket-core/src/main/java/org/apache/wicket/markup/head/INonceStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.markup.head;
+
+/**
+ * Nonce generation strategy. Implementation should define how and when the nonce is generated.
+ * <p>
+ * Strategy Should be configured in the Application.
```
martin-g: `The` `s`trategy should be configured in the application's `{@link SecuritySettings resource settings}`

wicket-core/src/main/java/org/apache/wicket/markup/head/INonceStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+ * The nonce strategy itself is only responsible for nonce generation.
```
martin-g: ... for `CSP nonce generation.

wicket-core/src/main/java/org/apache/wicket/markup/head/nonce/AbstractNonceStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * {@inheritDoc}
+ */
+public abstract class AbstractNonceStrategy implements INonceStrategy {
+
+    private static final int NONCE_LENGTH = 24;
+
+    MetaDataKey<String> NONCE_KEY = new MetaDataKey<String>() {};
+
+    private static final SecureRandom RND = new SecureRandom();
```
martin-g: Is this thread safe ? It would be better to use `ThreadLocalRandom` instead

wicket-core/src/main/java/org/apache/wicket/markup/head/nonce/CspUtils.java
```diff
@@ -0,0 +1,58 @@
...
+package org.apache.wicket.markup.head.nonce;
+
+import org.apache.wicket.Application;
+import org.apache.wicket.markup.head.INonceStrategy;
+
+import java.util.Optional;
+
+/**
+ * Provide some helpers for CSP security.
+ * {@see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script">https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script</a>}
```
martin-g: The text content could be just `CSP Unsafe inline script`. No need to use the full url again.

wicket-core/src/main/java/org/apache/wicket/markup/head/nonce/PerSessionNonceStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public void invalidate() {
+        saveNonce(Session.get(), generateNonce());
+    }
+
+    private void saveNonce(Session session, String nonce) {
+        Args.notNull(session, "session");
+        Args.notNull(nonce, "csp");
+        session.setMetaData(NONCE_KEY, nonce);
```
martin-g: This should also bind the Session. Otherwise the Wicket Session could stay temporary for the request cycle and a new one will be created for the next request.

wicket-core/src/main/java/org/apache/wicket/settings/SecuritySettings.java
```diff
@@ -243,4 +247,26 @@ public SecuritySettings setAuthenticationStrategy(final IAuthenticationStrategy
 		authenticationStrategy = strategy;
 		return this;
 	}
+
+	/**
+	 * Get CSP nonce strategy.
+	 *
+	 * @return returns the CSP nonce strategy. Null if not configured.
+	 */
+	public INonceStrategy getNonceStrategy() {
```
martin-g: get`Csp`NonceStrategy() ?!

andruhon:
Thanks @martin-g ! There's another PR for the same issue https://github.com/apache/wicket/pull/376 . Sven recommended to try different approach above.

andruhon:
I think this one should be closed.

PR #375, opened 2019-07-10
WICKET-6666 modal dialog

svenmeier:
This is my take on a new modal dialog/window for Wicket.

I've taken Igor's work as a base but changed a lot of things, so I thought a separate pull-request would be easier to digest:

- JavaScript

All JavaScript is optional now, ModalDialog works without Ajax too.
We can safely leave the dialog's markup where it is, no need to move it around. If people want to have the markup somewhere else, they can add the dialog somewhere else (see Java below).
JS enhancements (like trap-focus) are located in Wicket behaviors, allowing people to substitute their own solution.

- CSS

I've reorganized the CSS into a theme like we have for several other components. It is changed now to respect the size of the dialog's content.
This leaves all customizing to the user, no need for Wicket to get involved (see examples below).

- Java

No need to fiddle with forms (see JavaScript above). Full support for nesting. In my experience many projects like to have a single container on each page to host lazily created dialogs, I've added AjaxListPanel for that.
I like the #open(Componente, ART) method, to open a single dialog with different contents. But users like to subclass dialogs to use them from different places, thus we need a #setContent(Component) method too.

- Examples

I've extended the ModalDialogPage example to show-off all new features.


andruhon:
I just wanted to have a look at the modal, and I can't build:
`wicket> mvn clean install -DskipTests=true`
```
...
[INFO] Reactor Summary:
[INFO]
[INFO] Wicket Parent 9.0.0-M3-SNAPSHOT .................... SUCCESS [  3.094 s]
[INFO] Wicket Util 9.0.0-M3-SNAPSHOT ...................... FAILURE [  0.825 s]
[INFO] Wicket Request 9.0.0-M3-SNAPSHOT ................... SKIPPED
...
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  5.814 s
[INFO] Finished at: 2019-07-11T10:20:06+12:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project wicket-util: Fatal error compiling: invalid flag: --release -> [Help 1]
...
```

`wicket-util> mvn clean compile -DskipTests=true`
```
...
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.071 s
[INFO] Finished at: 2019-07-11T10:22:19+12:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project wicket-util: Fatal error compiling: invalid flag: --release -> [Help 1]
...
```

andruhon:
Don't mind the above. The mvn also needed JAVA_HOME to be same as one in the toolchain. I'm pretty sure it was working with other version of maven.

andruhon:
If I open dialog with a `Link` in a new tab, it just blinks and disappears immediately (tested in FF, Chrome and Edge).

svenmeier:
That's just AjaxNewWindowNotifyingBehavior creating a new page instance when it detects that you open a page in two different browser tabs.

andruhon:
It seems like this popup does what's needed and is simple enough.

> That's just AjaxNewWindowNotifyingBehavior creating a new page instance when it detects that you open a page in two different browser tabs.

This is a really annoying behavior. It's, probably, outside of the demo scope, but it would be nice to have this link working with a new tab. Maybe somewhat like this
```
Index: wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/modal/ModalDialogPage.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/modal/ModalDialogPage.java	(revision 5b01f231b7cf56b389596efd004bf1e800fe25c3)
+++ wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/modal/ModalDialogPage.java	(date 1574374168753)
@@ -46,8 +46,8 @@
  */
 public class ModalDialogPage extends BasePage
 {
-
-	private AjaxListPanel stackedDialogs;
+	private final ModalFragment modalFragment;
+	private final AjaxListPanel stackedDialogs;
 
 	/**
 	 * Should dialogs be stacked rather than nested
@@ -69,12 +69,18 @@
 		queue(new Radio<Boolean>("yes", Model.of(true)));
 		queue(new Radio<Boolean>("no", Model.of(false)));
 
-		queue(new ModalFragment("start"));
+		modalFragment = new ModalFragment("start");
+		queue(modalFragment);
 
 		stackedDialogs = new AjaxListPanel("stackedDialogs");
 		queue(stackedDialogs);
 	}
 
+	private ModalDialogPage widthDialog() {
+		modalFragment.openDialog(null);
+		return this;
+	}
+
 	@Override
 	public void renderHead(IHeaderResponse response)
 	{
@@ -116,7 +122,7 @@
 				@Override
 				public void onClick()
 				{
-					openDialog(null);
+					setResponsePage(new ModalDialogPage().widthDialog());
 				}
 			});
 

```

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/theme/DefaultTheme.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.wicket.markup.head.IHeaderResponse;
+import org.apache.wicket.request.resource.CssResourceReference;
+import org.apache.wicket.request.resource.ResourceReference;
+
+/**
+ * Default theme for {@link ModalDialog}.
+ * 
+ * @author svenmeier
+ */
+public class DefaultTheme extends Behavior
```
andruhon: It's supposed to be used in other panels and pages, so I would give it more detailed name such as ModalDialogDefaultTheme.
Same with CSS file.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/trap-focus.js
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+;
+(function($, window, document, undefined) {
```
andruhon: I think some header description is needed explaining when to use this focus trap.

svenmeier:
>If I open dialog with a Link in a new tab, it just blinks and disappears immediately (tested in FF, Chrome and Edge).
> This is a really annoying behavior. It's, probably, outside of the demo scope 

Yes, but this impacts all Ajax links opened in a new browser window. We don't have add a workaround for that here.

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/modal/theme/DefaultTheme.java
svenmeier: Actually I don't like it when a class name replicates its package name. I've avoided it with other 'themes' too, e.g. for palette and tree.

bitstorm:
Hi! I've tested the new ModlDialog and it looks cool! I've found just a minor UI issue under Firefox opening different nested dialogs: the resizable right border of the underline dialogs is still active:

Screenshot from 2019-12-07 20-56-45

The problem is not present using 'stacked dialogs' instead of 'nested'

svenmeier:
Thanks Andrea for given the implementation a test!
I'm aware of the issue with the resize handle. It seems to be a bug in Firefox with nested resizable containers. Since this is just the example CSS, I don't care too much about it.
I just wanted to demonstrate, that we don't need huge JS anymore for such features.

bitstorm:
@svenmeier Great! Give me a call when you are done, so we can proceed merging this PR.

chrisco484:
> This is my take on a new modal dialog/window for Wicket.
> 
> I've taken Igor's work as a base but changed a lot of things, so I thought a separate pull-request would be easier to digest:
> 
> * JavaScript
> 
> All JavaScript is optional now, ModalDialog works without Ajax too.
> We can safely leave the dialog's markup where it is, no need to move it around. If people want to have the markup somewhere else, they can add the dialog somewhere else (see Java below).
> JS enhancements (like trap-focus) are located in Wicket behaviors, allowing people to substitute their own solution.
> 
> * CSS
> 
> I've reorganized the CSS into a theme like we have for several other components. It is changed now to respect the size of the dialog's content.
> This leaves all customizing to the user, no need for Wicket to get involved (see examples below).
> 
> * Java
> 
> No need to fiddle with forms (see JavaScript above). Full support for nesting. In my experience many projects like to have a single container on each page to host lazily created dialogs, I've added AjaxListPanel for that.
> I like the #open(Componente, ART) method, to open a single dialog with different contents. But users like to subclass dialogs to use them from different places, thus we need a #setContent(Component) method too.
> 
> * Examples
> 
> I've extended the ModalDialogPage example to show-off all new features.

Have you had a chance to check out the wicketstuff ModalX project?

There may be some useful ideas in there that could improve the whole Modal experience for Wicket devs.

The way Wicketstuff Modal X facilitates modal dialogs makes it very easy to open any modal dialog from anywhere in the app just by calling the constructor on the class of the modal you want to open and then calling show().

It supports nesting up to N levels. The 'N' can be specified by the dev.

Any page that needs modal simply needs to implement the org.wicketstuff.modalx.ModalMgr interface (typically you do this once in a commonly used base page class so that the whole app becomes Modal capable very easily).

The ModalMgr base class auto adds 'N' placeholder components to any page that it renders and these are then automatically allocated in sequence each time a new modal is nested upon another one. As the modals are closed the placeholders are freed up in reverse sequence until, when the last modal is closed, they are all freed up and the user can then proceed to open another top level modal etc.,

svenmeier:
Hi @chrisco484, thanks for chiming in.

Yes, I have used ModalX before. I like that you can show arbitrary dialogs without the need to keep them in the component tree prematurely.
I'm not a big fan of the ModalMgr interface and other details though, but these might stem from the fact ModalX uses the old clumsy ModalWindow in the first place. And I don't think imposing any superclass or interface on pages is a bad idea, since Wicket pages have to much responsibility anyway.
I've seen many projects doing the same with their own implementation that keeps central 'pool' of dialogs somewhere in the component hierarchy. How this pool is managed has to be decided by each project individually.

Note how the example ModalDialogPage uses the new AjaxListPanel to do something similar but more generally:

https://github.com/apache/wicket/blob/a1a6206c30c1f388ec9c3ec32139addb8979c768/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/modal/ModalDialogPage.java#L188

IMHO this serves the same purpose. Please share your thoughts on what might still be missing or could be improved.

svenmeier:
Thanks, I've merge the implementation into master.

PR #376, opened 2019-07-12
WICKET-6682 add CSP nonce support: DecoratingHeaderResponse approach

andruhon:
Another approach to handle CSP nonce as alternative to #374


andruhon:
@svenmeier is it close to what you suggested about the DecoratingHeaderResponse?

wicket-core/src/main/java/org/apache/wicket/core/util/string/CssUtils.java
```diff
@@ -60,17 +64,45 @@ public static void writeCss(final Response response, final CharSequence text, St
...
+	 * @param attributes
+	 */
+	public static void writeOpenTag(final Response response, IValueMap attributes)
 	{
 		response.write(INLINE_OPEN_TAG_START);
-		if (id != null)
-		{
-			response.write(" id=\"" + id + "\"");
+		if (attributes != null) {
+			response.write(" "+attributes.toString());
```
solomax: `toString()` call is redundant here, missing white-spaces

wicket-core/src/main/java/org/apache/wicket/core/util/string/CssUtils.java
```diff
@@ -119,27 +151,33 @@ public static void writeLinkUrl(final Response response, final CharSequence url,
...
-		if (Strings.isEmpty(media) == false)
+		HeaderItemAttributeMap attributes = new HeaderItemAttributeMap();
+
+		if (Strings.isEmpty(rel) == false)
 		{
-			response.write(" media=\"");
-			response.write(Strings.escapeMarkup(media));
-			response.write("\"");
+			attributes.add(HeaderItemAttribute.LINK_REL, rel);
+		} else {
```
solomax: looks like the place for ternary operator ...

wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
```diff
@@ -227,10 +269,23 @@ public static void writeJavaScript(final Response response, final CharSequence t
 	 */
 	public static void writeOpenTag(final Response response, String id)
 	{
-		response.write("<script type=\"text/javascript\" ");
-		if (id != null)
-		{
-			response.write("id=\"" + Strings.escapeMarkup(id) + "\"");
+		HeaderItemAttributeMap attributes = new HeaderItemAttributeMap();
+		attributes.add(HeaderItemAttribute.TYPE, "text/javascript");
+		if (id != null) {
```
solomax: code style

wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
```diff
@@ -227,10 +269,23 @@ public static void writeJavaScript(final Response response, final CharSequence t
...
+	}
+
+	/**
+	 * @param response
+	 * @param attributes
+	 */
+	public static void writeOpenTag(final Response response, IValueMap attributes)
+	{
+		response.write("<script ");
+		if (attributes != null) {
```
solomax: code style

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/CspNonceHeaderResponse.java
```diff
@@ -0,0 +1,348 @@
...
+import org.apache.wicket.util.value.HeaderItemAttributeMap;
+
+/**
+ * Add CSP nonce to all relevant JavaScript and CSS header items
+ * <p>
+ * Note: please don't forget to wrap with {@link ResourceAggregator}
+ *  when setting it up with {@link Application#setHeaderResponseDecorator},
+ *  otherwise dependencies will not be rendered
+ */
+public abstract class CspNonceHeaderResponse extends DecoratingHeaderResponse {
```
solomax: open brace should be on separate line

svenmeier:
Yeah that's much simpler, isn't it? See my inline comments.

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/CspNonceHeaderResponse.java
```diff
@@ -0,0 +1,348 @@
...
+        }
+
+        attributes.add(HeaderItemAttribute.CSP_NONCE, nonce);
+
+        CssUtils.writeLinkUrl(response, attributes);
+
+        response.write("\n");
+    }
+
+    private static class JavaScriptContentWithNonceHeaderItem extends JavaScriptContentHeaderItem {
```
svenmeier: Just pull-up #nonce to JavaScriptContentHeaderItem, it's an official attribute so no reason not to have it there.

wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
```diff
@@ -152,28 +155,48 @@ public static void writeJavaScriptUrl(final Response response, final CharSequenc
 	public static void writeJavaScriptUrl(final Response response, final CharSequence url,
```
svenmeier: These method have to much arguments already: Just give it an IValueMap and write all out to the tag.

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/CspNonceHeaderResponse.java
```diff
@@ -0,0 +1,348 @@
...
+            this.nonce = nonce;
+            return this;
+        }
+    }
+
+    private static class JavaScriptReferenceWithNonceHeaderItem extends JavaScriptReferenceHeaderItem {
+
+        private String nonce;
+
+        public JavaScriptReferenceWithNonceHeaderItem(ResourceReference reference, PageParameters pageParameters, String id, boolean defer, String charset) {
```
svenmeier: Can a JavaScriptReferenceHeaderItem really have a nonce? From what I understood, nonce is need for content-scripts only.

wicket-util/src/main/java/org/apache/wicket/util/value/HeaderItemAttributeMap.java
```diff
@@ -0,0 +1,24 @@
+package org.apache.wicket.util.value;
+
+import java.util.Map;
+
+public class HeaderItemAttributeMap extends AttributeMap {
```
svenmeier: I don't see much value in this class.

wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
andruhon: I didn't quite get you.  Do you want the `writeJavaScriptUrl(final Response response, final CharSequence url, final String id, boolean defer, String charset, boolean async)` to be dropped?

wicket-util/src/main/java/org/apache/wicket/util/value/HeaderItemAttributeMap.java
andruhon: I mainly need it to have put and add below accepting the enum. Is there any other mechanism to avoid doing `attributes.put(SCRIPT_SRC.getName(), val)` every time? I mean the .getName()

wicket-core/src/main/java/org/apache/wicket/markup/head/filter/CspNonceHeaderResponse.java
andruhon: Yes it can. If the the CSP policy is strict as in my example the nonce is needed for the reference like `<script src="jquery.js"></script>` and even for links such as `<link  href="my.css" />`, for some reason nonce for link is not documented on developer.mozilla.org, but that's how it practically works with `script-src 'unsafe-eval' 'nonce-%s'; style-src 'nonce-%s';`

wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
svenmeier: Well, if we continue what we have done in the past, we have to add yet another parameter "nonce".  IMHO it's better to change it to receive a valueMap instead, and let the caller fill it.

andruhon: There's a new method just after this one. I suppose we can mark this one as Deprecated and get rid of all usages in the wicket code.

svenmeier: Ah yes, that's what I was looking for :). This one can be deprecated and should no longer be called.

andruhon:
@svenmeier @martin-g @solomax I pushed some after review improvements and fixed tests.

svenmeier:
@andruhon actually I'm impressed how well this worked out ;)

This definitely fits nicely into the existing framework and is a great feature to mention in the 9.x announcement.

I still have some minor quibbles to sort out before merging (regarding usage of ValueMap/AttributeMap/HeaderItemAttribute/IAttributeMapKey and documentation) but overall this is the way to go.

andruhon:
Thanks @svenmeier ! I'm waiting for comments then. 
This PR is pointing to wicket-8, I wanted to make it a non-breaking change. Potentially this can be first announced as a wicket-9 feature, and later be "retrofited" to the wicket-8.

andruhon:
Speaking about versions. I think that in wicket-8 this feature should produce same or almost the same HTML as before, but in wicket-9 we can drop charset of js, type of css and other deprecated things. There's a JIRA for the following cleanup https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6686

andruhon:
By the way. https://cwiki.apache.org/confluence/display/WICKET/Wicket+Source+Code+Style Is this one still somewhat up to date? It is really a pain for me to run Eclipse every time I want to format things. I configured styles in my IDEA according to this document.

solomax:
I'll try to set up `apache-rat-plugin` and `maven-checkstyle-plugin`
to perform automatic checks and will prepare PR
This should unify things

wicket-core/src/main/java/org/apache/wicket/core/util/string/CssUtils.java
```diff
@@ -119,27 +156,33 @@ public static void writeLinkUrl(final Response response, final CharSequence url,
...
+
+	/**
+	 * Writes a reference to a css file in the response object
+	 *
+	 * @param response
+	 * 		the response to write to
+	 * @param attributes
+	 * 		Attributes map
+	 */
+	public static void writeLinkUrl(final Response response, IValueMap attributes)
```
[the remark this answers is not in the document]
martin-g: writeLink() ?

wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
```diff
@@ -148,32 +151,53 @@ public static void writeJavaScriptUrl(final Response response, final CharSequenc
...
+
+	/**
+	 * Write a reference to a javascript file to the response object
+	 *
+	 * @param response
+	 *            The HTTP response
+	 * @param attributes
+	 *            Extra tag attributes
+	 */
+	public static void writeJavaScriptUrl(final Response response, IValueMap attributes)
```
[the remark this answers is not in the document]
martin-g: `writeScript()` ?

wicket-core/src/main/java/org/apache/wicket/markup/head/CssHeaderItem.java
```diff
@@ -365,6 +382,27 @@ protected final void internalRenderCSSReference(Response response, String url, S
...
+
+	/**
+	 * Set the CSP nonce
+	 *
+	 * @param nonce
+	 * @return {@code this} object, for method chaining
+	 */
+	public CssHeaderItem setNonce(String nonce)
+	{
+		Args.notNull(nonce, "nonce");
```
martin-g: Args.nonEmpty() ?!

wicket-core/src/main/java/org/apache/wicket/markup/head/JavaScriptHeaderItem.java
```diff
@@ -365,15 +372,54 @@ protected final void internalRenderJavaScriptReference(Response response, String
...
+		return nonce;
+	}
+
+	/**
+	 * Set the CSP nonce
+	 * @param nonce
+	 * @return {@code this} object, for method chaining
+	 */
+	public JavaScriptHeaderItem setNonce(String nonce) {
+		Args.notNull(nonce, "nonce");
```
martin-g: nonEmpty

wicket-core/src/main/java/org/apache/wicket/markup/head/OnDomReadyHeaderItem.java
```diff
@@ -127,4 +138,23 @@ public boolean equals(Object o)
...
+		return nonce;
+	}
+
+	/**
+	 * Set the CSP nonce
+	 * @param nonce
+	 * @return {@code this} object, for method chaining
+	 */
+	public OnDomReadyHeaderItem setNonce(String nonce) {
+		Args.notNull(nonce, "nonce");
```
martin-g: nonEmpty

wicket-core/src/main/java/org/apache/wicket/markup/head/OnLoadHeaderItem.java
```diff
@@ -103,6 +114,24 @@ public String toString()
...
+		return nonce;
+	}
+
+	/**
+	 * Set the CSP nonce
+	 * @param nonce
+	 * @return {@code this} object, for method chaining
+	 */
+	public OnLoadHeaderItem setNonce(String nonce) {
+		Args.notNull(nonce, "nonce");
```
martin-g: nonEmpty

wicket-util/src/main/java/org/apache/wicket/util/value/AttributeMap.java
```diff
@@ -46,4 +48,76 @@ public AttributeMap(final Map<String, Object> map)
...
+	}
+
+	/**
+	 * Convenience method computing value for {@link IAttributeMapKey}.
+	 * A convenience method for {@link #compute(Object, BiFunction)},
+	 * Which accepts simple supplier as value.
+	 * <p>
+	 * Doesn't add a mapping if computed value is <code>null</code>
+	 */
+	public Object compute(IAttributeMapKey key, Supplier supplier)
```
martin-g: wouldn't be better to use a BiFunction here instead of Supplier ? Just as java.util.Map does.

martin-g:
> By the way. [cwiki.apache.org/confluence/display/WICKET/Wicket+Source+Code+Style](https://cwiki.apache.org/confluence/display/WICKET/Wicket+Source+Code+Style) Is this one still somewhat up to date? It is really a pain for me to run Eclipse every time I want to format things. I configured styles in my IDEA according to this document.

I've setup IDEA to format only the lines I have edited.

wicket-util/src/main/java/org/apache/wicket/util/value/AttributeMap.java
andruhon: @martin-g The BiFunction would bring it to this look: `attrs.compute(key, (s, o) -> Application.getNonce())`, and the current look with a Supplier is `attrs.compute(key, Application::getNonce)`. I can switch to BiFunction if you insist.

andruhon:
Thank you everyone! I'll go through comments a bit later.

wicket-util/src/main/java/org/apache/wicket/util/value/AttributeMap.java
martin-g: For `nonce` indeed there is no need to use the old value, but for some other attribute maybe there would be such need.
Maybe we can add an overloaded method for BiFunction ?

andruhon:
I went a bit further with cleaning names in utility classes and emphasized those which are inline ones, it was hard to remember which one does what even after reading the code.

wicket-core/src/main/java/org/apache/wicket/markup/head/CssHeaderItem.java
```diff
@@ -365,6 +382,27 @@ protected final void internalRenderCSSReference(Response response, String url, S
...
+
+	/**
+	 * Set the CSP nonce
+	 *
+	 * @param nonce
+	 * @return {@code this} object, for method chaining
+	 */
+	public CssHeaderItem setNonce(String nonce)
+	{
+		Args.notEmpty(nonce, "nonce");
```
svenmeier: Why shouldn't it be null? We allow that for other attributes too.

wicket-util/src/main/java/org/apache/wicket/util/value/HeaderItemAttribute.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.apache.wicket.util.value;
+
+import org.apache.wicket.util.lang.Args;
+
+/**
+ * Some common header item attributes.
+ * This is not a complete list of all possible attributes.
+ */
+public enum HeaderItemAttribute implements IAttributeMapKey
```
svenmeier: This enum doesn't add any value, it's just a pit to throw constants into.
I'd prefer to have those attribute names used directly in the header items, as it was before.

wicket-util/src/main/java/org/apache/wicket/util/value/IAttributeMapKey.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.util.value;
+
+/**
+ * {@link AttributeMap} key interface
+ */
+public interface IAttributeMapKey
```
svenmeier: This is overkill for a bunch of attribute names.

wicket-util/src/main/java/org/apache/wicket/util/value/HeaderItemAttribute.java
andruhon: Well, yes, moving all these values into appropriate classes would make the code a lot simpler.

Is it something you're looking for:
in JavaScriptUtils
```
	public static final String ATTR_ID = "src";
	public static final String ATTR_TYPE = "type";
	public static final String ATTR_SCRIPT_SRC = "src";
	public static final String ATTR_SCRIPT_DEFER = "defer";
	public static final String ATTR_SCRIPT_ASYNC = "async";
	public static final String ATTR_SCRIPT_NOMODULE = "nomodule";
	public static final String ATTR_SCRIPT_REFERRERPOLICY = "referrerpolicy";
	public static final String ATTR_NONCE = "nonce";
	public static final String ATTR_SRI_INTEGRITY = "integrity";
	public static final String ATTR_SRI_CROSSORIGIN = "crossorigin";
```
And similar set in CSSUtils
```
	public static final String ATTR_ID = "src";
	public static final String ATTR_LINK_HREF = "href";
	public static final String ATTR_LINK_MEDIA = "media";
	public static final String ATTR_LINK_REL = "rel";
	public static final String ATTR_SRI_INTEGRITY = "integrity";
	public static final String ATTR_SRI_CROSSORIGIN = "crossorigin";
```
?

wicket-core/src/main/java/org/apache/wicket/markup/head/CssHeaderItem.java
andruhon: It makes sense to remove this assert. @solomax @martin-g do you agree?

solomax: I would say this argument can be `null`
But it is really weird, to call this method with empty/null argument
I would say it can raise WARN to show possible misuse ...

andruhon: My idea was, that it's a special security attribute, and you don't really want to erase it. So, that's why I wanted it to explode.

andruhon:
I've got into a trouble with this stuff. Need help. The issue is that we really want all attributes to be escaped, except the URL. 

I came up with this solutions
```
	/**
	 * Generates a <code>String</code> representation of this object.
	 *
	 * @param noEscapeMarkupKeys
	 * 		a set of keys to not aply
	 * @return <code>String</code> representation of this <code>ValueMap</code> consistent with the
	 * tag-attribute style of markup elements. For example: <code>a="x" b="y" c="z"</code>.
	 */
	@Override
	public String toString() {
		return toString(null);
	}

	/**
	 * Generates a <code>String</code> representation of this object.
	 *
	 * @param noEscapeMarkupKeys
	 * 		a set of keys to not aply
	 * @return <code>String</code> representation of this <code>ValueMap</code> consistent with the
	 * tag-attribute style of markup elements. For example: <code>a="x" b="y" c="z"</code>.
	 */
	public String toString(Collection<String> noEscapeMarkupKeys)
	{
		final StringBuilder buffer = new StringBuilder();
		boolean first = true;
		for (Map.Entry<String, Object> entry : entrySet())
		{
			if (first == false)
			{
				buffer.append(' ');
			}
			first = false;
			String key = entry.getKey();
			buffer.append(key);
			buffer.append("=\"");
			final Object value = entry.getValue();
			if (value == null)
			{
				buffer.append("null");
			}
			else if (value.getClass().isArray())
			{
				buffer.append(Arrays.asList((Object[])value));
			}
			else
			{
				buffer.append(
						noEscapeMarkupKeys != null && noEscapeMarkupKeys.contains(key)
								? value
								: Strings.escapeMarkup(String.valueOf(value))
				);
			}

			buffer.append('\"');
		}
		return buffer.toString();
	}
```
The problem is that the interface IValueMap is mostly used everywhere, and I can't really add a new method to the interface within wicket 8.

Any advice?

svenmeier:
I'm confused now - why shouldn't all attributes be escaped? I've read WICKET-4777 but I don't understand it.

andruhon:
@svenmeier I'm also confused. The url for the CSS is escaped, the url for JS is not. I have a solution, will push soon, once tests finished.

andruhon:
@svenmeier the issue, I think that escape markup simply swallows parameters spearated by &, because they are replaced with &amp;

andruhon:
I suppose, the answer is generally that all URLs should be escaped and there, probably, must be a special header item which does accept script urls with parameters

andruhon:
I removed enums, and updated tests. It now produces the same output, including links for JS

svenmeier:
We should reconsider the change introduced WICKET-4777 ... @martin-g can you explain why that JavScript url attribute should not be escaped?

martin-g:
WICKET-4777 is almost 7 years old. My memory capacity is not that big.

I guess my thinking was in the line of: If **Google** doesn't like Wicket's urls then Wicket should adapt, not the opposite.
We can try to revert it and see whether the browsers send non-escaped parameters.

wicket-core/src/main/java/org/apache/wicket/markup/head/CssHeaderItem.java
martin-g: This is how I understand it:

If CSP is not enabled then this method should not be called at all. 
If CSP is enabled then the parameter should be non-empty.

martin-g:
Is `AttributeMap#compute(Object, Supplier)` still used ? 
I see you replaced all (?) callers to use the BiFunction variant. 
If there is no usage of the one with Supplier then let's remove it from the API

andruhon:
@martin-g the `AttributeMap#compute` has already been removed.

andruhon:
This PR is for wicket-8.x, I propose to merge it to the wicket-8.x with behavior replicating existing one. And to create another PR for wicket-9.x removing no-escape stuff for JS.

wicket-core/src/main/java/org/apache/wicket/markup/head/JavaScriptHeaderItem.java
```diff
@@ -365,15 +372,54 @@ protected final void internalRenderJavaScriptReference(Response response, String
...
+		if (charset != null)
+		{
+			// XXX this attribute is not necessary for modern browsers
+			attributes.add("charset", charset);
+		}
+		if (async)
+		{
+			attributes.add(HeaderItemAttribute.SCRIPT_ASYNC, "async");
+		}
+		attributes.compute(HeaderItemAttribute.CSP_NONCE, this::getNonce);
```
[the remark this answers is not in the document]
andruhon: > I'm not ok with adding #noEscapeMarkupKeys to ValueMap.
> 
> Maybe ValueMap isn't the right class for this, note that you're still calling #add() when you should use #put() instead.
> Maybe it would be better to just add another "nonce" parameter to the JavaScriptUtils methods and postpone the attempt to clean-up its signature and delay the decision about 4777.

True. It's kind of ugly. There should be a way to render some parameters without escaping them, even if it's never done by default.
Honestly I don't quite like that toString does the magic. What do you think about adding a new method to IValueMap in wicket 9?

I would introduce two new methods, something like `#renderAttributes()` and `#renderAttributes(Collection<String> nonEscapedAttributes)` in IValueMap, so the map doesn't have to control them.

Somewhat like this:
```
	/**
	 * Generates a <code>String</code> representation of this object. All values are escaped by default.
	 *
	 * @param noEscapeMarkupKeys a collection of parameters which should not be escaped
	 * @return <code>String</code> representation of this <code>ValueMap</code> consistent with the
	 *         tag-attribute style of markup elements. For example: <code>a="x" b="y" c="z"</code>.
	 */
	public String renderAttributes(Collection<String> noEscapeMarkupKeys)
	{
		final StringBuilder buffer = new StringBuilder();
		boolean first = true;
		for (Map.Entry<String, Object> entry : entrySet())
		{
			if (first == false)
			{
				buffer.append(' ');
			}
			first = false;
			String key = entry.getKey();
			buffer.append(key);
			buffer.append("=\"");
			final Object value = entry.getValue();
			if (value == null)
			{
				buffer.append("null");
			}
			else if (value.getClass().isArray())
			{
				buffer.append(Arrays.asList((Object[])value));
			}
			else
			{
				buffer.append(
						!noEscapeMarkupKeys.isEmpty() && noEscapeMarkupKeys.contains(key)
								? value
								: Strings.escapeMarkup(String.valueOf(value))
				);
			}

			buffer.append('\"');
		}
		return buffer.toString();
	}

	/**
	 * Generates a <code>String</code> representation of this object. All values are escaped.
	 *
	 * @return <code>String</code> representation of this <code>ValueMap</code> consistent with the
	 *         tag-attribute style of markup elements. For example: <code>a="x" b="y" c="z"</code>.
	 */
	public String renderAttributes() {
		return renderAttributes(Collections.emptyList());
	}

	/**
	 * Generates a <code>String</code> representation of this object. All values are escaped.
	 *
	 * @return <code>String</code> representation of this <code>ValueMap</code> consistent with the
	 *         tag-attribute style of markup elements. For example: <code>a="x" b="y" c="z"</code>.
	 */
	@Override
	public String toString() {
		return renderAttributes();
	}
```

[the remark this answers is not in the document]
andruhon: Maybe escaping is just a wrong solution. This should be some kind of escaping, which doesn't destroy & characters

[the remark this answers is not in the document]
andruhon: org.apache.wicket.util.encoding.UrlEncoder or java.net.URLEncoder maybe?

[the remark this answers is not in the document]
andruhon: `buffer.append(UrlEncoder.QUERY_INSTANCE.encode(String.valueOf(value), "UTF-8"))` should solve WICKET-4777 and won't harm normal urls with parameters

andruhon:
I think this isn't going to be possible with wicket 8. Pointed this PR to master.
@svenmeier please have a look at recent changes with using url encoding to escape content.

andruhon:
I'm not sure that what I did with url encoding is correct. Essentially all attributes have slightly different rule. 
Most of attributes may content anything except the quote matching opening quote, the ID attribute can contain anything except spaces and blank line.

I think the ValueMap#toString should only take care of quotes when rendering and escape them as appropriate. All other special escaping should be done by the developer in the header items, when added by a developer.

andruhon:
Pushed another commit removing escaping from the ValueMap and delegating this job to appropriate header items

svenmeier:
I've merged this request into master with adjustments. Many thanks Andrew!

andruhon:
Thank you everyone!

PR #378, opened 2019-07-18
WICKET-6688 add RFC support (to avoid unsafe eval)

andruhon:
Initial commit addressing https://issues.apache.org/jira/browse/WICKET-6688.

Please have a look if it makes sense. If it is I can add documentation and do a cleanup.


wicket-core/src/main/java/org/apache/wicket/ajax/RemoteFunctionCallUtils.java
```diff
@@ -0,0 +1,43 @@
+package org.apache.wicket.ajax;
+
+import com.github.openjson.JSONObject;
+import org.apache.wicket.ajax.json.JSONFunction;
+import org.apache.wicket.ajax.json.JsonUtils;
+import org.apache.wicket.util.string.Strings;
+import org.junit.platform.commons.util.StringUtils;
```
solomax: I would not using junit in code other than tests

wicket-core/src/main/java/org/apache/wicket/ajax/RemoteFunctionCallUtils.java
```diff
@@ -0,0 +1,43 @@
...
+	 * @return
+	 */
+	public static String wrapAndEscape(Object arg)
+	{
+		if (arg instanceof Number || arg instanceof Boolean)
+		{
+			return String.valueOf(arg);
+		} else if (arg instanceof JSONObject) {
+			return arg.toString();
+		} else
```
solomax: Formatting + other `JSON*` type for ex. `JSONArray`

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/RFCPage.java
```diff
@@ -0,0 +1,151 @@
...
+		add(c2);
+
+		add(new AjaxLink<Void>("c1-link")
+		{
+			@Override
+			public void onClick(AjaxRequestTarget target)
+			{
+				counter1++;
+				target.add(c1);
+				target.prependRemoteFunctionCall("testFunc1Pre","something1", 1.5);
```
solomax: formatting

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/RFCPage.html
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<html xmlns:wicket="http://wicket.apache.org">
+<head>
+	<wicket:head>
+		<script language="JavaScript" type="text/javascript" src="jquery-ui-1.10.3.custom.js"></script>
```
solomax: this js seems to be not necessary

wicket-core/src/main/java/org/apache/wicket/ajax/RemoteFunctionCallUtils.java
andruhon: What's up with formatting?

solomax: curly braces should be on separate line i.e.
```
}
else if (arg instanceof JSONObject)
{
```
not
```
} else if (arg instanceof JSONObject) {
```

andruhon:
I went through comments.

andruhon:
(did a force push)

wicket-core/src/main/java/org/apache/wicket/ajax/RemoteFunctionCallUtils.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.wicket.util.string.Strings;
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+public class RemoteFunctionCallUtils
+{
+	public static String createFunctionJsonString(CharSequence functionName, Object... args)
+	{
+		return "{\"" +
```
solomax: I would change this to be:
```
return new JSONObject().put("func", functionName).put("args", new JSONArray().put(..........).toString();
```
IMO the code will be much cleaner
WDYT?

andruhon:
Hi! What's up with this one? Are there any alternative approaches to get rid of eval I can consider?

svenmeier:
I'll take another look.

svenmeier:
I've taken a second look and I'm all in favor of pursuing this improvement.

#appendJavaScript() and #appendRemoteFunctionCall() could be unified into a single method:

    target.appendJavaScriptCall("Window.eval", "Wicket.Ajax.ajax{...}");
    target.appendJavaScriptCall("Wicket.Ajax.RFC.testFunc1", "something2", 1);

But more importantly, in this current state of the proposal any Ajax refresh of any component breaks immediately due to blocking of eval(). Either we have to put much more effort into this and adapt most of Wicket's JS or this change is worthless.

andruhon:
I'm ready to contribute a substantial amount of time into resolving this issue.

My original concept was:
* add the RPC/RFC support
* create a new breed of CSP safe fetch based behaviors
* deprecate existing behaviors and let them die slowly over the years.

svenmeier:
Then we can put this pull-request on-hold until these fetch-based behaviors are available.

Until then eval() will be needed anyways, so there's no need to use this new RFC invocation.

andruhon:
Well, the whole point of this PR was that ajax *could* be used with strict CSP if proper behaviours used. (Old ajax behaviours using evals won't obviously work)

Let me merge back the master and see how it works together with nonce stuff.

andruhon:
Hi @svenmeier !

Now I see what do you mean. 

```Java
		final Label c1 = new Label("c1", LambdaModel.of(this::getCounter1)) {

			@Override
			public void renderHead(IHeaderResponse response)
			{
				super.renderHead(response);
				response.render(OnDomReadyHeaderItem.forScript("console.log('c1 rendered')"));
			}
		};
```

The code above will add
```JavaScript
<evaluate>
(function(){console.log('c1 rendered')})();
</evaluate>
```
to every response when `target.add(c1)` is called

This could be worked around if the JS is registered in advance with some name resolved from component name, however something like this:

```Java
		final Label c1 = new Label("c1", LambdaModel.of(this::getCounter1)) {

			@Override
			public void renderHead(IHeaderResponse response)
			{
				super.renderHead(response);
				if (Math.random() > 0.5)
				{
					response.render(OnDomReadyHeaderItem.forScript("console.log('c1 rendered more')"));
				} else {
					response.render(OnDomReadyHeaderItem.forScript("console.log('c1 rendered less')"));
				}
			}
		};
```
is not going to work. In this case both possible actions should be registered in advance.

What do you think?

svenmeier:
Yes, this concerns any JS that is added during an ajax call.
If a component has an attached AjaxBehavior, setting it visible is sufficient to trigger an JS evaluation - via AjaxCallListeners this might contain a plethora of JS snippets.

How would you 'register' all these?
Until each an every of these cases (in Wicket's code and all users' code) are migrated, you'll need &lt;evaluate&gt; and thus nobody will use this RFC feature.

andruhon:
@svenmeier please have a look at the proof of concept in my last commit.

Generally the issue is in header item being evaluated instead of added to the head, so what I did is made all header contributions in ajax to render these header items, also I added nonce support to the JS.

Essentially this can be used to replace all evals (as a legacy support feature). I don't think this approach is optimal, and ideally components should keep their JS in a JS file similar to what we do with properties and html file, these JS's will be rendered as header items and later used with some JS initialisation callback, or maybe with standard function.

Do you think it's a right direction?

cc @martin-g @solomax

solomax:
IMO we can't *force* user to use this or that technique
We only can propose sort of best-practices
Short inline JS is already being used everywhere (including multiple wicket based libraries)

I believe remove or reduce eval usage is good thing

svenmeier:
@andruhon
It's not clear who will clean-up all those header items. Can we garantee, thas this doesn't impact the browser performance (e.g. garbage collection)?
*If* this solution works, why bother with this RFC at all, and just use header items all the time?

andruhon:
> It's not clear who will clean-up all those header items. 
The header item wipes itself on the last line https://github.com/apache/wicket/pull/378/files#diff-a7f583254a200944d085608b18d15244R2153

> Can we garantee, thas this doesn't impact the browser performance (e.g. garbage collection)?
> _If_ this solution works, why bother with this RFC at all, and just use header items all the time?
The approach with header items is not particularly slow, however it's aimed to HEADER ITEMS. The direct call to predefined code is about 10 times faster than adding a new node in chrome. Please have a look at the exercise below:

```
    window.myNameSpace = {};
    window.myNameSpace.doSomething = function (){
      document.getElementById("myTestContainer").innerHTML = Date.now();
    };

    var cycles = 10000;
    var withDomStart = Date.now();
    for(var i = 0; i < cycles; i++) {
      var myScript = document.createElement("script");
      myScript.id = "myScr"+i;
      myScript.innerHTML = 'window.myNameSpace.doSomething(); document.getElementById("'+myScript.id+'").remove();';
      myScript.nonce = "4AEemGb0xJptoIGFP3Nd";
      document.head.appendChild(myScript);
    }
    var withDomTime = Date.now() - withDomStart;
    console.log("Execute with dom total, ms", withDomTime);
    console.log("Execute with dom average, ms", withDomTime/cycles);

    var withoutDomStart = Date.now();
    for(var i = 0; i < cycles; i++) {
      window.myNameSpace.doSomething();
    }
    var withoutDomTime = Date.now() - withoutDomStart;
    console.log("Execute without dom total, ms", withoutDomTime);
    console.log("Execute without dom average, ms", withoutDomTime/cycles);
```
The output is:
```
Execute with dom total, ms 786
Execute with dom average, ms 0.0786
Execute without dom total, ms 87
Execute without dom average, ms 0.0087
```

The difference is a way smaller if the function also creates and removes nodes, which usually happens when we add a new element onto the page. However, there must be a tidy way to call the function without touching the DOM at all (rfc/rpc).

martin-g:
would it be smarter if we reuse the `<script>` element and replace only its body ? would the browser re-evaluate the new body ? I haven't tested yet

andruhon:
@martin-g I thought about using the same JS. This could work if we set an id to this script tag. The question is how we going to identify them? Component id wouldn't work, because component could have several different header items added in renderHead. Any ideas?

svenmeier:
Thanks Andrew!

Actually the header items performance isn't so bad if we compare it to eval():
```
window.myNameSpace = {};
    window.myNameSpace.doSomething = function (){
      document.getElementById("myTestContainer").innerHTML = Date.now();
    };

    var cycles = 10000;
    var withDomStart = Date.now();
    for(var i = 0; i < cycles; i++) {
      var myScript = document.createElement("script");
      myScript.id = "myScr"+i;
      myScript.innerHTML = 'window.myNameSpace.doSomething(); document.getElementById("'+myScript.id+'").remove();';
      myScript.nonce = "4AEemGb0xJptoIGFP3Nd";
      document.head.appendChild(myScript);
    }
    var withDomTime = Date.now() - withDomStart;
    console.log("Execute with dom total, ms", withDomTime);
    console.log("Execute with dom average, ms", withDomTime/cycles);

    var withoutDomStart = Date.now();
    for(var i = 0; i < cycles; i++) {
      window.eval("window.myNameSpace.doSomething();");
    }
    var withoutDomTime = Date.now() - withoutDomStart;
    console.log("Execute without dom total, ms", withoutDomTime);
    console.log("Execute without dom average, ms", withoutDomTime/cycles);
    
    var withoutEvalStart = Date.now();
    for(var i = 0; i < cycles; i++) {
      window.myNameSpace.doSomething();
    }
    var withoutEvalTime = Date.now() - withoutEvalStart;
    console.log("Execute without eval total, ms", withoutEvalTime);
    console.log("Execute without eval average, ms", withoutEvalTime/cycles);
```
The output:
```
Execute with dom total, ms 2102
Execute with dom average, ms 0.2102
Execute without dom total, ms 1176
Execute without dom average, ms 0.1176
Execute without eval total, ms 49
Execute without eval average, ms
```

solomax:
`console.time` can be used for benchmarking :)
https://developer.mozilla.org/en-US/docs/Web/API/Console/time

svenmeier:
BTW just stumbled on https://api.jquery.com/jQuery.globalEval/ - note that it supports a nonce option.

That function just forwards to https://github.com/jquery/jquery/blob/master/src/core/DOMEval.js, interesting how jQuery removes the item immediately.

andruhon:
Ah, yes. Век живи — век учись (а умрешь дураком).

I think we can use jQuery.globalEval in wicket-ajax-jquery then. I'll push tidied code next week then. 

Have a nice weekend, everyone!

martin-g:
>  Век живи — век учись (а умрешь дураком)

well said! :-)

PR #380, opened 2019-08-14
WICKET-6558 no lock after detach

svenmeier:
A possible solution:
Session can prevent creating of locks after it has been detached for the end of the request.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Application.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Application.java
@@ -1569,23 +1569,12 @@ public abstract class Application implements UnboundListener, IEventSink, IMetad
 			@Override
 			public void onDetach(final RequestCycle requestCycle)
 			{
-				IPageManager pageManager;
-				
-				if (Session.exists())
-				{
-					pageManager = Session.get().getPageManager();
-				} else {
-					pageManager = internalGetPageManager();
-				}
-				pageManager.detach();
+				internalGetPageManager().detach();
 
-				if (Application.exists())
+				IRequestLogger requestLogger = getRequestLogger();
+				if (requestLogger != null)
 				{
-					IRequestLogger requestLogger = Application.get().getRequestLogger();
-					if (requestLogger != null)
-					{
-						requestLogger.requestTime((System.currentTimeMillis() - requestCycle.getStartTime()));
-					}
+					requestLogger.requestTime((System.currentTimeMillis() - requestCycle.getStartTime()));
 				}
 			}
 		});
--- a/wicket-core/src/main/java/org/apache/wicket/Session.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Session.java
@@ -925,7 +925,7 @@ public abstract class Session implements IClusterable, IEventSink, IMetadataCont
 	public final IPageManager getPageManager()
 	{
 		if (Boolean.TRUE.equals(RequestCycle.get().getMetaData(PAGES_UNLOCKED))) {
-			throw new WicketRuntimeException("pages have already been unlocked - synchronized access is no longer possible");
+			throw new WicketRuntimeException("The request has been processed. Access to pages is no longer allowed");
 		}
 		
 		IPageManager manager = Application.get().internalGetPageManager();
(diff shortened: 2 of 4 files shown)


wicket-core/src/main/java/org/apache/wicket/Application.java
```diff
@@ -1569,23 +1569,12 @@ public final RequestCycle createRequestCycle(final Request request, final Respon
...
-				IPageManager pageManager;
-				
-				if (Session.exists())
-				{
-					pageManager = Session.get().getPageManager();
-				} else {
-					pageManager = internalGetPageManager();
-				}
-				pageManager.detach();
+				internalGetPageManager().detach();
```
svenmeier: Sufficient to detach the page manager only. No need to detach the session here, RequestCycle#onDetach() does it already.

wicket-core/src/main/java/org/apache/wicket/Application.java
```diff
@@ -1569,23 +1569,12 @@ public final RequestCycle createRequestCycle(final Request request, final Respon
...
 
-				if (Application.exists())
```
svenmeier: Left over when this code was copied over from a request listener.

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -670,6 +676,9 @@ public void detach()
 	{
 		detachFeedback();
 
+		pageAccessSynchronizer.get().unlockAllPages();
+		RequestCycle.get().setMetaData(PAGES_UNLOCKED, true);
```
svenmeier: MetaData is set on request cycle, so no need to clear the flag.

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -915,6 +924,10 @@ public int nextPageId()
 	 */
 	public final IPageManager getPageManager()
 	{
+		if (Boolean.TRUE.equals(RequestCycle.get().getMetaData(PAGES_UNLOCKED))) {
```
svenmeier: When detached, Session no longer gives access to it's access-synchronizing manager.

wicket-core/src/main/java/org/apache/wicket/util/tester/BaseWicketTester.java
```diff
@@ -495,12 +494,21 @@ public void clearFeedbackMessages()
 	 */
 	protected void cleanupFeedbackMessages(IFeedbackMessageFilter filter)
 	{
-		ApplicationSettings applicationSettings = application.getApplicationSettings();
-		IFeedbackMessageFilter old = applicationSettings.getFeedbackMessageCleanupFilter();
-		applicationSettings.setFeedbackMessageCleanupFilter(filter);
-		getLastRenderedPage().detach();
-		getSession().detach();
-		applicationSettings.setFeedbackMessageCleanupFilter(old);
+
```
svenmeier: We cannot detach the session, otherwise it will block access to its page manager afterwards. Actually detachment isn't needed anyways, just the message have to be cleared. For symmetry don't call detach on components either.

seminolas:
Looks good to me. We had some custom logging asking for the name of the page class after the session had already been detached. We only noticed it in production with locked page exceptions - this would have made debugging much easier. 
@svenmeier I wonder if you can comment on why we only saw it very rarely. My theory was to do with unmounted pages: most of our pages are mounted and presumably resolving the page class uses some shortcut in code - without requiring a page lock. Resolving the name for unmounted pages does require a lock and caused the LockedPage exception on subsequent request. Does that sound reasonable? Our current fix only disables that (unnecessary) bit of logging -- and it seems to work -- but I wonder if I'm merely fixing a symptom of the real issue.

svenmeier:
Hard to tell what might increase the likelihood of that happening. Perhaps you can pinpoint a specific version upgrade?

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -915,6 +924,10 @@ public int nextPageId()
...
+			throw new WicketRuntimeException("pages have already been unlocked - synchronized access is no longer possible");
```
martin-g: I am not sure how understandable this message would be for the average developer. Maybe something like "The request [cycle] has been processed. Access to the pages is no more allowed" ?

PR #382, opened 2019-09-10
WICKET-6688 add RFC and replace eval with DOM eval

andruhon:
Addressing https://issues.apache.org/jira/browse/WICKET-6688.

Didn't update any tests yet. 

Differences from #378 :
* replaced `eval`s in wicket-ajax-jquery with jQuery.globalEval (DOM eval) with optional nonce
* introduced a new `meta` response item which currently only has nonse, which is used make all the old code with `appendJavaScript` work.

That's it old code can work with strict CSP, without a need to rewrite.

The RFC is still here, because it's a way faster, than eval or DOM eval. see https://github.com/apache/wicket/pull/378#issuecomment-525530622

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -341,6 +341,11 @@
 	Wicket.Ajax = {};
 
 	/**
+	 * A registry of remote functions for Wicket RFC
+	 */
+	Wicket.Ajax.RFC = {};
+
+	/**
 	 * Ajax call fires a Wicket Ajax request and processes the response.
 	 * The response can contain
 	 *   - javascript that should be invoked
@@ -823,15 +828,26 @@
 					return;
 				}
 
+				var nonce;
+				var meta = root.getElementsByTagName("meta")[0];
+				if (!isUndef(meta)) {
+					var nonceEl = meta.getElementsByTagName("nonce")[0];
+					if (!isUndef(nonceEl)) {
+						nonce = nonceEl.textContent;
+					}
+				}
+
 				var steps = context.steps;
 
 				// go through the ajax response and execute all priority-invocations first
 				for (var i = 0; i < root.childNodes.length; ++i) {
 					var childNode = root.childNodes[i];
 					if (childNode.tagName === "header-contribution") {
-						this.processHeaderContribution(context, childNode);
+						this.processHeaderContribution(context, childNode, nonce);
 					} else if (childNode.tagName === "priority-evaluate") {
-						this.processEvaluation(context, childNode);
+						this.processEvaluation(context, childNode, nonce);
+					} else if (childNode.tagName === "priority-rfc") {
+						this.processRemoteFunctionCall(context, childNode);
 					}
 				}
 
@@ -848,9 +864,11 @@
 						stepIndexOfLastReplacedComponent = steps.length;
 						this.processComponent(context, node);
 					} else if (node.tagName === "evaluate") {
-						this.processEvaluation(context, node);
+						this.processEvaluation(context, node, nonce);
 					} else if (node.tagName === "redirect") {
 						this.processRedirect(context, node);
+					} else if (node.tagName === "rfc") {
+						this.processRemoteFunctionCall(context, node);
 					}
 
 				}
@@ -929,9 +947,10 @@
 		/**
 		 * Adds a closure that evaluates javascript code.
 		 * @param context {Object} - the object that brings the executer's steps and the attributes
-		 * @param node {XmlElement} - the <[priority-]evaluate> element with the script to evaluate
+		 * @param node {Element} - the <[priority-]evaluate> element with the script to evaluate
+		 * @param nonce {String} - optional CSP nonce
 		 */
-		processEvaluation: function (context, node) {
+		processEvaluation: function (context, node, nonce) {
 
 			// used to match evaluation scripts which manually call FunctionsExecuter's notify() when ready
 			var scriptWithIdentifierR = new RegExp("\\(function\\(\\)\\{([a-zA-Z_]\\w*)\\|((.|\\n)*)?\\}\\)\\(\\);$");
@@ -956,7 +975,7 @@
 
 					try {
 						// do the evaluation in global scope
-						var f = window.eval(toExecute);
+						var f = jQuery.globalEval(toExecute, {nonce: nonce});
 						f(notify);
 					} catch (exception) {
 						log.error("Wicket.Ajax.Call.processEvaluation: Exception evaluating javascript: %s", text, exception);
@@ -970,7 +989,7 @@
 					// just evaluate the javascript
 					try {
 						// do the evaluation in global scope
-						window.eval(script);
+						var f = jQuery.globalEval(script, {nonce: nonce});
 					} catch (exception) {
 						log.error("Ajax.Call.processEvaluation: Exception evaluating javascript: %s", text, exception);
 					}
@@ -1007,10 +1026,32 @@
 			}
 		},
 
+		/**
+		 * execute Remote Function Call (RFC) with no eval
+		 * @param context {Object} - the object that brings the executer's steps and the attributes
+		 * @param node {Element} - the <[priority-]rfc> element with the script to evaluate
+		 */
+		processRemoteFunctionCall: function(context, node) {
+			// get the javascript body
+			var text = Wicket.DOM.text(node);
+			var functions = JSON.parse(text);
+			for (var f = 0; f < functions.length; f++) {
+				/**
+				 * @type {{func: string, args: []}}
+				 */
+				var fn = functions[f];
+				if (fn && fn.func && Wicket.Ajax.RFC[fn.func]) {
+					Wicket.Ajax.RFC[fn.func].apply(context, fn.args);
+				} else {
+					Wicket.Log.error("Remote functions Wicket.Ajax.RFC.%s does not exist");
+				}
+			}
+		},
+
 		// Adds a closure that processes a header contribution
-		processHeaderContribution: function (context, node) {
+		processHeaderContribution: function (context, node, nonce) {
 			var c = Wicket.Head.Contributor;
-			c.processContribution(context, node);
+			c.processContribution(context, node, nonce);
 		},
 
 		// Adds a closure that processes a redirect
@@ -1608,7 +1649,7 @@
 			 * Reads the text from the node's children nodes.
 			 * Used instead of jQuery.text() because it is very slow in IE10/11.
 			 * WICKET-5132, WICKET-5510
-			 * @param node {DOMElement} the root node
+			 * @param node {Element} the root node
 			 */
 			text: function (node) {
 				if (isUndef(node)) {
@@ -1789,7 +1830,7 @@
 				},
 
 				// Processes the parsed header contribution
-				processContribution: function (context, headerNode) {
+				processContribution: function (context, headerNode, nonce) {
 					var xmldoc = this.parse(headerNode);
 					var rootNode = xmldoc.documentElement;
 
@@ -1828,9 +1869,9 @@
 							if (name === "link") {
 								this.processLink(context, node);
 							} else if (name === "script") {
-								this.processScript(context, node);
+								this.processScript(context, node, nonce);
 							} else if (name === "style") {
-								this.processStyle(context, node);
+								this.processStyle(context, node, nonce);
 							} else if (name === "meta") {
 								this.processMeta(context, node);
 							}
@@ -1891,7 +1932,7 @@
 				},
 
 				// Process an inline style element
-				processStyle: function (context, node) {
+				processStyle: function (context, node, nonce) {
 					context.steps.push(function (notify) {
 						// if element with same id is already in document, skip it
 						if (Wicket.DOM.containsElement(node)) {
@@ -1905,6 +1946,7 @@
 
 						// copy id attribute
 						style.id = node.getAttribute("id");
+						style.nonce = nonce;
 
 						var textNode = document.createTextNode(content);
 						style.appendChild(textNode);
@@ -1917,7 +1959,7 @@
 				},
 
 				// Process a script element (both inline and external)
-				processScript: function (context, node) {
+				processScript: function (context, node, nonce) {
 					context.steps.push(function (notify) {
 
 						if (!node.getAttribute("src") && Wicket.DOM.containsElement(node)) {
@@ -1982,11 +2024,11 @@
 
 							if (typeof(id) === "string" && id.length > 0) {
 								// add javascript to document head
-								Wicket.Head.addJavascript(text, id, "", type);
+								Wicket.Head.addJavascript(text, id, "", type, nonce);
 							} else {
 								try {
 									// do the evaluation in global scope
-									window.eval(text);
+									jQuery.globalEval(text, {nonce: nonce});
 								} catch (e) {
 									Wicket.Log.error("Wicket.Head.Contributor.processScript: %s", text, e);
 								}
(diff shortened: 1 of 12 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -823,15 +828,26 @@
 					return;
 				}
 
+				var nonce;
+				var meta = root.getElementsByTagName("meta")[0];
+				if (!isUndef(meta)) {
+					var nonceEl = meta.getElementsByTagName("nonce")[0];
+					if (!isUndef(nonceEl)) {
+						nonce = nonceEl.textContent;
```
martin-g: does it make sense to use `Wicket.DOM.text(nonceEl)` here ?

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -1007,10 +1026,32 @@
...
+		 * @param context {Object} - the object that brings the executer's steps and the attributes
+		 * @param node {Element} - the <[priority-]rfc> element with the script to evaluate
+		 */
+		processRemoteFunctionCall: function(context, node) {
+			// get the javascript body
+			var text = Wicket.DOM.text(node);
+			var functions = JSON.parse(text);
+			for (var f = 0; f < functions.length; f++) {
+				/**
+				 * @type {{func: string, args: []}}
```
martin-g: are the extra curly braces needed ? they confuse me a bit

wicket-core/src/main/java/org/apache/wicket/core/request/handler/IPartialPageRequestHandler.java
```diff
@@ -78,6 +78,26 @@
 	 */
 	void appendJavaScript(CharSequence javascript);
 
+	/**
+	 * Add function for remote call execution on client side after components are replaced.
+	 * <p>
+	 * wicket-ajax-jquery will look for this function name inside of the Wicket.Ajax.RFC namespace
```
martin-g: How one register the function in this namespace ? This info would be very useful here.

wicket-core/src/main/java/org/apache/wicket/core/request/handler/IPartialPageRequestHandler.java
```diff
@@ -78,6 +78,26 @@
...
+	void appendRemoteFunctionCall(CharSequence functionName, Object... args);
+
+	/**
+	 * Add extra meta data to partial page response.
+	 * This is the response meta data, not HTML meta data.
+	 *
+	 * @param name meta datum name
+	 * @param value meta datum value
+	 */
+	void addMeta(CharSequence name, CharSequence value);
```
martin-g: may I call this method more than once ?

wicket-core/src/main/java/org/apache/wicket/page/PartialPageUpdate.java
```diff
@@ -68,6 +69,12 @@
 {
 	private static final Logger LOG = LoggerFactory.getLogger(PartialPageUpdate.class);
 
+	/**
+	 * Meta data to be added to the partial page response.
+	 * This is the response meta data, not HTML meta data.
+	 */
+	protected final List<CharSequence> meta = Generics.newArrayList();
```
martin-g: OK. I see I can call the `addMeta()` method more than once. But still I have no idea when to use this meta items. Some examples in the javadoc might be useful.

wicket-core/src/main/java/org/apache/wicket/page/PartialPageUpdate.java
```diff
@@ -657,34 +747,11 @@ public void render(HeaderItem item)
...
 				{
-					PartialPageUpdate.this.appendJavaScript(((OnEventHeaderItem) item).getCompleteJavaScript());
-					markItemRendered(item);
-				}
-			}
-			else if (item instanceof OnDomReadyHeaderItem)
-			{
-				if (!wasItemRendered(item))
-				{
-					if (priorityHeaderItem != null)
```
martin-g: Why the special handling of `priorityHeaderItem` is removed ?

wicket-core/src/main/java/org/apache/wicket/page/PartialPageUpdate.java
```diff
@@ -657,34 +747,11 @@ public void render(HeaderItem item)
...
-			if (item instanceof OnLoadHeaderItem)
-			{
-				if (!wasItemRendered(item))
-				{
-					PartialPageUpdate.this.appendJavaScript(((OnLoadHeaderItem) item).getJavaScript());
-					markItemRendered(item);
-				}
-			}
-			else if (item instanceof OnEventHeaderItem)
+			if (item instanceof OnLoadHeaderItem || item instanceof OnEventHeaderItem || item instanceof OnDomReadyHeaderItem)
```
martin-g: Why OnDomReadyHeaderItem is not handled with `PartialPageUpdate.this.domReadyJavaScripts.add()` anymore ?

wicket-examples/src/main/java/org/apache/wicket/examples/csp/CspApplication.java
```diff
@@ -45,9 +51,22 @@ protected void init()
...
+		setHeaderResponseDecorator(response -> new ResourceAggregator(
+				isCspApplicable() ? new CspNonceHeaderResponse(response, getNonce()) : response
+		));
+		// add nonce to ajax response
+		getAjaxRequestTargetListeners().add((new AjaxRequestTarget.IListener()
+		{
+			@Override
+			public void onBeforeRespond(Map<String, Component> map, AjaxRequestTarget target)
+			{
+				target.addMeta("nonce", getNonce());
```
martin-g: Should this be given some special name like `wicket-nonce` or `<wicket:nonce>` ? Because `IPartialPageUpdateHandler#addMeta()` can be called by the application and it can easily override this (add another) entry. Me, as a user, I'd like to be notified somehow when my application code breaks somehow the normal framework work.

wicket-examples/src/main/java/org/apache/wicket/examples/csp/CspApplication.java
```diff
@@ -69,4 +88,19 @@ public static String getNonce()
...
+	public static boolean isCspApplicable()
+	{
+		Request request = RequestCycle.get().getRequest();
+		if (request instanceof ServletWebRequest)
+		{
+			// Unfortunately Edge does things worse than just "doesn't support" it does support the CSP,
+			// but the 'nonce' and 'strict-dynamic' instructions were broken for ages.
+			// Edge issue https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13246371/
+			// It's OK in new Edge chromium beta, also the new Edge has Edg/ in User-Agent header instead of Edge/
+			return !((ServletWebRequest)request).getContainerRequest().getHeader("User-Agent").contains("Edge/");
```
martin-g: Does that mean that each and every application out there needs to add this snippet of code ?
One of the reasons why CSP has been postponed for few years is that the browser support was problematic for long time.

wicket-examples/src/main/java/org/apache/wicket/examples/csp/RfcPage.java
```diff
@@ -0,0 +1,180 @@
...
+						"var cnt = document.getElementById('resultsContainer');" +
+						"cnt.innerHTML = fn + ':' + JSON.stringify(args) + \"\\n\" + cnt.innerHTML;" +
+						"};" +
+						"Wicket.Ajax.RFC.testFunc1 = function(){otherGlobalFunction('testFunc1', arguments)};" +
+						"Wicket.Ajax.RFC.testFuncFunc = function(){otherGlobalFunction('testFuncFunc', arguments)};" +
+						"Wicket.Ajax.RFC.testFunc1Pre = function(){otherGlobalFunction('testFunc1Pre', arguments)};" +
+						"Wicket.Ajax.RFC.testFunc2 = function(){otherGlobalFunction('testFunc2', arguments)};" +
+						"Wicket.Ajax.RFC.testFunc2Pre = function(){otherGlobalFunction('testFunc2Pre', arguments)};" +
+						"eval('alert(\"this popup should not be displayed\")');"
+		));
```
martin-g: Let's rework this into a proper `.js` file that sets up the RFC entries. Users will follow this terrible/inline example.

wicket-examples/src/main/java/org/apache/wicket/examples/csp/CspApplication.java
andruhon: As of today - yes. But once new version of Edge is released 99.9 of buggy EdgeHTML/Chakra Edges supposed to be wiped out with windows update. It must happen pretty soon.

Otherwise works well in IE11, Chrome, Safari, Firefox and other modern browsers.

wicket-examples/src/main/java/org/apache/wicket/examples/csp/CspApplication.java
andruhon: Thanks @martin-g !

Honestly I'm not quite sure about this meta container. I was thinking about
1. placing a nonce as an attribute of `ajax-response`;
2. placing a nonce as a direct child of `ajax response`, without any kind of `meta` container;
3. adding some kind of universal container to add anything.

This one implements 3. What do you think maybe one of 1 or 2 is better?

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -956,7 +975,7 @@
 
 					try {
 						// do the evaluation in global scope
-						var f = window.eval(toExecute);
+						var f = jQuery.globalEval(toExecute, {nonce: nonce});
```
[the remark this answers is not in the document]
martin-g: I've written a blog post about this functionality: https://wicketinaction.com/2013/02/replace-components-with-animation/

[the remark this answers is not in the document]
andreikondratev: This kind of a black magic. 

Why can't this be done with something like this
```
function myHideFunction(id) {
  var el = $("#"+id);
  el.attr("id", id+"-anim").fadeOut(500);
  $("<span/>").attr("id", id).appendTo(el.parent());
}

function myShowFunction(id) {
  setTimeout(function() {
    var el = $("#"+id);
    el.fadeIn(500);
  }, 500);
}
```

or maybe
```
function mySubscriber(evt, id) {
  $("#"+id).fadeIn(500);
  Wicket.Event.unsubscribe("/myAnim/hidden/"+id, mySubscriber);
}

function myHideFunction(id) {
  var el = $("#"+id);
  el.attr("id", id+"-anim").fadeOut(500, function(){
    Wicket.Event.publish("/myAnim/hidden/"+id, id);
  });
  $("<span/>").attr("id", id).appendTo(el.parent());
}

function myShowFunction(id) {
  Wicket.Event.subscribe("/myAnim/hidden/"+id, mySubscriber);
}
```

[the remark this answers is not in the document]
andruhon: @martin-g This feels like kind of a black magic. 

Why can't this be done with something like this
```
function myHideFunction(id) {
  var el = $("#"+id);
  el.attr("id", id+"-anim").fadeOut(500);
  $("<span/>").attr("id", id).appendTo(el.parent());
}

function myShowFunction(id) {
  setTimeout(function() {
    var el = $("#"+id);
    el.fadeIn(500);
  }, 500);
}
```

or maybe
```
function mySubscriber(evt, id) {
  $("#"+id).fadeIn(500);
  Wicket.Event.unsubscribe("/myAnim/hidden/"+id, mySubscriber);
}

function myHideFunction(id) {
  var el = $("#"+id);
  el.attr("id", id+"-anim").fadeOut(500, function(){
    Wicket.Event.publish("/myAnim/hidden/"+id, id);
  });
  $("<span/>").attr("id", id).appendTo(el.parent());
}

function myShowFunction(id) {
  Wicket.Event.subscribe("/myAnim/hidden/"+id, mySubscriber);
}
```
?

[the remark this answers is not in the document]
martin-g: I cannot comment because now I do not understand how once should use your suggestion. Maybe now you should write a blog article ;-)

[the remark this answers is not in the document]
andruhon: @martin-g I don't really want to promote some dodgy non-streamline solution.

These are just examples of possible solutions for the problem:
https://github.com/andruhon/wicket-delayed-animation/tree/primitive-timeout-solution
https://github.com/andruhon/wicket-delayed-animation/tree/primitive-pubsub-solution
It has no wicket java, only wicket-ajax-jquery and a fake response in response.xml

Anyway I believe the solution should be javascripty, rather than something that has to be parsed. Maybe something like Wicket.Channel.pause() and Wicket.Channel.resume() if say we want to pause all ajax during this animation, or maybe some other methods *to be called from JS* to pause and notify particular request.

I'll think a little bit more.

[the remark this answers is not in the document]
andruhon: Hi @martin-g ! All right, I'm back with my dodgy experiments. I implemented the workaround to feed notify callback and parameters into global eval.

The change in wicket-ajax-jquery:
https://github.com/andruhon/wicket-delayed-animation/commit/f2aefa73902d89b9cfdae5f55fc2690c1b3282ac#diff-459f1fe1d12bc75eca45c5998e5dd29e

The branch in my quickstart project
https://github.com/andruhon/wicket-delayed-animation/tree/manual-notify-globalEval-closure

P.S.
Maybe I'm not very successful on expressing my thoughts. I do actually like the notify functionality itself, that's why I don't want to promote my non-streamline workarounds with timeouts. What I don't like is that it has to be added as a string `notify|` and then parsed. Why don't we add some kind of `notify="true"` attribute to the eval tag in response? What do you think?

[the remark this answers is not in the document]
svenmeier: IF we're using a global reference anyway, we can just get rid of that ugly parameter and allow scripts to suspend the execution explicitly:

https://gist.github.com/svenmeier/bd4375ab26e330617a834e1d53d86556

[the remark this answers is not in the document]
andruhon: @svenmeier I like idea. The one thing I don't understand why the count is needed? As far as I know it's either waiting or proceeding, so I think it won't get to the request (with a suspend) unless It's not suspended.

@svenmeier, @martin-g please have a look at this implementation:
https://github.com/andruhon/wicket-delayed-animation/commit/e799a1dfbd24fd35e0701d452c95cbf51f732186
Here I simply disallow to call suspend when evaluation is not happening, also it doesn't allow to call suspend twice, so I think boolean in this case is enough.

I added two buttons with slow 2.5s suspended fade out, you can try quickly clicking them both.

svenmeier:
The script might contain multiple functions, each requesting a separate suspension. So it's safer to count them.

andruhon:
Thanks @svenmeier I reproduced the case and added the count back. 
https://github.com/andruhon/wicket-delayed-animation/commit/b5d9408278376676d94ae1e4dc25f855846e7061

@svenmeier @martin-g do you think this is close to what we want to achieve? I'll create a new PR if yes.

andruhon:
Created a new one for globalEval #384

martin-g:
This issue can be closed now, right ?

andruhon:
Closed. Thank you!

PR #383, opened 2019-09-19
Configurable contentDisposition

espyrop:
ContentDisposition of the resourceStreamHandler is configurable, to allow inline display of the downloaded file.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/link/DownloadLink.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/link/DownloadLink.java
@@ -66,6 +66,14 @@ public class DownloadLink extends Link<File>
 	private Duration cacheDuration;
 
 	/**
+	 * Controls if the browser will save the file or display it inline.
+	 * <p>
+	 * The default is ATTACHMENT to maintain backward compatibility.
+	 * </p>
+	 */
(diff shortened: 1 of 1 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/html/link/DownloadLink.java
```diff
@@ -65,6 +65,14 @@
 	 */
 	private Duration cacheDuration;
 
+	/**
+	 * Controls if the browser will save the file or display it inline.
+	 * <p>
+	 * The default is ATTACHMENT to maintain backward compatibility.
```
martin-g: I am not sure there is a need for this in the javadoc. If we want to say what is the default then maybe something like: `The default is ATTACHMENT because this is what what <em>download</em> implies.`

espyrop: I have no objection to that. 
Since I am not familiar with the github workflow: Must I adjust this text, or you will do it?

svenmeier:
Thanks espyrop, I've applied the change with minor changes.

svenmeier:
Thanks @espyrop I've applied the change.
Please close this pull request.

PR #384, opened 2019-09-27
WICKET-6703 replace eval with jQuery.globalEval; add suspend/notify

andruhon:
https://issues.apache.org/jira/browse/WICKET-6703

Replaced `window.eval` with `jQuery.globalEval`, which allows to implement CSP policy without `unsafe-eval`

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -861,34 +861,27 @@
 				var steps = context.steps;
 
 				// go through the ajax response and execute all priority-invocations first
-				for (var i = 0; i < root.childNodes.length; ++i) {
-					var childNode = root.childNodes[i];
-					if (childNode.tagName === "header-contribution") {
-						this.processHeaderContribution(context, childNode, nonce);
-					} else if (childNode.tagName === "priority-evaluate") {
-						this.processEvaluation(context, childNode, nonce);
-					}
-				}
+				this.processHeaderContributions(context, root.childNodes, nonce, false);
 
 				// go through the ajax response and for every action (component, js evaluation, header contribution)
 				// ad the proper closure to steps
 				var stepIndexOfLastReplacedComponent = -1;
 				for (var c = 0; c < root.childNodes.length; ++c) {
 					var node = root.childNodes[c];
-
 					if (node.tagName === "component") {
 						if (stepIndexOfLastReplacedComponent === -1) {
 							this.processFocusedComponentMark(context);
 						}
 						stepIndexOfLastReplacedComponent = steps.length;
 						this.processComponent(context, node);
-					} else if (node.tagName === "evaluate") {
-						this.processEvaluation(context, node, nonce);
 					} else if (node.tagName === "redirect") {
 						this.processRedirect(context, node);
 					}
-
 				}
+
+				// go through the ajax response and execute all NON-priority-invocations
+				this.processHeaderContributions(context, root.childNodes, nonce, true);
+
 				if (stepIndexOfLastReplacedComponent !== -1) {
 					this.processFocusedComponentReplaceCheck(steps, stepIndexOfLastReplacedComponent);
 				}
@@ -961,52 +954,14 @@
 			});
 		},
 
-		/**
-		 * Adds a closure that evaluates javascript code.
-		 * @param context {Object} - the object that brings the executer's steps and the attributes
-		 * @param node {Element} - the <[priority-]evaluate> element with the script to evaluate
-		 * @param nonce {String} - optional CSP nonce
-		 */
-		processEvaluation: function (context, node, nonce) {
-
-			// get the javascript body
-			var text = Wicket.DOM.text(node);
-
-			// aliases to improve performance
-			var steps = context.steps;
-			var log = Wicket.Log;
-
-			var evaluate = function (script) {
-				return function(notify) {
-					// just evaluate the javascript
-					Wicket.Ajax.Call.currentNotify = notify;
-					try {
-						// do the evaluation in global scope
-						jQuery.globalEval(script, {nonce: nonce});
-					} catch (exception) {
-						log.error("Ajax.Call.processEvaluation: Exception evaluating javascript: %s", text, exception);
-					}
-					// continue to next step
-					if (Wicket.Ajax.Call._suspended) {
-						// suspended
-						return FunctionsExecuter.ASYNC;
-					} else {
-						// execution finished, cleanup the last notify
-						Wicket.Ajax.Call.currentNotify = undefined;
-						Wicket.Ajax.Call._suspended = 0;
-						// continue to next step
-						return FunctionsExecuter.DONE;
-					}
-				};
-			};
-
-			steps.push(evaluate(text));
-		},
-
-		// Adds a closure that processes a header contribution
-		processHeaderContribution: function (context, node, nonce) {
+		processHeaderContributions: function (context, nodes, nonce, afterComponents) {
 			var c = Wicket.Head.Contributor;
-			c.processContribution(context, node, nonce);
+			for (var i = 0; i < nodes.length; ++i) {
+				var childNode = nodes[i];
+				if (childNode.tagName === "header-contribution") {
+					c.processContribution(context, childNode, nonce, afterComponents);
+				}
+			}
 		},
 
 		// Adds a closure that processes a redirect
@@ -1785,7 +1740,7 @@
 				},
 
 				// Processes the parsed header contribution
-				processContribution: function (context, headerNode, nonce) {
+				processContribution: function (context, headerNode, nonce, afterComponents) {
 					var xmldoc = this.parse(headerNode);
 					var rootNode = xmldoc.documentElement;
 
@@ -1797,13 +1752,15 @@
 					// go through the individual elements and process them according to their type
 					for (var i = 0; i < rootNode.childNodes.length; i++) {
 						var node = rootNode.childNodes[i];
-
 						// Chromium reports the error as a child node
 						if (this._checkParserError(node)) {
 							return;
 						}
 
 						if (!isUndef(node.tagName)) {
+							if (afterComponents !== (node.getAttribute("data-wicket-evaluation") === "after")) {
+								continue;
+							}
 							var name = node.tagName.toLowerCase();
 
 							// it is possible that a reference is surrounded by a <wicket:link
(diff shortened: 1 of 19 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -350,6 +350,44 @@
...
+		}
+	};
+
+	Wicket.Ajax.Call.getFunctionsExecuterStatus = function() {
+		if (Wicket.Ajax.Call._suspended) {
+			// suspended
+			return FunctionsExecuter.ASYNC;
+		} else {
+			// execution finished, cleanup the last notify
+			Wicket.Ajax.Call.currentNotify = undefined;
```
svenmeier: currentNotify has to be cleared in either case, there's no need to keep it around any longer. Either a suspension was aquired at this point or not.
Actually we have to move the _suspended counter into the scope of currentNotify, otherwise we run intro troubles with parallel Ajax requests.

andruhon: ✔️ changed

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -350,6 +350,44 @@
...
 	Wicket.Ajax.Call = Wicket.Class.create();
 
+	Wicket.Ajax.Call._suspended = 0;
+
+	Wicket.Ajax.Call.currentNotify = undefined;
+
+	Wicket.Ajax.Call.suspend = function () {
+		if (typeof (Wicket.Ajax.Call.currentNotify) != "function") {
+			Wicket.Log.error("Can't suspend: no evaluation in process");
+			return;
```
martin-g: IMO this should either log error and return a no-op function or throw an exception instead of just logging.

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -350,6 +350,44 @@
...
+		}
+		Wicket.Ajax.Call._suspended++;
+		var notify = Wicket.Ajax.Call.currentNotify;
+		var released = false;
+
+		return function () {
+			// release only once
+			if (released === false) {
+				released = true;
```
martin-g: If `suspend()` is called multiple times (i.e. `_suspended` is for example `3`) then executing the `notify` (the returned) function will set `released` to `true` and `_suspended` to 2. I see no way to decrement `_suspended` to `0` from there on

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -350,6 +350,44 @@
...
+				Wicket.Ajax.Call._suspended--;
+				if (Wicket.Ajax.Call._suspended === 0) {
+					notify();
+				}
+			}
+		}
+	};
+
+	Wicket.Ajax.Call.getFunctionsExecuterStatus = function() {
+		Wicket.Ajax.Call.currentNotify = undefined;
```
martin-g: Why ?!
This is a huge side effect here.
Calling `getStatus()` should not unset such important state.
Also I think `getFunctionsExecuterStatus` should not be public, i.e. it should not be possible for the application code to be able to execute it at all. It shouldn't be a member of `Call`

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -823,15 +861,25 @@
...
 				}
 
+				var nonce;
+				var meta = root.getElementsByTagName("meta")[0];
+				if (!isUndef(meta)) {
+					// var nonceEl = meta.getElementsByTagName("wicket-nonce")[0];
+					// if (!isUndef(nonceEl)) {
+					// 	nonce = Wicket.DOM.text(nonceEl);
+					// }
+					nonce = Wicket.DOM.text(meta.getElementsByTagName("wicket-nonce")[0]);
```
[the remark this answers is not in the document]
andruhon: Hi! nonce attribute is what I did in my implementation without backend. However I add the nonce with 
```
**		// add nonce to ajax response
		getAjaxRequestTargetListeners().add((new AjaxRequestTarget.IListener()
		{
			@Override
			public void onBeforeRespond(Map<String, Component> map, AjaxRequestTarget target)
			{
				target.addMeta("wicket-nonce", getNonce());
			}
		}));**
```
In CspApplication.java

Modifying ajax-response tag would be less trivial from here. Do you want me to dig in this direction? Maybe separate tag like `<wicket-nonce>` is OK?

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
andruhon: Every notify returned should be called, for example this one works:

```
function myHideFunction(id) {
  var notify1 = Wicket.Ajax.Call.suspend();
  var notify = Wicket.Ajax.Call.suspend();
  $("#"+id).fadeOut(3500, function(){
    console.log("fade out done");
    notify();
    notify1();
  });
  otherSuspendingFunction();
}

function otherSuspendingFunction() {
  var notify = Wicket.Ajax.Call.suspend();
  setTimeout(function () {
    console.log("otherSuspendingFunction done");
    notify(); // attempts to notify before the fadeOut finished
  }, 1000);
}
```
But if you any of these three notifys will get lost the stuff will get stuck.

svenmeier:
If every JS snippet is ending up as a header item anyway, why not remove &lt;priority-evaluate&gt; and &lt;evaluate&gt; from the Ajax response and transport everything as a &lt;header-contribution&gt;?
In the correct order of course.

&lt;script&gt; and &lt;style&gt; already have a nonce-attribute, so we don't need to invent something new to transport that information in the Ajax response.
With WICKET-6703 we will have 4 paths* in wicket-ajax-jquery.js to add &lt;script&gt; to the DOM, which could all be unified:
```
[<header-contribution>]
-> processHeaderContribution	
  -> Wicket.Head.Contributor.processContribution
    -> processScript
      [has "src"]
        -> Wicket.Head.addElement
      [has "id"]
        -> addJavascript
          -> Wicket.Head.addElement
      [otherwise]
        -> jQuery.globalEval

[<priority-evaluate> or <evaluate>]
-> processEvaluation
  -> jQuery.globalEval
```

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
andruhon: I think this conflicts with https://github.com/apache/wicket/pull/384#discussion_r331761728
the currentNotify should be called either when the call is not async, or when last suspend/lock is released.

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
andruhon: Right, I'll move this logic into process evaluation

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
andruhon: I'll go with throw

andruhon:
> If every JS snippet is ending up as a header item anyway, why not remove <priority-evaluate> and <evaluate> from the Ajax response and transport everything as a <header-contribution>?
> In the correct order of course.
> 
> <script> and <style> already have a nonce-attribute, so we don't need to invent something new to transport that information in the Ajax response.
> With WICKET-6703 we will have 4 paths* in wicket-ajax-jquery.js to add <script> to the DOM, which could all be unified:
> 
> ```
> [<header-contribution>]
> -> processHeaderContribution	
>   -> Wicket.Head.Contributor.processContribution
>     -> processScript
>       [has "src"]
>         -> Wicket.Head.addElement
>       [has "id"]
>         -> addJavascript
>           -> Wicket.Head.addElement
>       [otherwise]
>         -> jQuery.globalEval
> 
> [<priority-evaluate> or <evaluate>]
> -> processEvaluation
>   -> jQuery.globalEval
> ```

I think this could work. Just before I start with this, what do others think? @martin-g ? @solomax ?

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
andruhon: I fixed this, please see a new EffectsPage.js example

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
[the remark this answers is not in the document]
martin-g: `<wicket-nonce>` looks good to me!

martin-g:
Sounds good to me! In this case there won't be a need for `<meta><wicket-nonce/></meta>`

solomax:
+1 sounds very clear and straightforward :)

andruhon:
All right, I've implemented the stuff discussed above. All "evaluate" and "priority-evaluate" tags are now gone from AJAX XML as well as meta and wicket-nonce. I had to add a new attribute to JS header contributions to mark those which must be rendered after components.

andruhon:
Any thoughts on this? Should I proceed with cleaning up tests?

svenmeier:
I've taken a short look only, but I have some changes already. It try to find some time soon.

svenmeier:
IMHO I've found a better solution, see https://github.com/apache/wicket/commits/WICKET-6703-replace-eval-with-domEval

By taking the ordering of the items in <ajax-response> into account, we can simplify this a lot.

andruhon:
@svenmeier I had a look at the code. It all makes sense. I'll test it soon.

svenmeier:
Actually your data-wicket-evaluation=after made me wonder whether we could just make the ordering relevant. It's just a script of instructions that is processed in JS.

andruhon:
`target.prependJavaScript((String.format("jQuery('#%s').fadeOut(500, Wicket.Ajax.suspendCall());", c3.getMarkupId())));` this line in demo certainly needs some explanation. At a first glance it's not very obvious where the suspension is released.

Somewhat like this:
`// i.e. suspendCall suspends the ajax queue here, the suspendCall itself returns a function which does release a lock once called. The fadeOut will call this function when animation is finished and the lock will be released.`
Potentially the notify can be pulled into variable to make things clear.

Also, I think it would be nice to have a demo with multiple locks.

svenmeier:
@andruhon I've seen that you preferred to put this into a separate js file. Personally I'd rather have this concise and *cool* :P

Regarding multiple locks - I don't know when/why this would actually be useful.

martin-g:
> Regarding multiple locks - I don't know when/why this would actually be useful.

Do we (want to) support it ?
If YES then let's test it whether it works.
If NO then let's document it and/or throw an error in JS code when a second suspend is called.

svenmeier:
Many thanks Andrew, we've got this enhancement into master.
Please close this pull request.

andruhon:
🎉 🎉 🎉
Thank you everyone! Have a great weekend!

PR #385, opened 2019-09-30
WICKET-6704 JavaSerializer.serialize causes the JVM crash !

martin-g:
Do not check instances of PropertyChangeSupport whether they are Serializable because PropertyChangeSupport#writeObject() adds extra fields which confuse CheckingObjectOutputStream


martin-g:
This PR fixes the problem in WICKET-6704 but there are more classes in Java which do the same as PropertyChangeSupport, e.g. java.security.BasicPermission, java.util.BitSet, and one dozen more. Should we whitelist them too ? Or someone sees how to fix the problem in CheckingObjectOutputStream ?

PR #394, opened 2019-10-24
Fix JavaDoc comment for PackageResource.getResourceStream()

ursjoss:
In the current implementation, the return value can never be null.


martin-g:
Why do you think it cannot be null ? 
`org.apache.wicket.core.util.resource.locator.IResourceStreamLocator#locate(java.lang.Class<?>, java.lang.String, java.lang.String, java.lang.String, java.util.Locale, java.lang.String, boolean)` can return null. PackageResource will just wrap it in ProcessingResourceStream if it not null.

ursjoss:
@martin-g Yes, you're completely right of course. Sorry for the noise...

PR #399, opened 2020-01-17
WICKET-6727: Configurable Content-Security-Policy

papegaaij:
I'm opening this PR to get some feedback on the suggested changes. The structure of the code is what I have in mind for this. I do know the documentation is still lacking a bit, but I plan on updating that when the code is ready to be merged.

The most important notes are:
- CSP is now enabled by default, but still with the unsafeEval() profile. I've clicked through several examples, and they worked fine. It does complain about a font css that cannot be loaded.
- CSP is part of WebApplication and configured via `getCSP()`. The header response decorator is automatically registered and will add nonces when required. The previous header decorator is removed.
- The CSP is rendered as HTTP header on every Page request, rather than in the markup. The nonce is changed on every request.
- websocket-core is broken, because `setHeader` throws an `UnsupportedOperationException`. I'm not yet sure on how to fix this, suggestions are welcome.


wicket-core/src/main/java/org/apache/wicket/csp/CSPDirective.java
```diff
@@ -0,0 +1,191 @@
...
+	IMG_SRC("img-src"),
+	CONNECT_SRC("connect-src"),
+	FONT_SRC("font-src"),
+	OBJECT_SRC("object-src"),
+	MANIFEST_SRC("manifest-src"),
+	MEDIA_SRC("media-src"),
+	CHILD_SRC("child-src"),
+	FRAME_ANCESTORS("frame-ancestors"),
+	@Deprecated
+	/** @deprecated Gebruik CHILD-SRC, deze zet ook automatisch FRAME-SRC. */
```
solomax: I believe this have to be translated

wicket-core/src/main/java/org/apache/wicket/csp/CSPDirective.java
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	/**
+	 * @return The CSPDirective constant whose value-parameter equals the input-parameter or
+	 *         {@code null} if none can be found.
+	 */
+	public static CSPDirective fromValue(String value)
+	{
+		if (Strings.isEmpty(value))
+			return null;
```
solomax: missing curly braces?
(also in some other places)

solomax:
@papegaaij we are currently adding CSP header using following conditions:
```
if (resp instanceof WebResponse && !(resp instanceof WebSocketResponse)) {
}
```

papegaaij:
That won't work because that class is not on the classpath. Also I don't like checking on subclasses to exclude them. What if another subclass is added by a user that also does not support headers? Maybe WebResponse needs a `isHeadersSupported`?

solomax:
`canSetHeader` ?

papegaaij:
I've resolved the issues with `setHeader`, updated the documentation and fixed some minor issues. I've whitelisted some resources in the examples, such as the fonts from `maxcdn.bootstrapcdn.com`. For the websockets examples, I had to whitelist several google URIs. This is not good practice, but it's just an example. It could be that some examples require other directives as well.

solomax:
API looks good to me
I'm trying to add CSP support from this branch to out main wicket app
The biggest problem so far: first request is being processed with ~5 minutes delay, I'll investigate and will report back

wicket-core/src/test/java/org/apache/wicket/csp/CSPSettingRequestCycleListenerTest.java
```diff
@@ -0,0 +1,477 @@
...
+import org.apache.wicket.util.tester.WicketTestCase;
+import org.apache.wicket.util.tester.WicketTester;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+@SuppressWarnings("deprecation")
+public class CSPSettingRequestCycleListenerTest extends WicketTestCase
+{
+	private static String HEADER_CSP = "Content-Security-Policy";
```
martin-g: maybe use CSPHeaderMode.BLOCKING.getHeader() instead ?

wicket-core/src/test/java/org/apache/wicket/csp/CSPSettingRequestCycleListenerTest.java
```diff
@@ -0,0 +1,477 @@
...
+
+		StringBuffer headerErrors = checkHeaders(cspListener);
+
+		if (headerErrors.length() > 0)
+		{
+			Assertions.fail(headerErrors.toString());
+		}
+	}
+
+	// FF 36+, IE (incl. Edge), Safari en Opera Mini hebben nog geen (volledige)
```
martin-g: OMG Dutch! :-)

wicket-core/src/test/java/org/apache/wicket/csp/CSPSettingRequestCycleListenerTest.java
```diff
@@ -0,0 +1,477 @@
...
+
+		if (headerErrors.length() > 0)
+		{
+			Assertions.fail(headerErrors.toString());
+		}
+	}
+
+	private StringBuffer checkHeaders(ContentSecurityPolicyEnforcer cspListener)
+	{
+		StringBuffer headerErrors = new StringBuffer();
```
martin-g: nit: StringBuilder ?! Or even better `List<String>`

wicket-examples/src/main/java/org/apache/wicket/examples/WicketExampleApplication.java
```diff
@@ -57,5 +58,7 @@ protected void init()
 		getDebugSettings().setDevelopmentUtilitiesEnabled(true);
 		
 		getResourceSettings().setCssCompressor(new CssUrlReplacer());
+		getCsp().blocking().add(CSPDirective.STYLE_SRC, "https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css")
+				.add(CSPDirective.FONT_SRC, "https://maxcdn.bootstrapcdn.com");
```
martin-g: is this enough ? The font path is not specified.
Maybe we should do the same for STYLE_SRC ?! Updating the version to a newer one will lead to changing it in several places

papegaaij: You can use hosts, with a path in CSP and it will match all resources on that host. The problem in this case is that this is a CDN. It could deliver anything. It is better not to use CDNs at all. With HTTP2 it adds latency to your page loads, because multiple http connections need to be established and it makes it very difficult to protect your site with CSP.

In this particular case I chose to only allow one css file, but all fonts from the CDN. This makes it impossible to inject styling and therefore very hard to load fake fonts as well.

papegaaij:
> API looks good to me
> I'm trying to add CSP support from this branch to out main wicket app
> The biggest problem so far: first request is being processed with ~5 minutes delay, I'll investigate and will report back

@solomax did you find the cause of the slowdown? Does it have to do with the CSP, perhaps the random?

wicket-core/src/main/java/org/apache/wicket/csp/CSPDirective.java
```diff
@@ -0,0 +1,208 @@
...
+				{
+					throw new IllegalArgumentException(
+						"A sandbox directive can't contain other values if it already contains an empty string");
+				}
+			}
+
+			if (!(value instanceof CSPDirectiveSandboxValue))
+			{
+				throw new IllegalArgumentException(
+					"A sandbox directive can only contain values from CSPDirectiveSandboxValue or be empty");
```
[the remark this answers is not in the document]
papegaaij: I've added the values to most errors

wicket-core/src/main/java/org/apache/wicket/csp/CSPRenderable.java
```diff
@@ -0,0 +1,42 @@
...
+	/**
+	 * Renders the value that should be put in the CSP header.
+	 * 
+	 * @param listener
+	 *            The {@link ContentSecurityPolicyEnforcer} that renders this value.
+	 * @param cycle
+	 *            The current {@link RequestCycle}.
+	 * @return The rendered value.
+	 */
+	public String render(ContentSecurityPolicyEnforcer listener, RequestCycle cycle);
```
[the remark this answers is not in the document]
papegaaij: The `render(null, null)` call was a bad idea. The nonce cannot be rendered without those values and will give an NPE. I've change the API to no longer need this call. IMHO the code read better now.

wicket-core/src/main/java/org/apache/wicket/csp/ContentSecurityPolicyEnforcer.java
```diff
@@ -0,0 +1,173 @@
...
+	{
+		return configs.values().stream().anyMatch(CSPHeaderConfiguration::isNonceEnabled);
+	}
+
+	public String getNonce(RequestCycle cycle)
+	{
+		String nonce = cycle.getMetaData(NONCE_KEY);
+		if (nonce == null)
+		{
+			nonce = getSecuritySettings().getRandomSupplier().getRandomBase64(12);
```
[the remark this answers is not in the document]
papegaaij: changed it into a constant. I don't think it's needed to make this configurable.

wicket-core/src/test/java/org/apache/wicket/csp/CSPSettingRequestCycleListenerTest.java
papegaaij: That's what you get when you copy large bodies of code from a Dutch codebase :) Some Dutch comments slipped through.

wicket-examples/src/main/java/org/apache/wicket/examples/WicketExampleApplication.java
martin-g: If a CDN delivers `anything` but the expected content then this CDN will be out of the game in no time.

The benefit of using CDNs is that there is a bigger chance the resource to be already in the browser cache. How exactly HTTP2 invalidates this benefit ?

solomax:
@papegaaij I'm still investigating
(suspectingthis might be caused by lots of `nonce`s need to be generated, and/or lot's of inline styles)
I need to change lot's of inline styles and js fragments
Will try to test against `csp` branch

wicket-examples/src/main/java/org/apache/wicket/examples/WicketExampleApplication.java
papegaaij: You can trust the CDN to deliver the font-awesome css when you request that URL, but you can't just whitelist a whole CDN, because it may also contain scripts and styles you absolutely do not want. Strictly speaking this also holds for fonts, but loading a strange font is very hard to do when you can't inject css and exploiting an application via fonts is even harder. That's why I chose to whitelist just that one CSS file for styling but the whole CDN for fonts.

With regards to HTTP2, that's a whole different discussion, but HTTP2 allows multiplexing multiple requests over a single connection. This reduces your startup time for loading a page. Any content served from the same location as the main page can be loaded directly over the same connection. For a CDN, a new HTTP connection has to be setup (when it cannot be loaded from the cache). This is also why the old technique of combining many resources into one bundle doesn't help much with HTTP2.

solomax:
@papegaaij, All
I was able to find the reason of slow down in my application
`SecureRandom.getInstanceStrong()` creates `NativePRNGBlocking` SecureRandom on my machine (Ubuntu)
I have changed the code to be `SecureRandom.getInstance("SHA1PRNG");` and it works as usual (no slow-down)

martin-g:
@solomax Is it better with `-Djava.security.egd=file:/dev/./urandom` ?

solomax:
just tried to start Tomcat with `JAVA_OPTS="-Djava.security.egd=file:/dev/./urandom" ./bin/catalina.sh run`
it is slow :(

papegaaij:
Good random data with a reliable entropy source always is a problem. On my system (kubuntu), I've got haveged running to increase the available entropy. SHA1PRNG is ok, but the native version is better. `getInstanceStrong` is suitable for key generation for long term storage. I'm ok with changing this to SHA1PRNG. Users with special need can change the implementation if they want.

solomax:
The main issue here: such slow down will appear with upgrade (with zero line of code changed)
So I guess it is better to use faster version by default (and add special section to javadocs/mirgation/docs)

solomax:
@papegaaij in our code we are using code like this:
https://github.com/apache/openmeetings/blob/csp/openmeetings-web/src/main/java/org/apache/openmeetings/web/pages/BasePage.html#L28

special CSS `custom.css` is editable by our users and placed last (after `<wicket:header-items/>`)
Are there any options to `auto` add this to CSP?
(Users might be able to generate `sha512`, but I would like to make this invisible somehow)

papegaaij:
@solomax I agree about the `SecureRandom`. `SHA1PRNG` is good enough for most usecases. And we need a generator that does not cause problems in simple applications like a quickstart.

For the custom.css, you can whitelist that in the CSP by adding that particular stylesheet or `'self'` to the `style-src` directive.

papegaaij:
I've replaced the random implementation on master, because WICKET-6730 was already merged to master. I would like to merge this branch without enabling CSP for now. I've created WICKET-6733 for that. Before we can enable CSP, we must make sure Wicket does not contain any inline styling and js and the examples all work fine.

martin-g:
> special CSS `custom.css` is editable by our users and placed last (after `<wicket:header-items/>`)
> Are there any options to `auto` add this to CSP?

@solomax Maybe you can use FilteredHeaderItem for it ?

solomax:
@martin-g I'm trying to implement this via FilteredHeaderItem, but this seems to be extremely not easy:

CSSItem added as `FilteredHeaderItem` is being processed by `IHeaderResponseFilter` as `CSSUrlReferenceHeaderItem`, so all info about `filterName` is lost

will spend some time on this later today

Thanks for the pointer :))

martin-g:
There might be a bug somewhere :-/

solomax:
Yep, seems to be bug
Filtered nature is dropped here https://github.com/apache/wicket/blob/csp/wicket-core/src/main/java/org/apache/wicket/csp/CspNonceHeaderResponseDecorator.java#L48

@papegaaij could you please take a look?

papegaaij:
@solomax you are right, the item is unwrapped to set the nonce, but then the unwrapped item is rendered. I've fixed the code, please try again.

solomax:
@papegaaij Thanks a lot!
Everything works as expected :)

ps surprisingly my original question wasn't offtopic in this PR :)

papegaaij:
I just noticed a problem with using a different nonce for every response. strict-dynamic only works for scripts and not for stylesheets. This means that a nonced script can load other scripts, but it cannot load additional css. Threfore, I decided to bind the nonce to a page and use the same nonce throughout the lifetime of the page.

In addition, I'm working on reporting CSP violation back to the application and log them.

andruhon:
Hi Emond

I also had difficulties with nonce per request and went with per-session
solution.

papegaaij:
@andruhon Per page to works fine in my application.

solomax:
Hello @papegaaij,
I found some weird behavior with CSP

In my app I'm setting `reporting csp` as follows: `getCsp().reporting().strict()`
Then I'm adding some images to allowed:
`getCsp().reporting().add(CSPDirective.IMG_SRC, "URL HERE")`

as a result I _surprisingly_ have both `Content-Security-Policy` and `Content-Security-Policy-Report-Only` set
images are allowed in `Content-Security-Policy-Report-Only` but blocked by `Content-Security-Policy`

I do expect blocking CSP will not be set or rules should be shared ....

papegaaij:
Hi @solomax,

The `reporting()` method is to configure the report only header. It must be separate from the blocking header, because you need to be able to specify different policies for both. This can be used to test a more strict policy while the old policy is still in effect.

It is not very useful to share the rules between the 2 headers, because the blocking header can have a reporting directive as well. Perhaps the documentation on these methods can be improved?

solomax:
@papegaaij the main problem: `reportong` method was the only **configured one**, but surprisingly response contains both `blocking` and `reporting`
if this is by design - rules should be shared (IMO)

papegaaij:
@solomax this is because the blocking header is already configured in`WebApplication`. If you do not want a blocking CSP, you need to call `getCSP().blocking().disable()`. Otherwise you are configuring the reporting header next to the blocking one. I can see how this can be a bit confusing, but I don't see an easy fix, do you?

solomax:
@papegaaij you are right
sorry for the noise :(

The only way I see this can be improved is create "reporting-only" example with `getCSP().blocking().disable()` and comment describing why this is necessary

solomax:
BTW while using this version I'm seeing multiple issues which cannot be targeted easily due to components containing inline styles are received via AJAX and all errors are being reported inside `jquery`
Maybe you know how detailed info on what is wrong can be retrieved in such case?

papegaaij:
I don't know, the report contains a script-sample, but I've never seen it filled. It would have been nice if the reporting was a bit more verbose in some cases.

solomax:
@papegaaij is it time to merge this PR into master?

papegaaij:
@solomax I was planning to wrap this up today. I want to run some more tests with our application, now that most issues in wicket itself are fixed.

PR #400, opened 2020-01-21
WICKET-6725: replace display: none by css class

papegaaij:
This PR replaces inline display: none with a CSS class and a stylesheet. The stylesheet is managed by Wicket and added automatically. You can change or disable this stylesheet via `ResourceSettings`. I've also disabled the stylesheet in `MockApplication` to prevent noise in our tests.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/mock/MockApplication.java
+++ b/wicket-core/src/main/java/org/apache/wicket/mock/MockApplication.java
@@ -71,5 +71,7 @@ public class MockApplication extends WebApplication
 
 		// for test cases we usually want stable resource names
 		getResourceSettings().setCachingStrategy(NoOpResourceCachingStrategy.INSTANCE);
+		// the core CSS causes noise in tests and isn't needed
+		getResourceSettings().disableWicketCoreCSS();
 	}
 }
--- a/wicket-core/src/main/java/org/apache/wicket/protocol/http/WebApplication.java
+++ b/wicket-core/src/main/java/org/apache/wicket/protocol/http/WebApplication.java
@@ -21,7 +21,6 @@ import java.nio.charset.Charset;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.Locale;
-import java.util.Optional;
 import java.util.function.Function;
 import javax.servlet.ServletContext;
 import javax.servlet.http.HttpServletRequest;
@@ -761,11 +760,9 @@ public abstract class WebApplication extends Application
 		getAjaxRequestTargetListeners().add(new AjaxEnclosureListener());
 		
 		getHeaderContributorListeners().add(head -> {
-			Optional<CssResourceReference> wicketCoreCSS = getResourceSettings().getWicketCoreCSS();
-			if (wicketCoreCSS.isPresent())
-			{
-				head.render(CssHeaderItem.forReference(wicketCoreCSS.get()));
-			}
+			getResourceSettings().getWicketCoreCSS().ifPresent(wicketCoreCSS -> {
+				head.render(CssHeaderItem.forReference(wicketCoreCSS));
+			});
 		});
 
 		// Configure the app.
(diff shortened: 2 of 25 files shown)


wicket-core/src/main/java/org/apache/wicket/protocol/http/WebApplication.java
```diff
@@ -758,6 +759,14 @@ protected void internalInit()
 		setAjaxRequestTargetProvider(AjaxRequestHandler::new);
 
 		getAjaxRequestTargetListeners().add(new AjaxEnclosureListener());
+		
+		getHeaderContributorListeners().add(head -> {
+			Optional<CssResourceReference> wicketCoreCSS = getResourceSettings().getWicketCoreCSS();
```
martin-g: personal preference: use `Optional.ifPresent()`

wicket-core/src/main/java/org/apache/wicket/settings/ResourceSettings.java
```diff
@@ -770,4 +777,51 @@ public ResourceSettings setEncodeJSessionId(boolean encodeJSessionId)
...
+	 * @param wicketCoreCSS
+	 *            The replacement styleheet.
+	 * @return {@code this} object for chaining
+	 */
+	public ResourceSettings setWicketCoreCSS(CssResourceReference wicketCoreCSS)
+	{
+		if (wicketCoreCSS == null)
+		{
+			throw new NullPointerException(
+				"Cannot set the Wicket core CSS to null, use disableWicketCoreCSS() instead.");
```
martin-g: why not ? `Optional.ofNullable()` is just fine to me.

wicket-core/src/test/java/org/apache/wicket/markup/html/basic/RedirectPage-expected1.html
```diff
@@ -0,0 +1,5 @@
+<html>
```
martin-g: why these new .html files are needed now ?

papegaaij: They were needed because the core stylesheet was added, but now they are not really needed anymore. These tests use the same markup as input and expected result. Is this something we would like to keep? The problem I see with that is that regenerating the expected result actually changes the input of the test. That's why I kept these files.

martin-g: I see. Keep them!

wicket-core/src/main/java/org/apache/wicket/settings/ResourceSettings.java
papegaaij: I deliberately chose not to use `ofNullable`, because disabling the core stylesheet has to be a conscious decision.

wicket-core/src/main/java/org/apache/wicket/protocol/http/WebApplication.java
papegaaij: I usually do not want to use lambda's inside lambda's, that's why I did it like this, but I rewrote it with `ifPresent` and it does look better that way. I'll change the code.

wicket-core/src/main/java/org/apache/wicket/mock/MockApplication.java
```diff
@@ -71,5 +71,7 @@ protected void internalInit()
 
 		// for test cases we usually want stable resource names
 		getResourceSettings().setCachingStrategy(NoOpResourceCachingStrategy.INSTANCE);
+		// the core CSS causes noise in tests and isn't needed
+		getResourceSettings().disableWicketCoreCSS();
```
andruhon: Could you please expand this comment explaining what kind of noise.

andruhon:
The change looks good to me

wicket-core/src/main/java/org/apache/wicket/mock/MockApplication.java
martin-g: The "noise" is the extra `<link>` element in the `<head>` of every `**-expected.html` in the test suite

solomax:
Maybe changes required for `WICKET-6726` can also be added to `wicket-core.css`?

papegaaij:
@solomax that's the idea, but I haven't started on that ticket yet :)

papegaaij:
This PR is merged into master manually

PR #401, opened 2020-01-22
WICKET-6726: remove inline styling and javascript from Form

papegaaij:
This PR removes the inline styling and javascript from Form. The inline JS is moved to an event handler that is rendered in the head. The styling to the new core stylesheet under the `wicket--hidden-fields` class.

Note: previously it was possible to change the name of the class used to hide these fields, but I see no use case for that, so I decided to drop that functionality. This was implemented as part of WICKET-6527, in which a user asked to make these elements more easily styleable. For this it is not needed to change the name of the class and the current implementation suits that need.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Application.properties
+++ b/wicket-core/src/main/java/org/apache/wicket/Application.properties
@@ -85,5 +85,3 @@ OddEvenListItem.CSS.even=even
 AutoLabel.CSS.required=required
 AutoLabel.CSS.invalid=error
 AutoLabel.CSS.disabled=disabled
-
-Form.CSS.hidden-fields=hidden-fields
--- a/wicket-core/src/main/java/org/apache/wicket/Application.properties
+++ b/wicket-core/src/main/java/org/apache/wicket/Application.properties
@@ -85,3 +85,6 @@ OddEvenListItem.CSS.even=even
 AutoLabel.CSS.required=required
 AutoLabel.CSS.invalid=error
 AutoLabel.CSS.disabled=disabled
+
+Component.CSS.hidden=wicket--hidden
+Form.CSS.hidden-fields=wicket--hidden-fields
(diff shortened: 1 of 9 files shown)


svenmeier:
I'm not ok with your removal of the changes from WICKET-6527.

solomax:
I believe customizing current CSS class is the same approach as to define own one.
The possibility to add one more class can be restored thus customized CSS selector will be stronger

papegaaij:
@svenmeier I didn't remove the changes for WICKET-6527, only a very small part of it, namely the customizable classname. In WICKET-6527, the user requests that the styling of the hidden fields is customizable. As the inline styling is now gone and the fields are styled via a classname selector, this is now very easy. In fact, the user can replace the whole core stylesheet. I see no need to keep the classname customizable.

Other customizable classnames serve a different need. Those classes are meant to be matched in user stylesheets. In those cases it is desirable to have them match other classnames in the application. This class had predefined styling, which must not conflict with user styling, hence the `wicket--` prefix. When the user wants to change the styling, he can match the wicket class with `div.wicket--hidden-fields` to make it more specific or replace the entire core stylesheet. Both solutions do not require changing the classname.

I can introduce a second class, as suggested by @solomax, but I would like to see a valid use case first.

svenmeier:
Over the last years I've tried to make Wicket's usage of CSS class names uniform and configurable - see CssUtils#key() - so please don't request a 'valid use case' from me, when you throw out my changes.
And no, I don't like a second CSS class either.

papegaaij:
I totally understand the use case for configurable css classes in cases I see as 'user facing': elements a user will need to style because Wicket will not. This case, I would say, is different. Here, Wicket provides the styling, and we need to make sure it does not clash with user styling. This is why these classes are prefixed with a `wicket--` namespace.

Of course, we can make these classname configurable as well, but here I do not see the need. Changing the classname will break the styling, so a user will need to supply it's own styling. But when you are supplying your own stylesheets anyway, why not use the classnames we picked?

solomax:
I would vote for adding customizable CSS class back
This way CSS can be customized much easily (`important` will not be required)

papegaaij:
I've restored the configurable class names (and did the same thing in component for the placeholder). Personally, I still do not see good use for this, but not much harm in it either. Note that I did change the default classname, to prevent styling clashes. The key stayed the same.

martin-g:
Thank you, @papegaaij !

PR #402, opened 2020-01-24
WICKET-6731: Inline JS in SubmitLink

papegaaij:
This PR moves the inline JS in SubmitLink to an event handler.

@martin-g You mentioned that we might need to call `setOutputMarkupId(true)`, but I haven't encountered any issues so far. As long as `renderHead` is called before `onComponentTag`, the id is rendered in the markup.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/SubmitLink.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/SubmitLink.java
@@ -17,7 +17,10 @@
 package org.apache.wicket.markup.html.form;
 
 import org.apache.wicket.markup.ComponentTag;
+import org.apache.wicket.markup.head.IHeaderResponse;
+import org.apache.wicket.markup.head.OnDomReadyHeaderItem;
 import org.apache.wicket.model.IModel;
+import org.apache.wicket.util.string.AppendingStringBuffer;
 
 /**
  * A link which can be used exactly like a Button to submit a Form. The onclick of the link will use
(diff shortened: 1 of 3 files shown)


andruhon:
Sorry for late comment. What if we add something like these to avoid repeated use of Strings concatenation programming?

```
	public static void addEventListener(final IHeaderResponse headerResponse, Component component, String event, CharSequence javascript)
	{
		addEventListener(headerResponse, component, event, "event", javascript, false);
	}

	public static void addEventListener(final IHeaderResponse headerResponse, Component component, String event, String eventParamName, CharSequence javascript, boolean priority)
	{
		OnDomReadyHeaderItem item = OnDomReadyHeaderItem.forScript(
				String.format(
						"var el = document.getElementById('%s'); if(el){el.addEventListener('%s', function(%s){ %s; });}",
						component.getMarkupId(),
						event,
						eventParamName,
						javascript
				)
		);
		headerResponse.render(priority ? new PriorityHeaderItem(item) : item);
	}
```

solomax:
String concatenation is faster than String.format in latest Java ...

andruhon:
I meant isolating this JavaScript string construction into some kind of
methods, maybe inside of JavaScriptUtils, not actual implementation.

andruhon:
Behaviour maybe

papegaaij:
I agree with @andruhon that these magic string constructions all over the place are not very nice. We will need these at more places. What about:
```
OnDomReadyHeaderItem.forEvent(component, "click", body);
```

papegaaij:
Well, it turned out this already exists and is called `OnEventHeaderItem` and according to the javadoc, I'm the author :) However, it does have a horrible API where `target` is a literal and needs to be quoted by the caller, so I gave it a small face lift.

PR #403, opened 2020-02-10
Iteration stops with nested containers

helmut8080:
Since there is only one pop() performed in hasNext() the iteration stops, if there are several nested container, when there are no additional childs in the parent container.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/MarkupContainer.java
+++ b/wicket-core/src/main/java/org/apache/wicket/MarkupContainer.java
@@ -2200,7 +2200,7 @@ public abstract class MarkupContainer extends Component implements Iterable<Comp
 			@Override
 			public boolean hasNext()
 			{
-				if (!currentIterator.hasNext() && !iteratorStack.isEmpty())
+				while (!currentIterator.hasNext() && !iteratorStack.isEmpty())
 				{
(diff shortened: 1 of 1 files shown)


martin-g:
Thank you for the Pull request, @helmut8080 !

Can you add a unit test demonstrating the problem ? It will help preventing a regression in the future.
A quickstart application is also good. We will convert it to a test case.

helmut8080:
Sure. Hope its ok to just post the unit test here:

@Test
	void streamChildrenNestedContainer() {
		WebMarkupContainer wmc = new WebMarkupContainer("parent");
		WebMarkupContainer wmc1 = new WebMarkupContainer("wmc1");
		wmc.add(wmc1);
		WebMarkupContainer wmc1_2= new WebMarkupContainer("wmc1_2");
		wmc1.add(wmc1_2);
		Label lbl2 = new Label("lbl2");
		wmc.add(lbl2);
		List<String> l = wmc.streamChildren().map(t -> t.getId()).collect(Collectors.toList());
//		System.out.println(l);
		assertEquals(l.toString(), "[wmc1, wmc1_2, lbl2]");
	}

martin-g:
https://issues.apache.org/jira/browse/WICKET-6754

martin-g:
Thank you, @helmut8080 !

PR #404, opened 2020-02-10
[WICKET-6748] CSS onload/onerror is rewritten

solomax:
tested using latest Chrome/Chromium/FF/IE11


solomax:
@svenmeier could you please take a look?

PR #407, opened 2020-02-22
[NO_ISSUE] roll an abstract class that contains common functionality …

reiern70:
Roll an abstract class that contains common functionality for AJAX and Web-socket request handlers


reiern70:
@martin-g  thanks for approval. Test were green locally.

PR #408, opened 2020-02-23
allow to abort currently executing request for an Ajax channel

reiern70:
Hi @martin-g ,

We have the following use case: 

1- Very big AJAX upload... user gets tired of waiting
2- User wants to cancel upload

Thus we would like to cancel at client side the current AJAX request for the channel used for the upload.  I had a look a channels with DROP type... and this does not seem to be useful in above use case... as scheduling a new request will wait fro current request to finish before erasing all other waiting requests (or so I understood).

Disclaimer: I haven't tested the code included on PR. I just want to know if in your opinion if this a good approach and it could work without side effects... Or if there is a better approach?

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -278,6 +278,15 @@
 			}
 		},
 
+		// aborts current request is there is any running
+		abort: function () {
+			if (isUndef(this.jqXHR)) {
+				Wicket.Log.debug("There is no executing request fro channel " + this.name)
+			} else {
+				this.jqXHR.abort();
+			}
+		},
+
 		done: function () {
 			var callback = null;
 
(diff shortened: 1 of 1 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -278,6 +278,15 @@
 			}
 		},
 
+		// aborts current request is there is any running
+		abort: function () {
+			if (isUndef(this.jqXHR)) {
+				Wicket.Log.debug("There is no executing request fro channel " + this.name)
```
martin-g: s/fro/for/

martin-g: missing semicolon at the end

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -319,6 +328,26 @@
 			return c.schedule(callback);
 		},
 
+		// register AJAX request
+		registerCurrentAjaxRequest: function(name, jqXHR) {
+			var c = this.channels[name];
+			if (isUndef(c)) {
+				Wicket.Log.error("No channel with name " + name)
```
martin-g: missing semicolon at the end

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -319,6 +328,26 @@
...
+			} else {
+				c.jqXHR = jqXHR;
+			}
+		},
+
+		// aborts the current request for channel with name name
+		abortCurrentQuestForChannel(name) {
```
martin-g: s/Quest/Request/ :-)

martin-g:
And please create a ticket. This is an improvement that should appear in the changelog.

reiern70:
> Hi @reiern70 !
> 
> Few minor comments to the changes.
> IMO it should do the job!
> 
> Please add some tests to https://github.com/apache/wicket/blob/master/wicket-core/src/test/js/ajax.js

Hi @martin-g,

Thanks for review and comments!  Now that I have confirmation that idea might work I will work to move it  production ready state.

reiern70:
See...

https://issues.apache.org/jira/browse/WICKET-6750

Closing this PR + delete remote branch + I will create a new PR when new feature is ready for review.

Thanks again!

PR #409, opened 2020-02-25
[WICKET-6750] allow to abort currently executing AJAX request 

reiern70:
[WICKET-6750] allow to abort currently executing request for a given Ajax channel + example

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -583,11 +583,20 @@
 		},
 
 		/**
-		 * Aborts current AJAX request, if any is running, for default channel.
-		 * WARNING! Mind that this does not implies and server immediately will know about
-		 * request being aborted and server side processing (and page lock night continue).
-		 * This method might prove useful, for instance, in order to abort big AJAX
-		 * uploads.
+		 * Aborts current AJAX request, if any is running, for channel @channel.
+		 *
+		 * WARNING! Mind that this does not implies the server will immediately know about
+		 * request being aborted. And there is no reliable way to tell server to stop process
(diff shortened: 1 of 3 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -547,13 +576,29 @@
...
-
 			var res = Wicket.channelManager.schedule(attrs.ch, Wicket.bind(function () {
 				this.doAjax(attrs);
 			}, this));
 			return res !== null ? res: true;
 		},
 
+		/**
+		 * Aborts current AJAX request, if any is running, for default channel.
+		 * WARNING! Mind that this does not implies and server immediately will know about
```
andruhon: This does not imply that the server will immediately know about that? Is it what you're trying to tell?

I think this comment needs rewording.

andruhon: I'd reword it somewhat like this:
```
Abort AJAX request if one in default channel is currently running.

WARNING! Mind that calling this method does not imply that the server will immediately know about the request being aborted.
Server side processing and page are likely to continue.
This method could be useful to cancel huge AJAX requests such as file uploads.
```

reiern70: > This does not imply that the server will immediately know about that? Is it what you're trying to tell?
> 
> I think this comment needs rewording.

What I mean is "there is not reliable way to tell the server stop what I asked you to do". In case of upload stream will be broken and upload halted with a server side exception.

reiern70: @andruhon thanks for your comment.

image

I have reworded it like above. Hope is is more clear now.

PR #410, opened 2020-02-26
[WICKET-6750] allow to abort currently executing request for a given …

reiern70:
allow to abort currently executing request for a given Ajax channel + example. It uses suggestion by @svenmeier. If these changes are not dimmed necessary or desirable I can create some sample project somewhere illustrating how to do this.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -336,6 +336,54 @@
 	Wicket.ChannelManager.FunctionsExecuter = FunctionsExecuter;
 
 	/**
+	 * AjaxRequestMonitor allows to get hold of jqXHR during AJAX request and
+	 * manipulate it.
+	 */
+	Wicket.AjaxRequestMonitor = Wicket.Class.create();
+
+	Wicket.AjaxRequestMonitor.prototype = {
+		initialize: function () {
+			this.requests = [];
+		},
+
+		init: function() {
+			var requests = this.requests;
+			Wicket.Event.subscribe('/ajax/call/beforeSend',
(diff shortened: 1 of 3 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -335,6 +335,58 @@
 
 	Wicket.ChannelManager.FunctionsExecuter = FunctionsExecuter;
 
+	/**
+	 * Channel manager maintains a map of channels.
```
reiern70: Change this.

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
```diff
@@ -335,6 +335,58 @@
...
+			);
+
+			Wicket.Event.subscribe('/ajax/call/complete',
+				function(jqEvent, attributes, jqXHR, errorThrown, textStatus) {
+					delete requests[attributes.ch];
+				}
+			);
+		},
+
+		getReuest: function (channel) {
```
reiern70: and this

svenmeier:
How about extracting that code into a behavior, e.g. something which can be added to a button?
This would be much nicer to use (showing/hiding the button, keeping track of the currrent Ajax request, cancelling it).
With its own JS file it we could keep this stuff out of the core JS, maybe add this feature to wicket-extensions.

reiern70:
> How about extracting that code into a behavior, e.g. something which can be added to a button?
> This would be much nicer to use (showing/hiding the button, keeping track of the currrent Ajax request, cancelling it).
> With its own JS file we could keep this stuff out of the core JS, maybe add this feature to wicket-extensions.

@svenmeier Many thanks for feedback. This sounds like a very good idea. I will do that and create a new PR.

PR #411, opened 2020-02-28
WICKET-6751 Page synchronizer strategies

theigl:
This PR adds support for providing a custom lock manager for page access synchronization. 

Current synchronization assumes that all requests of a session work on an identical session object, where it keeps all current locks. This isn't the case for Spring managed sessions, where the session is (de)serialized to external storage on every request.

For a detailed discussion of the subject see the [mailing list](http://mail-archives.apache.org/mod_mbox/wicket-users/202002.mbox/%3cCANtfzo5+YTBFzyP-U_dVjSrRu8xgv1qj4ozAWxDAbwVnF5=M9Q@mail.gmail.com%3e).

The following changes were made:
- Expose the currently package scoped methods for locking and waiting on a lock of `PageLock` so the class can be re-used in custom implementations
- Encapsulate all logic related to locking to an `IPageLockManager` and move the current locking implementation to `DefaultPageLockManager`

Open issues:

- All changes are backwards-compatible from an API perspective. The only method that was moved is the `getTimeout(int pageId)` method on `PageAccessSynchronizer` that was public for some reason. I could add the method to the `IPageLockManager` interface and restore it as a delegate in `PageAccessSynchronizer` if wanted.

- I'm not sure if `PageLock` is in the right place now. It is still a static inner class of `PageAccessSynchronizer` but it is only used by `DefaultLockManager` and can be used by custom lock manager implementations. It would probably be better to make it a top-level class.

- `lockPage`, `unlockPage` and `unlockAllPages` in `PageAccessSynchronizer` now simply delegate to the `IPageLockManager`. These methods could be made final or removed entirely if we do not care about backwards-compatibility.

See https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6751


svenmeier:
If PageAccessSynchronizer delegates everything to a IPageLockManager anyway, wouldn't it be simpler to introduce IPageAccessSynchronizer instead? Why add another redirection?

theigl:
@svenmeier: This would be possible. But the only method of this interface would be: 

`IPageManager adapt(IPageManager pagemanager)`

This is the only method every called by Wicket. All other methods related to locking are internal. `PageAccessSynchronizer` still serves its purpose because it contains the logic for decorating the page manager and coordinating the locking and unlocking via the new `IPageLockManager`.

theigl:
@svenmeier: Do you need some additional information for reviewing this PR? I'm happy to provide it and work on the PR some more if changes are requested.

wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
```diff
@@ -0,0 +1,201 @@
...
+import org.apache.wicket.util.LazyInitializer;
+import org.apache.wicket.util.lang.Threads;
+import org.apache.wicket.util.time.Durations;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default {@link IPageLockManager} that that holds a map of locks in the current session.
+ * 
+ * @author Igor Vaynberg (ivaynberg)
```
martin-g: Please remove this line. It is not really true anymore.

wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
```diff
@@ -0,0 +1,201 @@
...
+			{
+				// wait for a lock to become available
+				long remaining = remaining(start, timeout);
+				if (remaining > 0)
+				{
+					previous.waitForRelease(remaining, isDebugEnabled);
+				}
+			}
+		}
+		if (locked)
```
martin-g: please add an empty line before line 112

theigl:
@martin-g: Thanks for reviewing! I'll push the requested changes asap.

What do you think about this issue from my PR description:

> I'm not sure if PageLock is in the right place now. It is still a static inner class of PageAccessSynchronizer but it is only used by DefaultLockManager and can be used by custom lock manager implementations. It would probably be better to make it a top-level class.

Should we keep it in `PageAccessSynchronizer` even though it is not used by the class anymore? Instead of making it a top-level class, I could also move it to `DefaultLockManager`. This probably makes the most sense.

wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
theigl: Done

theigl:
@svenmeier : Hi. Since Martin already approved this, could you please take another look? I'd really like to see this merged at some point.

theigl:
@svenmeier: Thanks for your approval. I know this does not have high priority, I just wanted to make sure that this PR is in a mergeable state. 

As you rightly pointed out, it is already possible to provide my own synchronizer and I have successfully done so, but it forced me to copy a lot of quite complex code to my project (among others the whole `PageLock` class).  Since Spring Session is quite a well known project and others might run into this issue as well, I'd still vote for merging this. It will save the next person who runs into this a lot of work.

PR #413, opened 2020-03-15
WICKET-6714: Fixed regresssion about converting URLs to paths.

ams-tschoening:
During [WICKET-6714](https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6714) a [regression](https://github.com/apache/wicket/commit/4223b8d075cd6bfa4362914c19dbca5de79e4f69) has been introduced regarding handling paths with spaces. `URL.getFile()` [doesn't decode URLs with spaces correctly](https://stackoverflow.com/questions/2166039/java-how-to-get-a-file-from-an-escaped-url), resulting in wrongly generated paths. While the slashes in the following example are not that much of a problem, "%20" vs. " " is, because only the latter exists in the file system.

> /C:/Users/tschoening/Documents/Eclipse/Java%20Sm-Mtg%20bug_2590/de.am_soft.sm_mtg.view.report/build/resources/test/in/_other/./

vs.

> C:\\Users\\tschoening\\Documents\\Eclipse\\Java Sm-Mtg bug_2590\\de.am_soft.sm_mtg.view.report\\build\\resources\\test\\in\\_other\\.

The change additionally makes the returned path more in-line with the documented behaviour of that method:

> The real path returned will be in a form appropriate to the computer and operating system on which the servlet container is running, including the proper path separators.

I've searched the code base for similar problems and added commits to fix those for your review as well. Feel free to suggest splitting those commits if you think that's necessary. This PR is against `wicket-8.x` because MASTER doesn't work currently on my Windows.


svenmeier:
Many thanks!

Please create a separate issue and pull request for the other improvements.

ams-tschoening:
The fix for the regression has been merged already, the other two commits are separated into individual PRs #414 and #415 for further discussion.

PR #414, opened 2020-03-16
WICKET-6756: Avoid "URL.getFile()" when expecting file system paths.

ams-tschoening:
Like has been requested in PR #413, another commit removing `URL.getFile` when actually a path is expected. `getFile` [doesn't properly decode](https://stackoverflow.com/questions/2166039/java-how-to-get-a-file-from-an-escaped-url) paths. In theory this needs to be fixed everywhere, but I found only this place for now, someone else might want to have an additional look at some point.

https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6756


ams-tschoening:
This seems to have been merged:

https://github.com/apache/wicket/commit/1db8b1c2778b435b51a84e1354dbb56475b23855

PR #415, opened 2020-03-16
WICKET-6757: Avoid URL.getFile during mime type detection.

ams-tschoening:
Mime type detection is documented to work on paths or URLs, but URL.getFile provides neither of both and the default implementation is very limited regarding / vs. \ and such things. Forwarding an absolute URI always instead is compatible with the default implementation and makes absolutely clear what is available. URL.getFile for example might look like some path in the file system, but would work with paths containing spaces because of no decoding.

https://issues.apache.org/jira/browse/WICKET-6757
Related to PR #413.


ams-tschoening:
This seems to have been merged:

https://github.com/apache/wicket/commit/f1bd38690674be5578fc5c5e1c28cc81cdb7ae76

PR #416, opened 2020-03-20
WICKET-6759 Support disabling error notification for websockets

theigl:
This PR adds a function `shouldNotifyOnErrorEvent` to `WebSocketSettings` that allows the user the decide which types of websocket errors should be broadcast. It serves the same purpose as the existing `shouldNotifyOnCloseEvent` function, i.e. to avoid deserializing the relevant page instance.

See https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6759


martin-g:
Thank you, @theigl !

PR #417, opened 2020-03-20
WICKET-6761 Multiple websocket connections from the same session

theigl:
This PR adds an optional `resourceToken` to `BaseWebSocketBehavior` to support multiple websocket connections from the same session, i.e. from multiple browser tabs.

The token could be a UUID or some other calculated value.

Note: I'm not really happy about the name `resourceToken`. Maybe something like `connectionId` would be easier to understand.

See https://issues.apache.org/jira/browse/WICKET-6761

### DIFF
--- a/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/BaseWebSocketBehavior.java
+++ b/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/BaseWebSocketBehavior.java
@@ -40,7 +40,7 @@ import org.apache.wicket.util.template.PackageTextTemplate;
 public class BaseWebSocketBehavior extends Behavior
 {
 	private final String resourceName;
-	private final String resourceToken;
+	private final String connectionToken;
 
 	/**
 	 * Constructor.
@@ -51,7 +51,7 @@ public class BaseWebSocketBehavior extends Behavior
 	protected BaseWebSocketBehavior()
 	{
 		this.resourceName = null;
-		this.resourceToken = null;
+		this.connectionToken = null;
 	}
 
 	/**
@@ -90,13 +90,13 @@ public class BaseWebSocketBehavior extends Behavior
 	 *
 	 *  @param resourceName
 	 *          the name of the shared {@link org.apache.wicket.protocol.ws.api.WebSocketResource}
-	 *  @param resourceToken
+	 *  @param connectionToken
 	 *  		an optional token to support connections to the same resource from multiple browser tabs
 	 */
-	public BaseWebSocketBehavior(String resourceName, String resourceToken)
+	public BaseWebSocketBehavior(String resourceName, String connectionToken)
 	{
 		this.resourceName = Args.notEmpty(resourceName, "resourceName");
-		this.resourceToken = resourceToken;
+		this.connectionToken = connectionToken;
 	}
 
 	@Override
@@ -119,12 +119,12 @@ public class BaseWebSocketBehavior extends Behavior
 			int pageId = component.getPage().getPageId();
 			variables.put("pageId", pageId);
 			variables.put("resourceName", "");
-			variables.put("resourceToken", "");
+			variables.put("connectionToken", "");
 		}
 		else
 		{
 			variables.put("resourceName", resourceName);
-			variables.put("resourceToken", resourceToken);
+			variables.put("connectionToken", connectionToken);
 			variables.put("pageId", false);
 		}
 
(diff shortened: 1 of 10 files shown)


wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/res/js/wicket-websocket-jquery.js
```diff
@@ -76,6 +76,9 @@
 					url += '?pageId=' + encodeURIComponent(WWS.pageId);
 				} else if (WWS.resourceName) {
 					url += '?resourceName=' + encodeURIComponent(WWS.resourceName);
+					if (WWS.resourceToken) {
+						url += '&resourceToken=' + encodeURIComponent(WWS.resourceName);
```
martin-g: encodeURIComponent(WWS.resource**Token**);

wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/BaseWebSocketBehavior.java
```diff
@@ -93,10 +119,12 @@ public void renderHead(Component component, IHeaderResponse response)
 			int pageId = component.getPage().getPageId();
 			variables.put("pageId", pageId);
 			variables.put("resourceName", "");
+			variables.put("resourceToken", "");
```
martin-g: Why the token is supported only for Resource but not for Page ?
In case of a Page there will be locking on the page instance and it will work sequential but it should still work, I think.

wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/res/js/wicket-websocket-jquery.js
theigl: Fixed.

wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/api/BaseWebSocketBehavior.java
theigl: @martin-g: My reasoning was this: Pages already work with multiple tabs because the connection key contains the `pageId` with is unique it to each tab as far as I understand. We could pass it for pages as well and add it to the key but it wouldn't really improve things.

martin-g: > contains the `pageId` with is unique it to each tab

This is not correct! 
If the user copies the url from the browser's address bar and pastes it in a new tab then the page id is the same.
This is why we have https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/ajax/AjaxNewWindowNotifyingBehavior.java

theigl: Ah ok. I have been using wicket with "pretty" URLs without any `pageId` parameters for so long that I forgot how it works out of the box.

If we want to use it for pages as well, we definitely need a better name for the parameter. What do you think about `connectionId` or `connectionToken`?

martin-g: `connectionToken` sounds better.

But who is creating this token ? I didn't see anything about this in your PR.
Have you tested your PR? Does it really solve your issue ? Because due to the typo in the Javascript code I have the feeling you didn't actually used this code.
Please add changes to the Wicket Examples demo application that makes it easier for me to play with this new functionality. 
Thanks!

theigl: Martin: You are right! I did not test this PR yet, it was meant as a basis for discussion and to clarify what I'm trying to achieve. 

I now added a demo page to the websocket examples at: http://localhost:8080/websockets

It uses `UUID.randomUUID()` as a token. If you open two tabs of the new page, you will see that the charts are updated correctly. If you open two tabs of the existing demo page, the first tab will stop receiving updates because the connection has been overwritten.

I had to make two changes to the existing demo to showcase this:
- Increase the pool size of the scheduled executor to support multiple tabs
- Move connection lookup in `ChartUpdater` into the loop so the connection is not cached but looked up every time the chart wants to push new data. Even though the connection was overwritten in the registry, the demo still worked before because `ChartUpdater` held a reference to the connection.

theigl: I can add the `connectionToken` to pages as well if we want to go ahead with this approach. For now, I kept it just for resources.

martin-g:
Thank you, @theigl !

PR #418, opened 2020-03-26
WICKET-6762 Allow users to customize websocket setup

theigl:
This PR allows users to customize the setup of websocket connections.

Websocket connections are currently opened when the DOM is ready. If users open tabs in the background, a new websocket connection is initialized for each tab. I want to defer connection creation to the moment when the user activates the tab.

Instead of adding yet another configuration parameter, I thought it might be better to give users full control of the initialization script by exposing `getWebSocketSetupScript`.

See https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6762


martin-g:
Thank you, @theigl !

PR #423, opened 2020-04-29
WICKET-6774 Separate component data into model, behaviors and metadata

theigl:
This PR extracts the component's model and metadata from the `Component.data` object.

The changes make manipulating metadata and models much simpler and improve performance of retrieving metadata (see https://issues.apache.org/jira/browse/WICKET-6771).

The tradeoff is that metadata is now always stored as an array. Previously, a single metadata entry would have been stored as a simple object. That means that components that have exactly one metadata entry now have a slightly higher memory footprint (~20 bytes).

On the flip side, extracting the model from the data object can potentially save some memory. When a component has a model and exactly one behavior, both are now stored as objects. Previously, the model plus the behavior would have been stored in an array. So we save about 20 bytes in this case.

Note: I kept the `data` object as-is. It still behaves the same way, i.e. it stores an object for a single behavior and an array for multiple behaviors. I do not see room for improvement without sacrificing additional memory. Also, typing it as `Behavior[]` would not be possible because it also stores `BehaviorIdLists`.

https://issues.apache.org/jira/browse/WICKET-6774

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Behaviors.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Behaviors.java
@@ -73,17 +73,11 @@ final class Behaviors implements IDetachable
 	public <M extends Behavior> List<M> getBehaviors(Class<M> type)
 	{
 		final int len = component.data_length();
-		final int start = component.data_start();
-		if (len < start)
-		{
-			return Collections.emptyList();
-		}
-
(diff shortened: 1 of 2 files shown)


theigl:
Closing because of memory implications as discussed in https://issues.apache.org/jira/browse/WICKET-6774.

PR #432, opened 2020-05-06
WICKET-6774: minor change to contract of Component.getBehaviorId

papegaaij:
This PR suggest a minor change to the contract of IRequestComponent.getBehaviorId to prepare for a refactoring of the component internals.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -4448,6 +4448,11 @@ public abstract class Component
 	@Override
 	public final int getBehaviorId(Behavior behavior)
 	{
+		if (behavior.isTemporary(this))
+		{
+			throw new IllegalArgumentException(
+				"Cannot get a stable id for temporary behavior " + behavior);
+		}
 		return new Behaviors(this).getBehaviorId(behavior);
 	}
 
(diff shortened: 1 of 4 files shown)


papegaaij:
> Please do not merge this PR until we are clear on the way this is supposed to work.
> https://markmail.org/message/p4cnvfypzprrl2qt

Agreed, that's why I created a PR for such a simple change. The change is simple, but the implications of doing it wrong are much greater. We need have a clear understanding of how we want these stable id's to work.

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
```diff
@@ -83,12 +83,12 @@ protected void onBind()
...
 		component.setOutputMarkupId(true);
-		
-		if (getStatelessHint(component))
-		{
-			//generate behavior id
-			component.getBehaviorId(this);
-		}
+	}
+	
+	@Override
```
bitstorm, line 88: Hi,
forcing Ajax behavior to be stateful by default will break the current logic for stateless Ajax component. See [user guide](https://ci.apache.org/projects/wicket/guide/8.x/single.html#_stateless_ajax_components_behaviors). 
I would remove this isStableIdRequired override and just keep things as they are now, i.e. behaviors state (stateful/stateless) depends on container component.

papegaaij: `isStableIdRequired` does not make the behavior stateful, that's the purpose of this new method.

bitstorm: ok. thanks for the explanation!

PR #434, opened 2020-05-25
[WICKET-6789] 'base-uri' is added to CSPDirective enum

solomax:


wicket-core/src/main/java/org/apache/wicket/csp/CSPHeaderConfiguration.java
```diff
@@ -99,7 +100,8 @@ public CSPHeaderConfiguration unsafeInline()
 			.add(CONNECT_SRC, SELF)
 			.add(FONT_SRC, SELF)
 			.add(MANIFEST_SRC, SELF)
-			.add(CHILD_SRC, SELF);
+			.add(CHILD_SRC, SELF)
+			.add(BASE_URI, SELF);
```
papegaaij, line 104: The new directive should be added to the javadoc of this method as well.

solomax: Should be addressed :)
Thanks for review!

PR #435, opened 2020-05-28
WICKET-6791 Offload WebSocket push when initiated in Wicket request cycle

martin-g:
Use different thread executors when pushing WebSocketPushMessage in http worker thread and non-worker thread

### DIFF
--- a/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/WebSocketSettings.java
+++ b/wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/WebSocketSettings.java
@@ -39,6 +39,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.servlet.http.HttpServletRequest;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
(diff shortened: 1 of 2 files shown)


wicket-native-websocket/wicket-native-websocket-core/src/main/java/org/apache/wicket/protocol/ws/WebSocketSettings.java
```diff
@@ -388,9 +401,65 @@ public Integer getSecurePort()
...
+		 */
+		private final java.util.concurrent.Executor httpRequestExecutor;
+
+		/**
+		 * For non-http worker threads pushes the WebSocket runnable in the same request.
+		 * For http worker threads uses an elastic thread pool of 1-8 threads.
+		 */
+		public WebSocketPushMessageExecutor()
+		{
+			this(Runnable::run, new ThreadPoolExecutor(1, 8,
```
solomax, line 429: shall this be configurable?

martin-g: It is configurable by directly using the called constructor.

solomax: my bad :(

PR #436, opened 2020-05-29
WICKET-6794 Performance improvements for UrlEncoder and UrlDecoder

theigl:
This PR rewrites `UrlEncoder` and `UrlDecoder` based on Spring's `UriUtils` instead of `java.net.URLEncoder` and `java.net.URLDecoder`.

The new implementation is 30-500% faster depending on the use case. The biggest improvement is with encoding URLs that do not contain any character that requires escaping.

All tests pass and encoding should be fully backwards compatible.

For benchmark results, see the JIRA ticket.

See https://issues.apache.org/jira/browse/WICKET-6794

### DIFF
--- a/wicket-util/src/main/java/org/apache/wicket/util/encoding/UrlDecoder.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/encoding/UrlDecoder.java
@@ -16,25 +16,25 @@
  */
 package org.apache.wicket.util.encoding;
 
+import java.io.ByteArrayOutputStream;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.nio.charset.IllegalCharsetNameException;
 import java.nio.charset.UnsupportedCharsetException;
 
 import org.apache.wicket.util.lang.Args;
-import org.apache.wicket.util.string.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Adapted from java.net.URLDecoder, but defines instances for query string decoding versus URL path
+ * Adapted from Spring's UriUtils, but defines instances for query string decoding versus URL path
  * component decoding.
  * <p/>
  * The difference is important because a space is encoded as a + in a query string, but this is a
  * valid value in a path component (and is therefore not decode back to a space).
- * 
- * @author Doug Donohoe
- * @see java.net.URLDecoder
+ *
+ * @author Thomas Heigl
+ * @see org.springframework.web.util.UriUtils
  * @see <a href="http://www.ietf.org/rfc/rfc2396.txt">RFC-2396</a>
  */
 public class UrlDecoder
@@ -46,7 +46,7 @@ public class UrlDecoder
 	/**
 	 * Encoder used to decode name or value components of a query string.<br/>
 	 * <br/>
-	 * 
+	 *
 	 * For example: http://org.acme/notthis/northis/oreventhis?buthis=isokay&amp;asis=thispart
 	 */
 	public static final UrlDecoder QUERY_INSTANCE = new UrlDecoder(true);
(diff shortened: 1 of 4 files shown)


wicket-util/src/main/java/org/apache/wicket/util/encoding/UrlDecoder.java
```diff
@@ -16,25 +16,25 @@
...
  * component decoding.
  * <p/>
  * The difference is important because a space is encoded as a + in a query string, but this is a
  * valid value in a path component (and is therefore not decode back to a space).
- * 
- * @author Doug Donohoe
- * @see java.net.URLDecoder
+ *
+ * @author Thomas Heigl
+ * @see org.springframework.web.util.UriUtils
```
martin-g, line 37: I think this breaks the Javadoc build because wicket-util does not depend on spring-web.
Please the first line of the javadoc with something like "Adapted from Spring Framework's UrilUtils class"

wicket-util/src/main/java/org/apache/wicket/util/encoding/UrlEncoder.java
```diff
@@ -16,58 +16,140 @@
...
 import java.nio.charset.Charset;
 import java.nio.charset.IllegalCharsetNameException;
 import java.nio.charset.UnsupportedCharsetException;
-import java.util.BitSet;
 
 import org.apache.wicket.util.lang.Args;
 
 /**
- * Adapted from java.net.URLEncoder, but defines instances for query string encoding versus URL path
+ * Adapted from Spring's UriUtils, but defines instances for query string encoding versus URL path
```
martin-g, line 28: Same as above

wicket-util/src/main/java/org/apache/wicket/util/encoding/UrlDecoder.java
theigl: Done

theigl:
Can somebody else please take a look at this as well? This change is fully covered by tests, but I'd really like one more approval before merging this.

theigl:
Thanks @svenmeier and @martin-g!

wicket-util/src/main/java/org/apache/wicket/util/encoding/UrlDecoder.java
```diff
@@ -16,25 +16,24 @@
...
+ * Adapted from Spring Framework's UriUtils class, but defines instances for query string decoding versus URL path
  * component decoding.
  * <p/>
  * The difference is important because a space is encoded as a + in a query string, but this is a
  * valid value in a path component (and is therefore not decode back to a space).
- * 
- * @author Doug Donohoe
- * @see java.net.URLDecoder
+ *
+ * @author Thomas Heigl
```
dougdonohoe, line 36: I happened to be looking at Wicket source code again and noticed this change.  I'm the original author of the UrlDecode, originating back in 2008 in https://issues.apache.org/jira/browse/WICKET-1627.

This @author change is not cool.  It should at least be:

```
@author Doug Donohoe
@author Thomas Heigl
```

CC: @theigl

wicket-util/src/main/java/org/apache/wicket/util/encoding/UrlEncoder.java
```diff
@@ -16,58 +16,139 @@
...
+ * Adapted from Spring Framework's UriUtils class, but defines instances for query string encoding versus URL path
  * component encoding.
  * <p/>
  * The difference is important because a space is encoded as a + in a query string, but this is a
  * valid value in a path component (and is therefore not decode back to a space).
- * 
- * @author Doug Donohoe
- * @see java.net.URLEncoder
+ *
+ * @author Thomas Heigl
```
dougdonohoe, line 34: I happened to be looking at Wicket source code again and noticed this change.  I'm the original author of the UrlDecode, originating back in 2008 in https://issues.apache.org/jira/browse/WICKET-1627.

This @author change is not cool.  It should at least be:

```
@author Doug Donohoe
@author Thomas Heigl
```

CC: @theigl

PR #440, opened 2020-08-06
WICKET-6807 Improve the check for the submitting component

martin-g:
First check for the special 'name.x' parameter
If there is no such then check that there is exactly one 'name' parameter
Always use only the GET/POST parameters as the form's method


wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
```diff
@@ -1711,7 +1725,7 @@ public final void writeHiddenFields()
 	{
 		// if it's a get, did put the parameters in the action attribute,
 		// and have to write the url parameters as hidden fields
-		if (encodeUrlInHiddenFields())
+		if (METHOD_POST.equalsIgnoreCase(getMethod()))
```
martin-g, line 1728: Sven Meier comment from https://issues.apache.org/jira/browse/WICKET-6807?focusedCommentId=17172167&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17172167:

That is not correct!

Parameters **must** be written as hidden fields, when the Form is 'get' (because the browser would strip them from the form's action). For a 'post' Form the parameters **can** be either in hidden fields **or** in the form's action, both would work.

martin-g: In that case let's always write them as hidden fields.

martin-g: I've updated it.
I want to check what else could be simplified by using always hidden fields. I don't see a reason to write extra query parameters in the form action url instead of using hidden fields everywhere.

svenmeier:
Always writing parameters into hidden fields will mix them into the post parameters. This makes it impossible to differentiate them with form fields - something WICKET-6807 actually wanted to improve on.

PR #449, opened 2020-08-30
WICKET-6824 Replace slow `String.format` with concatenation

theigl:
This PR replaces two frequently called usages of `String.format` with concatenation because it is orders of magnitudes faster.

See https://issues.apache.org/jira/browse/WICKET-6824

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
@@ -61,16 +61,16 @@ public abstract class AbstractDefaultAjaxBehavior extends AbstractAjaxBehavior
 	public static final ResourceReference INDICATOR = new PackageResourceReference(
 		AbstractDefaultAjaxBehavior.class, "indicator.gif");
 
-	private static final String DYNAMIC_PARAMETER_FUNCTION_TEMPLATE = "function(attrs){%s}";
-	private static final String PRECONDITION_FUNCTION_TEMPLATE = "function(attrs){%s}";
-	private static final String COMPLETE_HANDLER_FUNCTION_TEMPLATE = "function(attrs, jqXHR, textStatus){%s}";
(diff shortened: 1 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
```diff
@@ -61,16 +61,16 @@
...
-	private static final String PRECONDITION_FUNCTION_TEMPLATE = "function(attrs){%s}";
-	private static final String COMPLETE_HANDLER_FUNCTION_TEMPLATE = "function(attrs, jqXHR, textStatus){%s}";
-	private static final String FAILURE_HANDLER_FUNCTION_TEMPLATE = "function(attrs, jqXHR, errorMessage, textStatus){%s}";
-	private static final String SUCCESS_HANDLER_FUNCTION_TEMPLATE = "function(attrs, jqXHR, data, textStatus){%s}";
-	private static final String AFTER_HANDLER_FUNCTION_TEMPLATE = "function(attrs){%s}";
-	private static final String BEFORE_SEND_HANDLER_FUNCTION_TEMPLATE = "function(attrs, jqXHR, settings){%s}";
-	private static final String BEFORE_HANDLER_FUNCTION_TEMPLATE = "function(attrs){%s}";
-	private static final String INIT_HANDLER_FUNCTION_TEMPLATE = "function(attrs){%s}";
-	private static final String DONE_HANDLER_FUNCTION_TEMPLATE = "function(attrs){%s}";
+	private static final String DYNAMIC_PARAMETER_FUNCTION_TEMPLATE = "function(attrs)";
```
solomax, line 64: Maybe there is no need in all these constants with same value?

svenmeier: IMHO these constants serve as a good documentation for the functions' actual parameters.

And it will be hard to come up with good names of all used parameter combinations anyway :)

solomax: make sense :)

PR #452, opened 2020-09-13
WICKET-6830 Convert `Behaviors` into a static utility

theigl:
We currently create a `Behaviors` instance for every invocation of its methods. While this is slightly "more OO" than a static utility class, it allocates hundreds of MiB per minute in my application.

This PR converts `Behaviors` into a static utility class.

See https://issues.apache.org/jira/browse/WICKET-6830

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Behaviors.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Behaviors.java
@@ -22,25 +22,22 @@ import java.util.List;
 
 import org.apache.wicket.behavior.Behavior;
 import org.apache.wicket.behavior.InvalidBehaviorIdException;
-import org.apache.wicket.model.IDetachable;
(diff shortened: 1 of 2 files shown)


solomax (approved):

svenmeier (approved):

martin-g (commented):
I guess the build fails for you after this change.
clirr-maven-plugin should complain that this is not backward compatible API change.

IMO we should introduce `BehaviorsUtil/Behaviors2/...` class and use it internally in Wicket.
The current `Behaviors` class should be deprecated and removed in the next major version.

theigl:
@martin-g: The build does not fail for me. `Behaviors` is a package private, final class. This should not affect the public API at all, or am I missing something?

svenmeier:
I thought so too.

martin-g:
Ah, OK!
Then it should be good!

PR #453, opened 2020-09-22
Make FilterForm#onComponentTagBody generated Markup overridable

renoth:
We have the problem that by using a subclass of Filterform that is Ajax enabled we have to generate a different markup in onComponentTagBody. This PR would enable anyone to subclass Filterform and override the generated markup to something different.


martin-g (approved):

theigl (approved):

svenmeier (approved):

solomax:
Thanks for the contribution @renoth !

PR #454, opened 2020-09-25
WICKET-6833 Reduce allocations for PageParameters.mergeWith

theigl:
This PR reduces allocations in `PageParameters.mergeWith`. As a side-effect, it increases throughput 4x.

It avoids allocating new collections where possible and attempts to initialize `namedParameters` with the correct size.

This method is called more than 1000 times per request in my application because we have a lot of mounted pages and resources and it is called for every single one of them.

**Master**
Benchmark    |                      Mode |  Cnt    |      Score     |      Units
------------ | ------------- |  ------------- |  --: | -------------
ParameterBenchmarks.mergeWith                               |    thrpt  |  3 |  **1746723,436**  | ops/s
ParameterBenchmarks.mergeWith:·gc.alloc.rate.norm           |    thrpt  |  3 |     1600,190  | B/op
ParameterBenchmarks.mergeWith:·gc.churn.G1_Eden_Space.norm  |    thrpt  |  3 |     1602,183  | B/op

**Branch**
Benchmark    |                      Mode |  Cnt    |      Score     |      Units
------------ | ------------- |  ------------- |  --: | -------------
ParameterBenchmarks.mergeWith                                |   thrpt |   3 | **7223753,855** |  ops/s
ParameterBenchmarks.mergeWith:·gc.alloc.rate.norm            |   thrpt |   3 |     320,046 |  B/op
ParameterBenchmarks.mergeWith:·gc.churn.G1_Eden_Space.norm   |   thrpt |   3 |     320,760 |  B/op

https://issues.apache.org/jira/browse/WICKET-6833

### DIFF
--- a/wicket-request/src/main/java/org/apache/wicket/request/mapper/parameter/PageParameters.java
+++ b/wicket-request/src/main/java/org/apache/wicket/request/mapper/parameter/PageParameters.java
@@ -17,7 +17,6 @@
 package org.apache.wicket.request.mapper.parameter;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


svenmeier (approved):

bitstorm (approved):

wicket-request/src/main/java/org/apache/wicket/request/mapper/parameter/PageParameters.java
```diff
@@ -294,35 +293,48 @@ public PageParameters add(final String name, final Object value, final int index
...
-			values.add(value.toString());
+			addNamed(name, value.toString(), index, type);
+		}
+
+		return this;
+	}
+
+	private void addNamed(String name, String[] values, int index, Type type) 
+	{
+		if (namedParameters == null)
```
martin-g, line 310: It would be better to add `&& values.length > 0`

martin-g (approved):

wicket-request/src/main/java/org/apache/wicket/request/mapper/parameter/PageParameters.java
theigl: Done

PR #457, opened 2020-10-09
WICKET-6823: Add a container to AbstractTransformerBehavior to support multiple instances per component.

ams-tschoening:
The current implementation of AbstractTransformerBehavior replaces responses and [doesn't properly take into account](https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6823) use cases in which multiple instances doing the same are assigned to the same component. This adds a container for those use cases to combine all those transformers into one dealing with replacing the responses instead. While that container needs to be used explicitly, it's the easiest implementation to workaround the current limitations.

The container is embedded into `AbstractTransformerBehavior`, because it's small overall, can easily be spotted this way by users and I wasn't too sure about a good alternative name anyway. There would be many possibilities like `TransformerBehaviors`, `TransformerBehavior[Container|Multi]`, `MultiTransformerBehavior` etc.

This PR is based on wicket-8.x instead of master, because that is what I'm using currently and the only thing I'm able to test and build right now.


wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
 import org.apache.wicket.request.http.WebResponse;
 
 /**
- * A {@link Behavior} which can be added to any component. It allows to post-process (transform) the
+ * A {@link Behavior} which can be added to any component, allowing to post-process (transform) the
  * markup generated by the component.
- * 
+ * <p>
+ * There's one important limitation with the current implementation: Multiple different instances of
+ * this behavior CAN NOT be assigned to the same component! If one whiches to do so, the contained
```
martin-g, line 38: s/whiches/wishes/

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
...
+ * container needs to be used to wrap existing behaviors and that container needs to be added to the
+ * component instead. The current implementation of works with temporary responses, but does not
```
martin-g, line 40: s/The current implementation of works with temporary responses/The current implementation works with temporary responses/

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
 import org.apache.wicket.request.http.WebResponse;
 
 /**
- * A {@link Behavior} which can be added to any component. It allows to post-process (transform) the
+ * A {@link Behavior} which can be added to any component, allowing to post-process (transform) the
  * markup generated by the component.
- * 
+ * <p>
+ * There's one important limitation with the current implementation: Multiple different instances of
```
martin-g, line 37: This Javadoc is obsolete with the improvement below, no ?

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
...
+ * this behavior CAN NOT be assigned to the same component! If one whiches to do so, the contained
+ * container needs to be used to wrap existing behaviors and that container needs to be added to the
+ * component instead. The current implementation of works with temporary responses, but does not
+ * support nesting itself properly, which results in missing rendered output and most likely broken
+ * HTML documents in the end.
+ * </p>
  * @see org.apache.wicket.markup.transformer.AbstractOutputTransformerContainer
- * 
+ * @see <a href="https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6823">JIRA issue</a>
```
martin-g, line 45: No need of this link here

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
...
+ *
  * @author Juergen Donnerstag
  */
 public abstract class AbstractTransformerBehavior extends Behavior implements ITransformer
 {
 	private static final long serialVersionUID = 1L;
 
+	/**
+	 * Container to apply multiple {@link AbstractTransformerBehavior} to some component.
```
martin-g, line 54: This javadoc should really be a comment in the PR, not Javadoc.

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
...
+	{
+		private static final long serialVersionUID = 1L;
+
+		/**
+		 * All transformers which need to be applied in the order they need to be applied.
+		 */
+		private final List<AbstractTransformerBehavior> transes;
+
+		/**
+		 * CTOR simply storing the given transformers.
```
martin-g, line 84: s/CTOR/Constructor/

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
...
+	 * @see <a href="https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6823">JIRA issue</a>
+	 */
+	public static class Multi extends AbstractTransformerBehavior
+	{
+		private static final long serialVersionUID = 1L;
+
+		/**
+		 * All transformers which need to be applied in the order they need to be applied.
+		 */
+		private final List<AbstractTransformerBehavior> transes;
```
martin-g, line 81: s/trances/transformers/

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
...
+	 * well.
+	 * </p>
+	 * @see <a href="https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6823">JIRA issue</a>
+	 */
+	public static class Multi extends AbstractTransformerBehavior
+	{
+		private static final long serialVersionUID = 1L;
+
+		/**
+		 * All transformers which need to be applied in the order they need to be applied.
```
martin-g, line 79: Too many "need to be applied". The sentence does not sound good/clear.

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
```diff
@@ -26,17 +31,114 @@
...
+		 * of use at all. In most cases users do have individual transformers to apply only anyway and
+		 * don't need to provide a collection themself this way. OTOH, a collection could be empty or
+		 * contain only one element would then defeat the purpose of this container again.
+		 * </p>
+		 * @param first First transformer to apply.
+		 * @param second Second transformer to apply.
+		 * @param moreIf All other transformers to apply, if at all, in given order.
+		 * @return A container with multiple transformers being applied.
+		 */
+		public static Multi newFor(	AbstractTransformerBehavior		first,
```
martin-g, line 124: s/newFor/of/ ?! Like Model.of() and the new collection methods in Java 9

wicket-core/src/test/java/org/apache/wicket/markup/transformer/AbstractTransformerBehaviorTest.java
```diff
@@ -69,7 +75,77 @@ public void transformationInAjaxRequest()
...
+	 * </p>
+	 * @see <a href="https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6823">JIRA issue</a>
+	 */
+	@Test
+	public void multiTransesSameComp()
+	{
+		TestPage	testPage	= new TestPage();
+		Label		label		= new Label(TestPage.LABEL_ID, TestPage.LABEL_VAL);
+
+		Function<String, AbstractTransformerBehavior> transProd = (val) ->
```
martin-g, line 96: s/transProd/transformerProducer/

martin-g (changes requested):

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
ams-tschoening: I disagree, `AbstractTransformerBehavior` itself keeps the limitation and especially existing users should be made aware. The container is only one possible workaround/solution, a different `AbstractTransformerBehavior` itself would be another one. With that the sentence could be removed, but with the container only the sentence still applies.

martin-g: You can just say "If you need to apply several transformations on the same component then use {@linkplain AbstractTransformerBehavior.Multi}".
I see no need to write 20 lines as Javadoc explaining history.

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
ams-tschoening: As someone who ran into that problem, I prefer having such details in the file to get a basic understanding of problems like these without looking at source history, JIRA etc. I regularly miss such details from Wicket's docs and rephrased things to not sound like a commit message too much anymore.

wicket-core/src/main/java/org/apache/wicket/markup/transformer/AbstractTransformerBehavior.java
ams-tschoening: Those 20 lines of history are exactly what I regularly miss when running into trouble with Wicket, because without them there's no basic understanding of why things are the way they are. When running into this problem, I needed to look at the entire GIT-history just to find no reason for the current implementation, no clue on how things were supposed to work, for a workaround fitting the existing design etc.

The docs often lack important context from my experience, so I'm trying to add one here.

PR #459, opened 2020-10-23
WICKET-6839 Use request flag for caching `isVisibleInHierarchy`

theigl:
This PR aligns caching of `isVisibleInHierarchy` with `isEnabledInHierarchy`.

We are currently clearing the flag `RFLAG_VISIBLE_IN_HIERARCHY_SET` every time the visibility of a component changes, but we never use the flag. This PR adds `RFLAG_VISIBLE_IN_HIERARCHY_VALUE` to cache the result of the computation.

~~I had to change the type of `requestFlags` from `short` to `int` because all available 16 bits are already used up. If my understanding of the JVM memory layout is correct, this should *not* increase memory usage because all object fields use at least one slot of 32 bits anyways (see https://stackoverflow.com/a/27123302/441266).~~

I moved `RFLAG_HAS_REMOVALS` from `MarkupContainer` to `Component` so we do not accidentally assign the same bit to two different flags.

See https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6839

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -421,26 +421,34 @@ public abstract class Component
 	private int flags = FLAG_VISIBLE | FLAG_ESCAPE_MODEL_STRINGS | FLAG_VERSIONED | FLAG_ENABLED |
 		FLAG_IS_RENDER_ALLOWED | FLAG_VISIBILITY_ALLOWED | FLAG_RESERVED5 /* page's stateless hint */;
 
-	private static final short RFLAG_ENABLED_IN_HIERARCHY_VALUE = 0x1;
-	private static final short RFLAG_ENABLED_IN_HIERARCHY_SET = 0x2;
-	private static final short RFLAG_ON_CONFIGURE_SUPER_CALL_VERIFIED = 0x4;
-	private static final short RFLAG_VISIBLE_IN_HIERARCHY_SET = 0x8;
+	// @formatter:off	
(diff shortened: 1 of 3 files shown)


theigl:
The build fails because of Clirr:

```
Error:  6004: org.apache.wicket.Component: Changed type of field RFLAG_CONTAINER_DEQUEING from short to int
Warning:  6003: org.apache.wicket.Component: Value of compile-time constant RFLAG_CONTAINER_DEQUEING has been changed
Error:  7005: org.apache.wicket.Component: Parameter 1 of 'protected boolean getRequestFlag(short)' has changed its type to int
```

Neither the constant nor the method are part of the public API. Is there any way to tell Clirr to ignore these changes?

svenmeier:
Actually we still have one spare bit (short)0x8000 ... yes this is still a short :P

I'd suggest you keep the change small:
0x4 has been the value I've borrowed when adding RFLAG_ON_CONFIGURE_SUPER_CALL_VERIFIED, so it's time to reclaim it for the hierarchy:

```
private static final short RFLAG_VISIBLE_IN_HIERARCHY_SET_VALUE = 0x4;
private static final short RFLAG_ON_CONFIGURE_SUPER_CALL_VERIFIED = (short)0x8000;

```

theigl:
@svenmeier: I wasn't aware of that! I just pushed the simplified change.

svenmeier (approved):
I don't care about the formatting.

martin-g (approved):

theigl:
Thank you @svenmeier and @martin-g!

PR #460, opened 2020-11-13
Added method to set the component's renderer.

fincatto:
After created, the component does not allow to set the renderer and the class has many constructors.


svenmeier:
Hi @fincatto, thanks for your contribution.

I see one issue with your proposed change:
The nested DropDownChoice keeps its own reference to the renderer. So setting the renderer after initilialization of the editor (the DropDownChoice) will have no effect.

fincatto:
Hi @svenmeier.
I saw that passing the null model at startup has no practical effect on the class, so I'm using the parameter of the constructor.
I will close the PR and use it this way, which already serves me.
Thank you!

PR #461, opened 2021-01-20
WICKET-6862: Add @Autowired annotation to SpringInjector

bhuism:
Besides annotations @Inject and @Springbean also support @Autowired annotation for spring bean injection

### DIFF
--- a/wicket-spring/src/main/java/org/apache/wicket/spring/injection/annot/AnnotProxyFieldValueFactory.java
+++ b/wicket-spring/src/main/java/org/apache/wicket/spring/injection/annot/AnnotProxyFieldValueFactory.java
@@ -36,6 +36,7 @@ import org.apache.wicket.util.lang.Generics;
 import org.apache.wicket.util.string.Strings;
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryUtils;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.support.AbstractBeanDefinition;
@@ -117,18 +118,26 @@ public class AnnotProxyFieldValueFactory implements IFieldValueFactory
 		{
 			SpringBean annot = field.getAnnotation(SpringBean.class);
 
-			String name;
-			boolean required;
+			String name = null;
+			Boolean required = true;
+
 			if (annot != null)
 			{
 				name = annot.name();
 				required = annot.required();
 			}
-			else
+
+			Autowired autowiredAnnotation = field.getAnnotation(Autowired.class);
+
+			if (autowiredAnnotation != null)
+			{
+				required = autowiredAnnotation.required();
+			}
+
+			if (name == null)
 			{
 				Named named = field.getAnnotation(Named.class);
 				name = named != null ? named.value() : "";
-				required = true;
 			}
 
 			Class<?> generic = ResolvableType.forField(field).resolveGeneric(0);
@@ -324,6 +333,6 @@ public class AnnotProxyFieldValueFactory implements IFieldValueFactory
 	@Override
 	public boolean supportsField(final Field field)
 	{
-		return field.isAnnotationPresent(SpringBean.class) || field.isAnnotationPresent(Inject.class);
+		return field.isAnnotationPresent(SpringBean.class) || field.isAnnotationPresent(Inject.class) || field.isAnnotationPresent(Autowired.class);
 	}
 }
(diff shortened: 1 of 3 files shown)


wicket-spring/src/main/java/org/apache/wicket/spring/injection/annot/AnnotProxyFieldValueFactory.java
```diff
@@ -117,18 +118,26 @@ public Object getFieldValue(final Field field, final Object fieldOwner)
 		{
 			SpringBean annot = field.getAnnotation(SpringBean.class);
 
-			String name;
-			boolean required;
+			String name = null;
+			Boolean required = true;
```
solomax, line 122: I would declare this one as `boolean` it seems to be not nullable :)

bitstorm (approved):

svenmeier (changes requested):
@Autowired is Spring's own injection annotation.

Since Spring will not actually do the injection on Wicket components, I don't see the need to add support for @Autowired here.

wicket-spring/src/main/java/org/apache/wicket/spring/injection/annot/AnnotProxyFieldValueFactory.java
bhuism: Good idea, thanks

bhuism:
@svenmeier I agree, it is the spring (defacto) standard injection annotation, and does not _need_ to be added. The rationale is for consistency. In normal spring beans one can (obviously) not use wickets @SpringBean. The other option @Inject is not the spring defacto standard, I think new wicket users (who know spring) will find it intuitive (or even expect) to be able to use @Autowired

ps: I would love to see that Spring does the injection, buts thats another story

svenmeier:
This change needs to be discussed. Please open a Jira issue so other core devs are aware of this.

bhuism:
@svenmeier done, please see: https://issues.apache.org/jira/browse/WICKET-6862

martin-g:
I agree with @svenmeier - by using `@SpringBean` one knows who is responsible to do the injection - Wicket.
If we start supporting `Autowired` then soon we will have to take care for `@Qualifier` too...
If Spring's BeanFactory had a method like `getBean(Field)` then maybe this would have been cleaner. Just pass the field to Spring and let it deal with all possible annotations it supports.

Actually `SpringBean` is an artefact from the past. The idea was everyone (Spring included) to move to javax.inject APIs but this never happened.

bhuism:
Hmm, I hear 'we' want to migrate away from @SpringBean (a thing from the past, and unknown to non wicket users) and away from @Inject because it was never adopted. Can this PR be seen as a step in that direction? toward using spring to do the actual injection? am trying to make wicket better, for widest audience possible. And won't be offended if it doesn't get merged.

martin-g:
We cannot drop `@SpringBean` in a minor version. The earliest would be Wicket v 10.0.0. We can deprecate it though!
But I'd prefer to use `@javax.inject.Inject` instead of `@Autowired`. I doubt that Spring people will ever do it but they can also decide to drop `@Autowired` in favor of `@Inject`

`@Inject` is already supported by wicket-spring, wicket-guice and wicketstuff-jee.

bhuism:
I'll doubt that too, It is a _spring_-injector, because of that I though following spring semantics (`@Inject` and `@Autowired`) would be the way to go. Would you or @svenmeier consider supporting `@Autowired` when spring does the injection?

martin-g:
> when spring does the injection

What do you mean ?

At the moment Wicket delegates to Spring/Guice to find the respective bean and then Wicket just sets the Field value.
What do you want to change in this area ?

bhuism:
Yes, I mean just the setting of the fields part, maybe we can leverage spring (more) in doing the actual injection, a quick search in the spring api gave me:

[`AutowireCapableBeanFactory.autowire(...)`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/config/AutowireCapableBeanFactory.html#autowireBeanProperties-java.lang.Object-int-boolean-)

I think @svenmeier also meant this in https://github.com/apache/wicket/pull/461#pullrequestreview-573137771

This way maybe also annotations like `@Value` can be supported in the future, or other spring bean features.

martin-g:
This looks promising!

svenmeier:
>Would you or @svenmeier consider supporting @Autowired when spring does the injection?

Yes. I remember trying with AutowireCapableBeanFactory a decade ago, but if I recall correctly I didn't find a solution to interject the Wicket proxies.

martin-g:
> Wicket proxies

Right! Now I remember this problem too - Wicket needs to inject a Serializable proxy to be able to store the page. Spring returns the bean as is - it could be Serializable by chance but most of the service classes are not.

bhuism:
Ok, I'll close it. I did expect the _spring_-injector to follow _spring_ semantics as much as possible. I had a different idea with the functionality.

@martin-g is it not the case that beans are not serialized with the page (sorry for not checking) i.e. nulled before serialization, and re-injected after deserialization? if the beans _are_ serialized, why would you want the (stateless) beans to be part of the page serialisation state? I am curious.

martin-g:
Wicket does not keep a hard reference to the Spring/Guice/JEE beans.
Wicket injects a Serializable proxy that lazy loads the bean on first usage. The lazy loading is request scoped, i.e. the beans are transient and the proxy re-loads them for each http request.

bhuism:
a, thanks @martin-g I thought as much

PR #462, opened 2021-02-07
WICKET-6864 updated crypt configuration

svenmeier:
It's time to update the crypt defaults to recommended standards.


bitstorm (approved):

wicket-core/src/main/java/org/apache/wicket/authentication/strategy/DefaultAuthenticationStrategy.java
```diff
@@ -57,25 +65,50 @@
...
-			return Application.get().getName();
-		}
-		return cookieKey;
+		return UUID.randomUUID().toString();
 	}
 
+	/**
+	 * @deprecated supply a crypt instead TODO remove in Wicket 10
+	 */
+	@Deprecated
```
solomax, line 85: Mayde it worth to add `forRemoval` instead of `TODO` :) ?

solomax (approved):

papegaaij:
I think this `SunJceCrypt` requires a lot more work to get it up to date. For example, `PBEWithMD5AndDES` is hopelessly outdated. For the cipher we should definitely not use DES, but AES, like `AES/CBC/PKCS5Padding`. 

If we want to use PBE, we should switch to PBKDF2 and use PBKDF2WithHmacSHA512 with a key-length of 256 bits and a lot of iterations (way more than 1000, probably 100.000), but actually I fail to see why this class uses password based encryption and not a key directly. IMHO the key should not be a string, but a SecretKey built from 256 bits of secure random.

Note we already have `UnlimitedStrenghtJurisdictionPolicyCrypt` in a test in `wicket-util`. This implementation already is much better and this unlimited strength jce-restriction is not a real issue anymore. I don't know about the latest Oracle JVMs, but OpenJDK does not limit the strength of JCE.

IMHO it's better to deprecate the whole class and replace it with a more secure version.

svenmeier:
Even better if we have an improved default implementation +1.

IMHO the worst thing about the current state is having a fixed salt and DefaultAuthenticationStrategy using the application name as default key.

papegaaij:
`UnlimitedStrenghtJurisdictionPolicyCrypt` is much better wrt the algorithms used, however, the unpredictability of keys, salt and initialization vectors (iv) is even more important. It makes no sense to encrypt something with a key that can be guessed with just a couple of tries, no matter what algorithm you use. Unfortunately I'm very limited in time at the moment. We've got a very tight schedule at the moment. But IMHO, we should make sure not only the algorithms are up to date, but the input for the keys, salt and iv is secure random as well.

bitstorm:
I also agree we should provide a stronger default crypt and UnlimitedStrenghtJurisdictionPolicyCrypt is a good candidate. Unfortunately the current cryptography support is tailored for PB encryption, that's why some time ago I've tried to improve it to allow the adoption of other algorithms. At that time I couldn't finalize the work as we were focused on releasing Wicket 7, but the branch is still around: https://github.com/apache/wicket/compare/WICKET-5768-improve-crypt.
It's nothing more than a palyground branch but I think it might contain some useful ideas.

wicket-core/src/main/java/org/apache/wicket/authentication/strategy/DefaultAuthenticationStrategy.java
```diff
@@ -16,12 +16,14 @@
  */
 package org.apache.wicket.authentication.strategy;
 
-import org.apache.wicket.Application;
+import java.util.Random;
```
martin-g, line 19: Is Random used in this class ?

wicket-core/src/main/java/org/apache/wicket/authentication/strategy/DefaultAuthenticationStrategy.java
```diff
@@ -30,6 +32,9 @@
 /**
  * Wicket's default implementation of an authentication strategy. It'll concatenate username and
  * password, encrypt it and put it into one Cookie.
+ * <p>
+ * Note: To support automatic authentication across application restarts you have to use
+ * the constructor {@link DefaultAuthenticationStrategy#DefaultAuthenticationStrategy(String, String, byte[])}.
```
martin-g, line 37: I don't see such constructor below.

wicket-core/src/main/java/org/apache/wicket/authentication/strategy/DefaultAuthenticationStrategy.java
```diff
@@ -57,25 +65,50 @@
...
+
+	/**
+	 * Constructor
+	 * 
+	 * @param cookieKey
+	 *            The name of the cookie
+	 * @param crypt
+	 *            the crypt
+	 */
+	public DefaultAuthenticationStrategy(final String cookieKey, ICrypt crypt)
```
martin-g, line 108: I guess this is the stable constructor that should be used by the applications if they want ti be able to decrypt after restarts. Let's add a comment to make it clear/recommended.

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/KeyInSessionSunJceCryptFactory.java
```diff
@@ -89,25 +91,44 @@ public ICrypt newCrypt()
...
+		SunJceCrypt crypt = new SunJceCrypt(cryptMethod, data.salt, 1000);
+		crypt.setKey(data.key);
+		
 		return crypt;
 	}
 
 	/**
 	 * @return the {@link org.apache.wicket.util.crypt.ICrypt} to use
+	 * 
+	 * @deprecated this method is no longer called
```
martin-g, line 117: Please add "TODO Remove in Wicket 10" or `forRemoval=10.0.0`

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/KeyInSessionSunJceCryptFactory.java
```diff
@@ -16,8 +16,10 @@
  */
 package org.apache.wicket.core.util.crypt;
 
+import java.io.Serializable;
 import java.security.Provider;
 import java.security.Security;
+import java.util.Random;
```
martin-g, line 22: Again, looks like this new import is not used.

wicket-util/src/main/java/org/apache/wicket/util/crypt/ClassCryptFactory.java
```diff
@@ -28,6 +28,8 @@
  * must implement {@link ICrypt}.
  * 
  * @author Igor Vaynberg (ivaynberg)
+ * 
+ * @deprecated use a lambda expression instead TODO remove in Wicket 10
  */
 public class ClassCryptFactory implements ICryptFactory
```
martin-g, line 34: Do we still need `ICryptFactory` ? Or it should be deprecated too ?

wicket-util/src/main/java/org/apache/wicket/util/crypt/SunJceCrypt.java
```diff
@@ -44,39 +45,77 @@
...
+	public SunJceCrypt(byte[] salt, int iterationCount)
+	{
+		this(DEFAULT_CRYPT_METHOD, salt, iterationCount);
+	}
+
+	/**
+	 * Constructor
+	 *
+	 * @deprecated
+	 */
```
martin-g, line 96: `@Deprecated`

wicket-util/src/main/java/org/apache/wicket/util/crypt/SunJceCrypt.java
```diff
@@ -44,39 +45,77 @@
...
+	 *              salt for encryption
+	 * @param iterationCount
+	 * 				iteration count
 	 */
-	public SunJceCrypt(String cryptMethod)
+	public SunJceCrypt(String cryptMethod, byte[] salt, int iterationCount)
 	{
 		this.cryptMethod = Args.notNull(cryptMethod, "Crypt method");
+		this.salt = Args.notNull(salt, "salt");
+		this.iterationCount = iterationCount;
```
martin-g, line 118: do we need a check for a positive `iterationCount` ?

wicket-util/src/main/java/org/apache/wicket/util/crypt/SunJceCrypt.java
```diff
@@ -138,12 +177,13 @@ protected SecretKey generateSecretKey() throws NoSuchAlgorithmException,
 		return keyFactory.generateSecret(spec);
 	}
 
+
```
martin-g, line 180: the rest of the code uses single empty line between the methods

wicket-util/src/main/java/org/apache/wicket/util/crypt/SunJceCrypt.java
```diff
@@ -153,4 +193,17 @@ protected KeySpec createKeySpec()
...
 	}
+
+	/**
+	 * Create a random salt.
+	 * 
+	 * @return salt
+	 */
+	public static byte[] randomSalt()
+	{
+		// only 8 bytes long supported
```
martin-g, line 204: Please add an explanation `why` only 8 bytes long is supported. I honestly don't know why more is not possible. I guess with the time even you may forget the reason.

wicket-util/src/main/java/org/apache/wicket/util/crypt/ClassCryptFactory.java
svenmeier: CryptoMapper still utilizes SecuritySettings#cryptFactory -> ICryptFactorty

svenmeier:
Thanks for your feedback.

Of course developers can always configure their own crypt implementation. But Wicket should use sensible defaults too - currently this is not the case with a hard-coded salt and DefaultAuthenticationStrategy's default encryption-key.

I'm fine with anybody stepping up and building a better solution. As Andrea pointed out, this might get difficult with the current API.

bitstorm:
With my commit I've tried to bring in some ideas to make Wicket more independent from the specific solution currently adopted (a password-based DES cypher):

- I've created a new class GenericJceCrypt, which should serve as base class to implement a specific JCE chyper. This class is meant to replace AbstractCrypt in Wicket 10.
- I've also created class AbstractKeyInSessionCryptFactory to provide a base class to store a session relative key used for URL encryption. Again, the current class KeyInSessionSunJceCryptFactory only supports password based algorithms. The new class should simplify the implementation of session crypt factories that use more complex values as key (for exemple a  javax.crypto.SecretKey).
- Finally I've deprecated ICyper#setKey because it supports just string key and (IMHO) in any case it shouldn't be partof the API contract.  

Let me know what you think about this.

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/KeyInSessionSunJceCryptFactory.java
```diff
@@ -37,7 +36,7 @@
  *
  * @author igor.vaynberg
  */
-public class KeyInSessionSunJceCryptFactory implements ICryptFactory
+public class KeyInSessionSunJceCryptFactory extends AbstractKeyInSessionCryptFactory<CryptData>
 {
 	/** metadata-key used to store crypt data in session metadata */
 	private static final MetaDataKey<CryptData> KEY = new MetaDataKey<>()
```
solomax, line 42: this `KEY` can be removed

wicket-util/src/main/java/org/apache/wicket/util/crypt/GenericJceCrypt.java
```diff
@@ -0,0 +1,162 @@
...
+
+
+/**
+ * Base class for JCE based ICrypt implementations.
+ * 
+ */
+public class GenericJceCrypt implements ICrypt
+{
+	/** Encoding used to convert java String from and to byte[] */
+	private static final String CHARACTER_ENCODING = "UTF-8";
```
solomax, line 40: Maybe it would be better to use `StandardCharsets.UTF8` ?

svenmeier:
Fine by me.  @papegaaij wdyt?

papegaaij:
I like the idea. The new implementation has one problem though: the IV must be random on every encryption step. Most importantly, you should never encrypt the same data twice with the same IV. You can think of the IV as a salt for your encryption. It ensures that the same data encrypts to different values every time. It is common practice to store the IV together with the ciphertext. The code we use in our application for AES256 is like this:

```
	public byte[] encrypt(SecureRandom rnd, SecretKey key, byte[] plaintext) throws GeneralSecurityException {
		Cipher cipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
		cipher.init(Cipher.ENCRYPT_MODE, key, rnd);
		AlgorithmParameters params = cipher.getParameters();
		byte[] iv = params.getParameterSpec(IvParameterSpec.class).getIV();
		byte[] ciphertext = cipher.doFinal(plaintext);
		return Bytes.concat(iv, ciphertext);
	}

	public byte[] decrypt(SecureRandom rnd, SecretKey key, byte[] cipherInput) throws GeneralSecurityException {
		byte[] iv = new byte[16];
		byte[] ciphertext = new byte[cipherInput.length - 16];
		System.arraycopy(cipherInput, 0, iv, 0, iv.length);
		System.arraycopy(cipherInput, 16, ciphertext, 0, ciphertext.length);

		Cipher cipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
		cipher.init(Cipher.DECRYPT_MODE, key, new IvParameterSpec(iv), rnd);
		return cipher.doFinal(ciphertext);
	}
```

bitstorm:
I followed Emond's suggestion and I turned GenericJceCrypt into a more generic AbstractJceCrypt. The idea behind this class is to let final users implement the desired configuration logic for ciphers, so they can decide which level of "strength" to adopt. For example in some use cases we could decide to sacrifice security strength  for a faster encryption/decryption.  This might be the case of class CryptoMapper which uses ICypt to encrypt/decrypt page URLs.

PS:  class UnlimitedStrenghtJurisdictionPolicyCrypt has been reworked  according to the example code provided by Edmond.

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/AbstractKeyInSessionCryptFactory.java
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Base class to implement crypt factories that store crypt into user session. Note that the use of
+ * this crypt factory will result in an immediate creation of a http session.
+ * 
+ * @author andrea del bene
+ *
+ * @param <T>
+ *            the type for the secret key.
+ */
+public abstract class AbstractKeyInSessionCryptFactory<T extends Serializable>
```
martin-g, line 36: s/Serializable/IClusterable/ ?!

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/AbstractKeyInSessionCryptFactory.java
```diff
@@ -0,0 +1,91 @@
...
+	implements
+		ICryptFactory
+{
+	/** metadata-key used to store crypto-key in session metadata */
+	private static final MetaDataKey<Serializable> KEY = new MetaDataKey<Serializable>()
```
martin-g, line 41: MetaDataKey<T>

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/AbstractKeyInSessionCryptFactory.java
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	@Override
+	@SuppressWarnings("unchecked")
+	public ICrypt newCrypt()
+	{
+		Session session = Session.get();
+		session.bind();
+
+		// retrieve or generate encryption key from session
+		Serializable key = session.getMetaData(KEY);
```
martin-g, line 65: s/Serializable/T/

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/AbstractKeyInSessionCryptFactory.java
```diff
@@ -0,0 +1,91 @@
...
+		if (key == null)
+		{
+			// generate new key
+			key = generateKey(session);
+			session.setMetaData(KEY, key);
+		}
+
+		// build the crypt based on session key
+		ICrypt crypt = createCrypt((T)key);
```
martin-g, line 74: if the above changes are applied then there is no need to cast here and to annotate with `@SuppressWarnings("unchecked")`

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/KeyInSessionSunJceCryptFactory.java
```diff
@@ -82,32 +76,53 @@ public KeyInSessionSunJceCryptFactory(String cryptMethod)
...
+	@Override
+	protected ICrypt createCrypt(CryptData keyParams)
+	{
+	    SunJceCrypt crypt = new SunJceCrypt(cryptMethod, keyParams.salt, 1000);
+        crypt.setKey(keyParams.key);
+        
+        return crypt;
+	}
+
+    static final class CryptData implements Serializable
```
martin-g, line 111: IClusterable

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/KeyInSessionSunJceCryptFactory.java
```diff
@@ -82,32 +76,53 @@ public KeyInSessionSunJceCryptFactory(String cryptMethod)
...
+	{
+		/**
```
martin-g, line 113: Empty javadoc. Remove it

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/KeyInSessionSunJceCryptFactory.java
```diff
@@ -82,32 +76,53 @@ public KeyInSessionSunJceCryptFactory(String cryptMethod)
...
+         * 
+         */
+        private static final long serialVersionUID = 1L;
+
+        final String key;
```
martin-g, line 118: the indentation is wrong

wicket-core/src/main/java/org/apache/wicket/settings/SecuritySettings.java
```diff
@@ -49,8 +49,11 @@
 public class SecuritySettings
 {
 	/**
-	 * encryption key used by default crypt factory
+	 * encryption key is no longer used
+	 * 
+	 * @deprecated
 	 */
+	@Deprecated
```
martin-g, line 56: forRemoval

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
```diff
@@ -0,0 +1,122 @@
...
+	public static final Charset CHARACTER_ENCODING = StandardCharsets.UTF_8;
+
+	/** Log. */
+	private static final Logger log = LoggerFactory.getLogger(AbstractJceCrypt.class);
+
+    protected Cipher buildCipher(int opMode, SecretKey secretKey, String transformation, AlgorithmParameterSpec params)
+    {
+        try 
+        {
+            Cipher crypter = Cipher.getInstance(transformation);
```
martin-g, line 50: s/crypter/cipher/

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
```diff
@@ -0,0 +1,122 @@
...
+	public final String decryptUrlSafe(final String text)
+	{
+		try
+		{
+			byte[] decoded = java.util.Base64.getUrlDecoder().decode(text);
+			return new String(decryptByteArray(decoded), CHARACTER_ENCODING);
+		}
+		catch (Exception ex)
+		{
+			log.debug("Error decoding text: " + text, ex);
```
martin-g, line 76: Use slf4j's {} instead of String concatenation

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
```diff
@@ -0,0 +1,122 @@
...
+	 * Encrypt a string into a string using URL safe Base64 encoding.
+	 * 
+	 * @param plainText
+	 *            text to encrypt
+	 * @return encrypted string
+	 */
+	@Override
+	public final String encryptUrlSafe(final String plainText)
+	{
+		byte[] encrypted = encryptStringToByteArray(plainText);
```
martin-g, line 91: indentation

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
```diff
@@ -0,0 +1,122 @@
...
+
+	/**
+	 * Encrypts the given text into a byte array.
+	 * 
+	 * @param plainText
+	 *            text to encrypt
+	 * @return the string encrypted
+	 * @throws GeneralSecurityException
+	 */
+	abstract protected byte[] encryptStringToByteArray(final String plainText);
```
martin-g, line 115: Do we need these long method names ? The return type and the argument type already give this info. Just `encrypt` is enough to me

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
```diff
@@ -0,0 +1,122 @@
...
+	
+
+    @Override
+    public void setKey(String key) {
```
martin-g, line 119: `final` ?!

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
```diff
@@ -0,0 +1,122 @@
...
+        throw new UnsupportedOperationException();
```
martin-g, line 120: A message why it is not supported would be good

wicket-util/src/main/java/org/apache/wicket/util/crypt/SunJceCrypt.java
```diff
@@ -153,4 +194,18 @@ protected KeySpec createKeySpec()
...
+	 * Create a random salt to be used for this crypt. 
+	 * 
+	 * @return salt, always 8 bytes long
+	 */
+	public static byte[] randomSalt()
+	{
+		// must be 8 bytes - for anything else PBES1Core throws
+		// InvalidAlgorithmParameterException: Salt must be 8 bytes long  
+		byte[] salt = new byte[8];
+		new Random().nextBytes(salt);
```
martin-g, line 208: Should we use CipherUtils.randomByteArray(8) here ? It would use SecureRandom

wicket-util/src/test/java/org/apache/wicket/util/crypt/AESCryptTest.java
```diff
@@ -0,0 +1,55 @@
...
+		SecretKey secretKey = CipherUtils.generatePBEKey(
+			"myWeakPassword", "PBKDF2WithHmacSHA1", "AES", 
+			CipherUtils.randomByteArray(16), 65536, 256);
+		
+		AbstractJceCrypt crypt = new AESCrypt(secretKey);
+
+		String input1 = "input1";
+		String encrypted = crypt.encryptUrlSafe(input1);
+
+		String input2 = "input2";
```
martin-g, line 46: let's use some UTF-8 characters here

wicket-util/src/main/java/org/apache/wicket/util/crypt/CipherUtils.java
```diff
@@ -0,0 +1,107 @@
...
+	 * Generates a random generated byte array of the given size.
+	 * 
+	 * @param size
+	 *              the size of the array.
+	 * @return the new byte array of the given size.
+	 */
+	public static byte[] randomByteArray(int size)
+	{
+		byte[] iv = new byte[size];
+		new SecureRandom().nextBytes(iv);
```
solomax, line 104: Maybe it would be better to have only one instance of `SecureRandom`?

wicket-util/src/main/java/org/apache/wicket/util/crypt/CipherUtils.java
```diff
@@ -0,0 +1,107 @@
...
+	 * @param keyLength
+	 *              the key length
+	 * @return a new {@link SecretKey}
+	 */
+	public static SecretKey generateKey(String algorithm, int keyLength)
+	{
+		try
+		{
+			KeyGenerator keyGenerator = KeyGenerator.getInstance(algorithm);
+			keyGenerator.init(keyLength);
```
papegaaij, line 50: Pass in a SecureRandom here as well using the one from SecuritySettings.getRandomSupplier().getRandom(). That will make sure we use one reliable source of random and it's customizable by the user.

wicket-util/src/main/java/org/apache/wicket/util/crypt/CipherUtils.java
papegaaij: Yes, creating a SecureRandom over and over will exhaust your systems entropy. This method already exists in ISecureRandomSupplier (SecuritySettings.getRandomSupplier), so it can be removed.

wicket-util/src/main/java/org/apache/wicket/util/crypt/SunJceCrypt.java
papegaaij: Better yet, remove this method entirely. Usage of this method should be replaced by ISecureRandomSupplier.getRandomBytes(length). The optimal length of the salt depends on the algorithm used. Some might require 8 bytes, but for example PBKDF2WithHmacSHA512 can use much longer salts and it's recommended to use a salt that's equal in length to your derived key.

wicket-util/src/main/java/org/apache/wicket/util/crypt/AESCrypt.java
```diff
@@ -0,0 +1,112 @@
...
+
+	/**
+	 * Constructor
+	 * 
+	 * @param secretKey
+	 *              The {@link SecretKey} to use to initialize the {@link Cipher}.
+	 * @param algorithm
+	 *              The cipher algorithm to use, for example "AES/CBC/PKCS5Padding".
+	 * @param ivSize
+	 *              The size of the Initialization Vector to use with the cipher.
```
papegaaij, line 58: The ivSize is not needed. It is always equal to the blocksize of the cipher. So you can simply use cipher.getBlockSize() to get it.

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/AbstractKeyInSessionCryptFactory.java
bitstorm: To use the generic type T filed KEY must be an instance field, I don't think it's a problem even if we usually have static fields for metadata keys. WDYT?

martin-g: I see. +1 if it works!

wicket-util/src/main/java/org/apache/wicket/util/crypt/CipherUtils.java
bitstorm: Unfortunately we are talking of classes from module wicket-util which can't see or use classes Application  or ISecureRandomSupplier. But we could move AESCrypt to wicket-core module so it can access to application setting.
WDYT?

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
bitstorm: I don't think it can be used if we want to log the exception as well. Am I wrorng?

solomax: `log.debug("Error decoding text: {}", text, ex);` will work as expected: will print message and trace :)

wicket-util/src/main/java/org/apache/wicket/util/crypt/CipherUtils.java
papegaaij: Having access to a good source of secure random is rather critical for most forms of cryptography. It's unfortunate that these classes live in a place that does not have direct access to the random managed by the application. I'd say move AbstractJceCrypt, AESCrypt and everything that goes with it to core. AbstractJceCrypt can then have convenience methods to fetch the SecureRandom from the application. That way we hopefully encourage other developers to use this as well if they want to implement their own.

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
bitstorm: Nice!

wicket-util/src/test/java/org/apache/wicket/util/crypt/AESCryptTest.java
bitstorm: We already use it inside AbstractJceCrypt and AESCrypt.

wicket-util/src/main/java/org/apache/wicket/util/crypt/AbstractJceCrypt.java
bitstorm: Good point. Now the abstract class define just "encrypt" and "decrypt" , both operating on byte arrays.

wicket-util/src/main/java/org/apache/wicket/util/crypt/SunJceCrypt.java
bitstorm: Since this is a legacy class I would leave it as it is as it can not be moved to wicket-core

wicket-util/src/main/java/org/apache/wicket/util/crypt/AESCrypt.java
bitstorm: Right, but during decipher phase you need ivSize before creating the cipher reading the first ivSize of encrypted data. So I would leave this constructor parameter

papegaaij: But that's only because the lookup and unit of the Cipher are in `buildCipher`. IMHO that method does not really make things easier to understand. I think the code would be better with that method unlined. Then you can bring the Cipher lookup to the front and use that to get the block size.

This is just my opinion. If you do decide to keep the method, then rename the field to `blockSize`, as that is what it is.

bitstorm: Sorry but you are loosing me here. How could you build a Cipher in DECRYPT_MODE without already having the expected IV (not just the size of it)?  Is it possible to write a version of your method _decrypt_ posted above without knowing the fixed size of IV?

papegaaij: You get the Cipher with just `Cipher.getInstance(alg)`. There you can get the block size from. You only need the iv for the call to `init`. I'm typing this on my phone, so posting code is a bit difficult. What about this?

```public byte[] decrypt(SecureRandom rnd, SecretKey key, byte[] cipherInput) throws GeneralSecurityException {
    Cipher cipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
    int blocksize = cipher.getBlockSize();
    byte[] iv = new byte[blocksize];
    byte[] ciphertext = new byte[cipherInput.length - blocksize];
    System.arraycopy(cipherInput, 0, iv, 0, iv.length);
    System.arraycopy(cipherInput, blocksize, ciphertext, 0, ciphertext.length);

    
    cipher.init(Cipher.DECRYPT_MODE, key, new IvParameterSpec(iv), rnd);
    return cipher.doFinal(ciphertext);
}```

Comment (bitstorm) [reply wicket-util/src/main/java/org/apache/wicket/util/crypt/AESCrypt.java]: Thank you, now I see... :-)

**Review by papegaaij — APPROVED** [review]
Looks good to me. Great work on bringing the ciphers back up to date!

File: wicket-core/src/main/java/org/apache/wicket/core/util/crypt/AESCrypt.java [hunk 33]
```diff
@@ -0,0 +1,132 @@
...
+ * Based on http://stackoverflow.com/a/992413
+ * 
+ * @see ICrypt
+ */
+public class AESCrypt extends AbstractJceCrypt
+{
+
+	private final SecretKey secretKey;
+	private final String algorithm;
+	private ISecureRandomSupplier randomSupplier;
```
martin-g, line 45: final

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/AbstractJceCrypt.java
```diff
@@ -0,0 +1,105 @@
...
+
+
+/**
+ * Base class for JCE based ICrypt implementations.
+ * 
+ */
+public abstract class AbstractJceCrypt implements ICrypt
+{
+	/** Encoding used to convert java String from and to byte[] */
+	public static final Charset CHARACTER_ENCODING = StandardCharsets.UTF_8;
```
martin-g, line 36: Any reason this to be `public` ?
Why not just use `StandardCharsets.UTF_8` ? If the encoding should be configurable then there should be a constructor parameter for it.

wicket-core/src/main/java/org/apache/wicket/core/util/crypt/AbstractKeyInSessionCryptFactory.java
```diff
@@ -0,0 +1,89 @@
...
+	implements
+		ICryptFactory
+{
+	/** metadata-key used to store crypto-key in session metadata */
+	private final MetaDataKey<T> KEY = new MetaDataKey<T>()
+	{
+		private static final long serialVersionUID = 1L;
+	};
+
+	public AbstractKeyInSessionCryptFactory()
```
martin-g, line 45: No need of this constructor, it is implicit

wicket-core/src/test/java/org/apache/wicket/core/util/crypt/AESCryptTest.java
```diff
@@ -0,0 +1,54 @@
...
+		SecretKey secretKey = CipherUtils.generatePBEKey(
+			"myWeakPassword", "PBKDF2WithHmacSHA1", "AES", 
+			randomSupplier.getRandomBytes(16), 65536, 256);
+		
+		AbstractJceCrypt crypt = new AESCrypt(secretKey, randomSupplier);
+
+		String input1 = "input1";
+		String encrypted = crypt.encryptUrlSafe(input1);
+
+		String input2 = "input2";
```
martin-g, line 45: In my previous review I have suggested to use a more fancy input, e.g. one with umlauts or cyrillic characters, to test UTF-8, not just ASCII

bitstorm:
@martin-g if everything is ok now let me now and I will proceed with the merge

martin-g (approved):
Approved!
But please either edit the commit messages or squash them before merging to master!

PR #463, opened 2021-02-20
WICKET-6868 ajax submit allow calling onsubmit on form

svenmeier:
In Wicket 7.x a form's `onsubmit()` was called for multipart Ajax submits only. This made sense, since multipart requests were submitted with a hidden form.

Since Wicket 8.x `onsubmit()` is no longer called, since a multipart Ajax submit is now performed like any other Ajax request. For consistency I would not want to go back to calling `onsubmit()` for multipart submits only.
But we can't call `onsubmit()` for *all* Ajax submit either, as this would be a change to a much more frequent usage.

Thus I propose to make this configurable with a new `AjaxFormSubmitBehavior#shouldTriggerJavaScriptSubmitEvent()`, default would be `false`.


svenmeier:
I improved the solution, so all event 'submit' handlers are triggered.

svenmeier:
Merged

PR #465, opened 2021-03-17
Update pom.xml

hhovhann:
Fix incompatibilities with new plugin version


solomax (approved):
good catch!

martin-g:
Thank you, @hhovhann !

hhovhann:
Cool, happy that was a helpful catch.

PR #466, opened 2021-04-07
add an AJAX behavior to transfer client side file information to ser…

reiern70:
Add an AJAX behavior to transfer client side file information to server side when a file is selected for upload. Now in AJAX upload example, when user selects a file, server side gets notified... with no need to submit file

image

IMHO this could be useful for instance to enable/disable some actions depending on file size and so on.

PR #467, opened 2021-04-08
[WICKET-6876] add an AJAX behavior to transfer client side file infor…

reiern70:
[WICKET-6876] add an AJAX behavior to transfer client side file information to server side (when a file is selected for upload at client side).

This PR replaces 

https://github.com/apache/wicket/pull/466

I have included comments/remarks provided there. Now also submit buttons will be disabled is chosen file is bigger than allowed

image

and enabled in case file is Ok

image

Thus this illustrates how AJAX behavior can be used to do on the spot "minimum validation" without need to submit file.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.java
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.markup.html.form.upload;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.wicket.Component;
+import org.apache.wicket.WicketRuntimeException;
+import org.apache.wicket.ajax.AjaxRequestTarget;
+import org.apache.wicket.ajax.attributes.AjaxRequestAttributes;
+import org.apache.wicket.ajax.attributes.IAjaxCallListener;
+import org.apache.wicket.ajax.form.OnChangeAjaxBehavior;
+import org.apache.wicket.markup.head.IHeaderResponse;
+import org.apache.wicket.markup.head.JavaScriptHeaderItem;
+import org.apache.wicket.request.IRequestParameters;
+import org.apache.wicket.request.Request;
+import org.apache.wicket.request.cycle.RequestCycle;
+import org.apache.wicket.request.resource.JavaScriptResourceReference;
+import org.apache.wicket.request.resource.ResourceReference;
+import org.apache.wicket.util.lang.Args;
+import org.apache.wicket.util.string.StringValue;
+import org.danekja.java.util.function.serializable.SerializableBiConsumer;
+import com.github.openjson.JSONArray;
+import com.github.openjson.JSONObject;
+
+/**
+ * {@link org.apache.wicket.ajax.form.OnChangeAjaxBehavior} that streams back to server properties
+ * of the selected file(s) (at client side), before uploading it (them).
+ *
+ * @author Ernesto Reinaldo Barreiro (reiern70@gmail.com).
+ */
+public abstract class FilesSelectedBehavior extends OnChangeAjaxBehavior {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final ResourceReference JS = new JavaScriptResourceReference(FilesSelectedBehavior.class, "FilesSelectedBehavior.js");
+
+    @Override
+    protected void onBind() {
+        super.onBind();
+        Component component = getComponent();
+        if (!(component instanceof FileUploadField)) {
+            throw new WicketRuntimeException("Behavior " + getClass().getName()
+                    + " can only be added to an instance of a FileUploadField");
+        }
+    }
+
+    @Override
+    protected void onUpdate(AjaxRequestTarget target) {
+        Request request = RequestCycle.get().getRequest();
+        List<FileDescription> fileDescriptions = new ArrayList<>();
+        IRequestParameters parameters = request.getRequestParameters();
+        // data is streamed as JSON.
+        StringValue fileInfos = parameters.getParameterValue("fileInfos");
+        JSONArray jsonArray = new JSONArray(fileInfos.toString());
+        for (int i = 0; i < jsonArray.length(); i++)
+        {
+            fileDescriptions.add(new FileDescription((JSONObject)jsonArray.get(i)));
+        }
+        onSelected(target, fileDescriptions);
+    }
+
+
+    /**
+     * Called when a file, at client side is selected.
+     *
+     * @param target           The {@link org.apache.wicket.ajax.AjaxRequestTarget}
+     * @param fileDescriptions A list of FileDescription
+     */
+    protected abstract void onSelected(AjaxRequestTarget target, List<FileDescription> fileDescriptions);
+
+    @Override
+    protected void updateAjaxAttributes(AjaxRequestAttributes attributes)
+    {
+        super.updateAjaxAttributes(attributes);
+        attributes.getAjaxCallListeners().add(new IAjaxCallListener()
+        {
+            @Override
+            public CharSequence getPrecondition(Component component)
+            {
+                return "return Wicket.FilesSelected.precondition(this);";
+            }
+        });
+        attributes.getDynamicExtraParameters().add("return Wicket.FilesSelected.collectFilesDetails('" + getComponent().getMarkupId() + "');");
+    }
+
+    @Override
+    public void renderHead(Component component, IHeaderResponse response)
+    {
+        super.renderHead(component, response);
+        response.render(JavaScriptHeaderItem.forReference(JS));
+    }
+
+    /**
+     * Creates an {@link FilesSelectedBehavior} based on lambda expressions
+     *
+     * @param select {@link SerializableBiConsumer}
+     *
+     * @return the {@link FilesSelectedBehavior} behavior
+     */
+    public static FilesSelectedBehavior onSelected(
+            SerializableBiConsumer<AjaxRequestTarget, List<FileDescription>> select)
+    {
+        Args.notNull(select, "select");
+
+        return new FilesSelectedBehavior()
+        {
+            private static final long serialVersionUID = 1L;
+
+            @Override
+            protected void onSelected(AjaxRequestTarget target, List<FileDescription> fileDescriptions) {
+                select.accept(target, fileDescriptions);
+            }
+        };
+    }
+}
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.js
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.js
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+;(function (undefined) {
+
+    'use strict';
+
+    if (typeof(Wicket.FilesSelected) === 'object') {
+        return;
+    }
+
+    /**
+     * Contains JS code for FilesSelectedBehavior.
+     */
+    Wicket.FilesSelected = {
+
+        // the id of temporary object holding files.
+        varId: function (componentId) {
+            return componentId + "_files";
+        },
+
+        /**
+         * Precondition to trigger (or not) server round-trip.
+         *
+         * @param inputField the file input field
+         * @returns {boolean} true if some files (file) were (was) selected,
+         */
+        precondition : function (inputField) {
+            if (inputField.files && inputField.files.length > 0) {
+                var id = this.varId(inputField.id);
+                Wicket.FilesSelected[id] = inputField.files;
+                return true;
+            }
+            return false;
+        },
+
+        /**
+         * Collects selected files details
+         *
+         * @param componentId The id of file upload input field.
+         *
+         * @returns array with file infos.
+         */
+        collectFilesDetails: function(componentId) {
+            var id = this.varId(componentId);
+            var sources = Wicket.FilesSelected[id];
+            var files = [];
+            for (var i = 0; i < sources.length; i++) {
+                var file = sources[i];
+                var info = {
+                    'fileName': file.name,
+                    'fileSize': file.size,
+                    'lastModified': file.lastModified,
+                    'mimeType': file.type
+                }
+                files.push(info);
+            }
+            // clean temporary holder object
+            delete(Wicket.FilesSelected[id]);
+            // return information about selected files.
+            return {'fileInfos': JSON.stringify(files)};
+        }
+    };
+})();
--- a/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/FileUploadPage$MultipleFileUploadsSamplePanel.html
+++ b/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/FileUploadPage$MultipleFileUploadsSamplePanel.html
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:wicket="http://wicket.apache.org">
+<head>
+    <wicket:head>
+        <script type="text/css">
+            table.files {
+                width: 1000px;
+                display: table;
+            }
+            table.files tbody {
+                width: 100%;
+            }
+            table.files td {
+                width: 25%;
+            }
+        </script>
+    </wicket:head>
+</head>
+<body>
+<wicket:panel>
+    <h2>Uploading a multiple files</h2>
+
+    <div wicket:id="feedback"></div>
+
+    <form wicket:id="form">
+        Text field: <input wicket:id="text" type="text"/><br/>
+        File field (can select multiple files): <input wicket:id="file" type="file" multiple="multiple"/> (<span wicket:id="max"></span> max)<br/><br/>
+        <br/>
+        <table wicket:id="selectedFileInfo" class="files">
+        </table>
+        <br/>
+        <div wicket:id="progress"></div>
+        <input wicket:id="submit" type="submit" value="Regular Submit"/> <input wicket:id="ajaxSubmit" type="button" value="Ajax Submit"/>
+    </form>
+</wicket:panel>
+</body>
+</html>
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileDescription.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileDescription.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.markup.html.form.upload;
+
+import java.io.Serializable;
+import java.util.Date;
+import java.util.Objects;
+import com.github.openjson.JSONObject;
+
+/**
+ * Description of file properties as in browser client side.
+ */
+public class FileDescription implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private final String fileName;
+    private final long fileSize;
+    private final Date lastModified;
+    private final String mimeType;
+
+   public FileDescription(JSONObject jsonObject) {
+        this(jsonObject.getString("fileName"), jsonObject.getLong("fileSize"),
+                jsonObject.getLong("lastModified"), jsonObject.getString("mimeType"));
+    }
+
+    public FileDescription(String fileName, long fileSize, long lastModified, String mimeType) {
+        this.fileName = fileName;
+        this.fileSize = fileSize;
+        this.lastModified = new Date(lastModified);
+        this.mimeType = mimeType;
+    }
+
+    public String getFileName() {
+        return fileName;
+    }
+
+    public long getFileSize() {
+        return fileSize;
+    }
+
+    public Date getLastModified() {
+        return lastModified;
+    }
+
+    public String getMimeType() {
+        return mimeType;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        FileDescription that = (FileDescription) o;
+        return fileName.equals(that.fileName);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(fileName);
+    }
+}
--- a/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/FileUploadPage$SingleFileUploadSamplePanel.html
+++ b/wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/FileUploadPage$SingleFileUploadSamplePanel.html
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:wicket="http://wicket.apache.org">
+<body>
+<wicket:panel>
+
+    <h2>Uploading a single file</h2>
+
+    <div wicket:id="feedback"></div>
+
+    <form wicket:id="form">
+        Text field: <input wicket:id="text" type="text"/><br/>
+        File field: <input wicket:id="file" type="file"/> (<span wicket:id="max"></span> max)<br/><br/>
+        <br/>
+        <div wicket:id="selectedFileInfo"></div>
+        <br/>
+        <div wicket:id="progress"></div>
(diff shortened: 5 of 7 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
```diff
@@ -47,6 +57,74 @@
...
+		/**
+		 * Called when a file, at client side is selected.
+		 *
+		 * @param target The {@link org.apache.wicket.ajax.AjaxRequestTarget}
+		 * @param fileName The client file name
+		 * @param fileSize The client file size
+		 * @param lastModified The {@link java.util.Date} when file was last modified.
+		 * @param mimeType The  MIME type of file.
+		 */
+		protected abstract void onFileSelected(AjaxRequestTarget target, String fileName, long fileSize,
```
martin-g, line 102: I think we should think about the `multiple` support now, because later this method signature may need to be changed to something like `protected abstract void onFilesSelected(AjaxRequestTarget target, List<FileDescription>)`

reiern70: Oh. I was thinking to roll a different behavior. But you are probably right. I will add support for multiple values ASAP

reiern70: My latest commit adds support for multiple files. I have tested it locally and it seems to work (with multiple files). Not sure if the way I did it is the best/correct one.

svenmeier:
IMHO OnFileSelectedBehavior deserves to be a top-level class. Maybe with its own JS resource, since the script to collect dynamicExtraParameters is getting very long now. 

With a static factory method users could use a lambda expression (like we have for several other behaviors, e.g. AjaxFormSubmitBehavior#onSubmit):

    file.add(OnFileSelectedBehavior.onSelected((AjaxRequestTarget target, List<OnFileSelectedBehavior.FileDescription> fileDescriptions) -> {
        // ...
    }));
	
Maybe drop the "On"-prefix too, since almost all current behaviors don't have it - are you doing Android development? ;)

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
```diff
@@ -47,6 +59,146 @@
...
+			List<FileDescription> fileDescriptions = new ArrayList<>();
+			IRequestParameters parameters = request.getRequestParameters();
+			String[] fileNames = parse(parameters.getParameterValue("fileName"));
+			String[] fileSizes = parse(parameters.getParameterValue("fileSize"));
+			String[] lastModifiedDates = parse(parameters.getParameterValue("lastModified"));
+			String[] mimeTypes = parse(parameters.getParameterValue("mimeType"));
+			for (int i = 0; i < fileNames.length; i++) {
+				FileDescription fileDescription = new FileDescription(
+						fileNames[i],
+						Long.parseLong(fileSizes[i]),
```
martin-g, line 145: Prefer `Long.parseLong(String s, int radix)` instead

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
```diff
@@ -47,6 +59,146 @@
...
+			return stringValue.toString().split(",");
+		}
+
+		/**
+		 * Called when a file, at client side is selected.
+		 *
+		 * @param target The {@link org.apache.wicket.ajax.AjaxRequestTarget}
+		 * @param fileDescriptions A list of FileDescription
+		 */
+		protected abstract void onFileSelected(AjaxRequestTarget target, List<FileDescription> fileDescriptions);
```
martin-g, line 163: s/onFileSelected/onSelect/ ? I think @svenmeier also suggested this name

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
```diff
@@ -47,6 +59,146 @@
...
+			});
+
+			String varName = "window."+getComponent().getMarkupId()+"_files";
+			attributes.getDynamicExtraParameters().add("var files = {};" +
+					"var fileNames = [];" +
+					"var fileSizes = [];" +
+					"var lastModifiedDates = [];" +
+					"var mimeTypes = [];" +
+					"for (var i = 0; i < "+varName+".length; i++) {" +
+							" f = " + varName + "[i];" +
```
martin-g, line 185: missing `var` for `f`

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
```diff
@@ -47,6 +59,146 @@
...
+							"lastModifiedDates.push(f.lastModified);" +
+							"mimeTypes.push(f.type);" +
+					"};" +
+					"var info = {" +
+							"'fileName': fileNames, " +
+							"'fileSize': fileSizes, " +
+							"'lastModified': lastModifiedDates, " +
+							"'mimeType': mimeTypes, " +
+							"};"
+					+ "console.log(info);"
```
martin-g, line 197: Debug leftover

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
```diff
@@ -47,6 +59,146 @@
...
+
+		@Override
+		protected void updateAjaxAttributes(AjaxRequestAttributes attributes) {
+			super.updateAjaxAttributes(attributes);
+			attributes.getAjaxCallListeners().add(new IAjaxCallListener()
+			{
+				@Override
+				public CharSequence getPrecondition(Component component)
+				{
+					return "if (this.files) {  window." + component.getMarkupId() + "_files = this.files; return true; } " +
```
martin-g, line 173: Let's put the new property to `window.Wicket` namespace instead of to the global `window` namespace.

martin-g (approved):

reiern70:
@svenmeier and @martin-g 

Thanks for your comments:

> IMHO OnFileSelectedBehavior deserves to be a top-level class.

 My rationale was the following: this class will only be used in relation with FileUploadField, why then pollute package? Of course, this is a matter of taste and consistency with project practices/conventions. Moving it top-level class.

> Maybe with its own JS resource, since the script to collect dynamicExtraParameters is getting very long now.

True.
 
> With a static factory method users could use a lambda expression (like we have for several other behaviors, e.g. AjaxFormSubmitBehavior#onSubmit):
> 
> ```
> file.add(OnFileSelectedBehavior.onSelected((AjaxRequestTarget target, List<OnFileSelectedBehavior.FileDescription> fileDescriptions) -> {
>     // ...
> }));
> ```
> 
> Maybe drop the "On"-prefix too, since almost all current behaviors don't have it - are you doing Android development? ;)

Never did Android development in my life. If you both agree this should be done I will gladly do it ;-)

reiern70:
**Note:** A just thought that a possible addition in case of multiple file uploads is a component with a table displaying all the selected files. WDYT?

reiern70:
> IMHO OnFileSelectedBehavior deserves to be a top-level class. Maybe with its own JS resource, since the script to collect dynamicExtraParameters is getting very long now.
> 
> With a static factory method users could use a lambda expression (like we have for several other behaviors, e.g. AjaxFormSubmitBehavior#onSubmit):
> 
> ```
> file.add(OnFileSelectedBehavior.onSelected((AjaxRequestTarget target, List<OnFileSelectedBehavior.FileDescription> fileDescriptions) -> {
>     // ...
> }));
> ```
> 
> Maybe drop the "On"-prefix too, since almost all current behaviors don't have it - are you doing Android development? ;)

image

I see a lot of 'On' hanging around. Thus unless there is a very strong opinion about it from most of the developers I will keep it.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
reiern70: ok.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
reiern70: Also I though about cleaning garbage after use it: this only needs to be kept in context of this AJAX request. Maybe we can even use attrs object as a holder. I will check.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
```diff
@@ -47,6 +59,146 @@
...
+			}
+		}
+
+		@Override
+		protected void onUpdate(AjaxRequestTarget target)
+		{
+			Request request = RequestCycle.get().getRequest();
+			List<FileDescription> fileDescriptions = new ArrayList<>();
+			IRequestParameters parameters = request.getRequestParameters();
+			String[] fileNames = parse(parameters.getParameterValue("fileName"));
```
reiern70, line 138: To myself: can we use getParameterValues instead?

reiern70:
@svenmeier and @martin-g 

Thanks for your feedback and useful comments. I have reworked PR taking into account your comments. Can you please review again?

reiern70:
The new section on examples looks like 

image

reiern70:
> @svenmeier and @martin-g
> 
> Thanks for your comments:
> 
> > IMHO OnFileSelectedBehavior deserves to be a top-level class.
> 
> My rationale was the following: this class will only be used in relation with FileUploadField, why then pollute package? Of course, this is a matter of taste and consistency with project practices/conventions. Moving it top-level class.
> 
> > Maybe with its own JS resource, since the script to collect dynamicExtraParameters is getting very long now.
> 
> True.
> 
> > With a static factory method users could use a lambda expression (like we have for several other behaviors, e.g. AjaxFormSubmitBehavior#onSubmit):
> > ```
> > file.add(OnFileSelectedBehavior.onSelected((AjaxRequestTarget target, List<OnFileSelectedBehavior.FileDescription> fileDescriptions) -> {
> >     // ...
> > }));
> > ```
> > 
> > 
> > Maybe drop the "On"-prefix too, since almost all current behaviors don't have it - are you doing Android development? ;)
> 
> Never did Android development in my life. If you both agree this should be done I will gladly do it ;-)

@svenmeier My latest commit introduces factory method as  what you suggested

image

thanks for pointing this out.

reiern70:
My IDE suggest me to use let instead of var

image

I was using let, reverting to use var otherwise build fails.

reiern70:
once reviewed and approved I will rebase on top of master and squash commits.

reiern70:
rebased on top oof latest master + squashed commits. Waiting for ok to merge )or do more changes if asked)

martin-g:
> **Note:** A just thought that a possible addition in case of multiple file uploads is a component with a table displaying all the selected files. WDYT?

I don't think we should add such component. Every application may want to display it in a different way.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
```diff
@@ -21,12 +21,21 @@
...
+import org.apache.wicket.ajax.form.OnChangeAjaxBehavior;
 import org.apache.wicket.markup.ComponentTag;
 import org.apache.wicket.markup.html.form.FormComponent;
 import org.apache.wicket.model.IModel;
 import org.apache.wicket.protocol.http.IMultipartWebRequest;
+import org.apache.wicket.request.IRequestParameters;
 import org.apache.wicket.request.Request;
+import org.apache.wicket.request.cycle.RequestCycle;
 import org.apache.wicket.util.convert.ConversionException;
+import org.apache.wicket.util.string.StringValue;
```
martin-g, line 38: It seems these imports are no more used.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
```diff
@@ -0,0 +1,134 @@
...
+import org.danekja.java.util.function.serializable.SerializableBiConsumer;
+import org.danekja.java.util.function.serializable.SerializableConsumer;
+import com.github.openjson.JSONArray;
+import com.github.openjson.JSONObject;
+import com.github.openjson.JSONTokener;
+
+/**
+ * {@link org.apache.wicket.ajax.form.OnChangeAjaxBehavior} that streams back to server properties
+ * of the selected file (at client side), even when file has not yet being uploaded.
+ * s	 *
```
martin-g, line 47: this line should be removed

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.wicket.util.string.StringValue;
+import org.danekja.java.util.function.serializable.SerializableBiConsumer;
+import org.danekja.java.util.function.serializable.SerializableConsumer;
+import com.github.openjson.JSONArray;
+import com.github.openjson.JSONObject;
+import com.github.openjson.JSONTokener;
+
+/**
+ * {@link org.apache.wicket.ajax.form.OnChangeAjaxBehavior} that streams back to server properties
+ * of the selected file (at client side), even when file has not yet being uploaded.
```
martin-g, line 46: ... selected file`(s)` before uploading it/them.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
```diff
@@ -0,0 +1,134 @@
...
+                return "return Wicket.OnFilesSelected.precondition(this, '" + component.getMarkupId() + "');";
+            }
+        });
+        attributes.getDynamicExtraParameters().add("return Wicket.OnFilesSelected.collectFilesDetails('" + getComponent().getMarkupId() + "');");
+    }
+
+    @Override
+    public void renderHead(Component component, IHeaderResponse response)
+    {
+        response.render(JavaScriptHeaderItem.forReference(JS));
```
martin-g, line 108: Any reason `super.**` to be called after ? I don't think there is any difference which one is first in this case. It just looks suspicious this way.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Creates an {@link OnFilesSelectedBehavior} based on lambda expressions
+     *
+     * @param select {@link SerializableBiConsumer}
+     *
+     * @return the {@link org.apache.wicket.markup.html.form.upload.OnFilesSelectedBehavior} behavior
+     */
+    public static OnFilesSelectedBehavior select(
+            SerializableBiConsumer<AjaxRequestTarget, List<FileDescription>> select)
```
martin-g, line 120: For consistency with the other similar methods the method name should be `onSelect`
https://github.com/apache/wicket/blob/048e5681df960018722237ee6254273e839b5fd2/wicket-core/src/main/java/org/apache/wicket/ajax/AjaxEventBehavior.java#L166

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.js
```diff
@@ -0,0 +1,79 @@
...
+     */
+    Wicket.OnFilesSelected = {
+
+        // the id of temporary object holding files.
+        varId: function (componentId) {
+            return componentId + "_files";
+        },
+
+        /**
+         * Precondition to trigger (oor not) server round-trip.
```
martin-g, line 37: s/oor/or/

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.js
```diff
@@ -0,0 +1,79 @@
...
+         *
+         * @param component
```
martin-g, line 39: what is the type of `component` here ? 
This parameter name is a little confusing. It sounds like a Wicket Component, but it is something else.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.js
```diff
@@ -0,0 +1,79 @@
...
+         * @param componentId
+         * @returns {boolean} true if some files (file) were (was) selected,
+         */
+        precondition : function (component, componentId) {
+            var id = this.varId(componentId);
+            if (component.files && component.files.length > 0) {
+                window[id] = component.files;
```
martin-g, line 46: Can we use `Wicket.OnFilesSelected[id]` instead of `window[id]` ?
We should try to not pollute the global namespace, even temporarily.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.js
```diff
@@ -0,0 +1,79 @@
...
+                return true;
+            }
+            return false;
+        },
+
+        /**
+         * Collects selected files IDs
```
martin-g, line 53: s/IDs/details/ ?

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/FileUploadPage$MultipleFileUploadsSamplePanel.html
```diff
@@ -0,0 +1,37 @@
...
+                width: 1000px;
+                display: table;
+            }
+            table.files tbody {
+                width: 100%;
+            }
+            table.files td {
+                width: 25%;
+            }
+        </script>s
```
martin-g, line 16: dangling `s` at the end of the line

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/EffectsPage.html
```diff
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <html xmlns:wicket="http://wicket.apache.org">
+<head>
```
martin-g, line 3: Why EffectsPage.html is modified ?

martin-g (changes requested):

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
reiern70: Hum... Maybe I understood not correctly what @svenmeier was asking?

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/EffectsPage.html
reiern70: Hm... no idea. Reverting back. Thanks for spotting this.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
martin-g: @svenmeier said "file.add(OnFileSelectedBehavior.onSelected"
I am fine with either `onSelect` and `onSelected` but I like `onSelect` better, it is more consistent with the rest similar methods.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
reiern70: No particular reason I think. I will fix.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
reiern70: Thanks.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
reiern70: image

?

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
reiern70: Fixed.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
reiern70: fixed.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
martin-g: Here, I think, he meant the class name, not the method name,

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
reiern70: done.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.js
reiern70: Yes. I think we can.

wicket-examples/src/main/java/org/apache/wicket/examples/ajax/builtin/EffectsPage.html
reiern70: Fixed.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.js
reiern70: Fixed.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.js
reiern70: fixed

reiern70:
@martin-g Many thanks for review and feedback. I have addressed your comments.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.js
reiern70: Fixed this too

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/OnFilesSelectedBehavior.java
reiern70: renamed behavior to FilesSelectedBehavior and factory method to  onSelected.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FileUploadField.java
reiern70: Fixed too.

reiern70:
> > **Note:** A just thought that a possible addition in case of multiple file uploads is a component with a table displaying all the selected files. WDYT?
> 
> I don't think we should add such component. Every application may want to display it in a different way.

True. I have added some example on how to do that on ajax-upload page

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.js
```diff
@@ -0,0 +1,79 @@
...
+
+        /**
+         * Precondition to trigger (or not) server round-trip.
+         *
+         * @param inputField the file input field
+         * @returns {boolean} true if some files (file) were (was) selected,
+         */
+        precondition : function (inputField) {
+            var id = this.varId(inputField.id);
+            console.log(id)
```
martin-g, line 44: debug leftover

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.js
```diff
@@ -0,0 +1,79 @@
...
+        },
+
+        /**
+         * Precondition to trigger (or not) server round-trip.
+         *
+         * @param inputField the file input field
+         * @returns {boolean} true if some files (file) were (was) selected,
+         */
+        precondition : function (inputField) {
+            var id = this.varId(inputField.id);
```
martin-g, line 43: you can move this line inside the `if` below

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.java
```diff
@@ -0,0 +1,130 @@
...
+        super.updateAjaxAttributes(attributes);
+        attributes.getAjaxCallListeners().add(new IAjaxCallListener()
+        {
+            @Override
+            public CharSequence getPrecondition(Component component)
+            {
+                return "return Wicket.FilesSelected.precondition(this);";
+            }
+        });
+        attributes.getDynamicExtraParameters().add("return Wicket.FilesSelected.collectFilesDetails('" + getComponent().getMarkupId() + "');");
```
martin-g, line 98: For simplification you can use again JavaScript's `this` as in the precondition. And in the JS code : `inputField.id` to get the id.

martin-g (approved):

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.java
reiern70: Hum... I was not sure this would work: i.e. that in this case this will point to inputField, this is the reason I didn't try that. Giving it a try

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.js
reiern70: Do we have a Wicket.log? On our private project I have added a log function what can be enabled via some flag you can change at browser console (to switch on debugging).

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.js
reiern70: done.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.js
martin-g: Yes - https://github.com/apache/wicket/blob/048e5681df960018722237ee6254273e839b5fd2/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js#L189-L222
It could be enabled/disabled via DebugSettings#ajaxDebugModeEnabled

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.java
martin-g: I might be wrong but I liked how you did it in `precondition()`!

reiern70: This does not seems to refer to inputField in this case. I could use attrs.c as it is the ID but I do not see and extra benefit in this. Keeping as it is.

reiern70:
@svenmeier and @martin-g Many thanks for reviews. Rebased and squashed commits. Merging now.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/FilesSelectedBehavior.js
reiern70: I see... Thanks for info. I see you can maybe also type Wicket.Log.enable=true; in console to enable it for some page.

martin-g: Yes! It will survive until the next page reload.

PR #473, opened 2021-06-16
WICKET-6895: Links to examples in documentation points to old version of 8x

jlindman:
Updated url to wicket 9 examples in asciidoc document (was pointing to 8.x examples).


bitstorm:
Solved https://issues.apache.org/jira/browse/WICKET-6895

PR #474, opened 2021-07-01
Java 9+ module-aware Wicket 9.x 

martin-g:
This PR makes Wicket 9.x fully (?!) supporting Java 9+ module system.

I decided to use 9.x instead of 10.x because this way it will be easier to test it with our production applications.
If there is a stopper then we will apply these changes only to 10.x. But the module-aware Wicket is backward compatible then we could merge it for 9.5.0.

At the moment the only issue I am aware of is with clirr-maven-plugin:

```
[ERROR] Failed to execute goal org.codehaus.mojo:clirr-maven-plugin:2.8:check (clirr-check) on project wicket-util: Execution clirr-check of goal org.codehaus.mojo:clirr-maven-plugin:2.8:check failed: Invalid byte tag in constant pool: 19 -> [Help 1]
```
So you need to build with `mvn ... -Dclirr.skip=true`

I will contact Clirr project to see whether a new version could be released.

Another (smaller) issue is that few dependencies do not provide proper Java modules (e.g. javax.servlet-api, cglib, velocty, commons-fileupload, ...). I am not sure whether this is a deal-breaker or not.

I'll be thankful if you could test this branch and report issues!


martin-g:
TODO JPMS-fy the http2 modules. I decided to do it in separate commit so that it is easier to cherry-pick in 10.x where those were merged to -core.

bitstorm:
I see you did also some code clean-up (for util classes for example). Good!

martin-g:
Not really a cleanup.
I needed to move some test classes to unique packages.
We have a test for this in testing/common-testts but it covers just the
main sources.
JPMS complained that some packages are in more than one modules.
So I moved **.util.* to **.core.util.*

pom.xml
```diff
@@ -142,6 +142,7 @@
 		<commons-collections.version>3.2.2</commons-collections.version>
 		<commons-collections4.version>4.4</commons-collections4.version>
 		<commons-fileupload.version>1.4</commons-fileupload.version>
+<!--		<commons-fileupload.version>0.68.0.1</commons-fileupload.version>-->
```
solomax, line 145: I guess this one can be removed

pom.xml
```diff
@@ -169,7 +170,7 @@
 		<mockito.version>3.11.1</mockito.version>
 		<objenesis.version>3.2</objenesis.version>
 		<openjson.version>1.0.12</openjson.version>
-		<slf4j.version>1.7.25</slf4j.version>
+		<slf4j.version>2.0.0-alpha1</slf4j.version>
```
solomax, line 173: Why to update to `alpha`?

pom.xml
```diff
@@ -290,6 +290,7 @@
 				<version>${commons-fileupload.version}</version>
 				<exclusions>
 					<exclusion>
+<!--						<groupId>com.jwebmp.jpms.commons</groupId>-->
```
solomax, line 293: I guess this one can be dropped

solomax (approved):
Great job! :)

pom.xml
martin-g: Done!

pom.xml
martin-g: Because SLF4J added support for JPMS in 2.x.
They just released alpha2. 
Hopefully they will release 2.0.0 soon! But SLF4J project is not very active lately.

solomax: they had alphas and betas only for very long time :(
let's see :))

bitstorm (approved):

martin-g:
I've tested this branch in a real application both with Tomcat 9.0.50 and Jetty 10.0.6. 
I didn't find any issue with the application related to JPMS.

My only concern before merging the branch to `wicket-9.x` is the usage of slf4j 2.0.0-alpha2. This might scare some people.

svenmeier (approved):
+1 works flawless here

PR #478, opened 2021-09-21
WICKET-6921 Avoid updating hidden forms

mat128:
Currently, MultipartFormComponentListener ensures form enctype is
correct on the client-side when visibility for multipart-enabling
components is toggled. It does so by visiting all form components,
detecting multipart-enabling components and emitting javascript code
to refresh the encoding type on the client-side.

There are a few issues solved by this commit:

- The identified form component could be part of a hidden hierarchy.
  This prevents finding it in the DOM, and using it's form attribute
  to set the form enctype.

- There could be more than one form with multipart-enabling components
  which would require an enctype refresh on the client-side.

This commit modifies MultipartFormComponentListener to find all forms,
ensure they are visible in the hierarchy, then visiting all form
components to find multipart-enabling. A single statement is emitted
per form.


martin-g:
Thank you, @mat128 !

PR #481, opened 2021-11-14
several minor datatable improvements

reiern70:


wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigationToolbar.java
```diff
@@ -44,17 +46,18 @@ public NavigationToolbar(final DataTable<?, ?> table)
...
 		span.add(newPagingNavigator("navigator", table));
-		span.add(newNavigatorLabel("navigatorLabel", table));
+		Component complexLabel = newComplexNavigatorLabel("navigatorLabel", table);
+		if (complexLabel != null)
+		{
+			span.add(complexLabel);
+		}
+		else
+		{
+			span.add(newNavigatorLabel("navigatorLabel", table));
```
solomax, line 59: the `null` will be added here in case `newNavigatorLabel` is not overwritten and `complexLabel == null`

reiern70: Bot following you. We don't want to break API. New method will allows you to replace simple label with a come complex thing: this is not possible now because newNavigatorLabel  factory method returns a WebComponent and not a component. By default new method returns null. Thus existing applications will not be broken.

martin-g: I think it is backward compatible to change the return type from `X` to `AParentOfX`.
Try it! The `japicmp` plugin will tell you!

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigatorLabel.java
```diff
@@ -46,11 +46,21 @@
...
 			new Model<>(new LabelModelObject(pageable))));
 	}
 
-	private static class LabelModelObject implements IClusterable
+	/**
+	 * Allows to override the key used for navigation label.
+	 *
+	 * @return The i18n key.
+	 */
+	protected String getNavigationLabelKey()
```
martin-g, line 58: What is the problem you try to solve here ?
You know that you could be specific in your i18n files to use the component path in the tree to specify different labels for different usages ?

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigationToolbar.java
reiern70: @martin-g I did. It was the first thing I tried and the plugin complained. Thus I tried this.

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigatorLabel.java
reiern70: I know all of the above. 1) we use a a single i18n file at APP level (not my decision) 2) path is error prone and this is part of a component used in many places in app.

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigationToolbar.java
reiern70: @martin-g for wicket 10.x maybe we can change return type for 9.x plugin will complain and project does not build.

martin-g: That's unfortunate! It is source compatible but not binary compatible.
Let's see what others think about it. 
I don't like the change but I see no other way to do it.
And we have the same problem for org.apache.wicket.extensions.ajax.markup.html.AjaxEditableChoiceLabel#newLabel()

reiern70: > That's unfortunate! It is source compatible but not binary compatible. Let's see what others think about it. I don't like the change but I see no other way to do it. And we have the same problem for org.apache.wicket.extensions.ajax.markup.html.AjaxEditableChoiceLabel#newLabel()

Ma neither. I first try to do as you suggested + mvn clean build and I build failed. Thus I had too opt for this HACKY ugly thing. :-( In wicket 10.x we can drop the new label. The use case here is I want to plug there a panel with pagination label + some text filter. TO do this I had to roll my own NavigationToolbar... not nice at all

martin-g: I understand.
But IMO we should not promote the new method in 9.x. I.e. we should not deprecate the old method.
Here is what I think would be the best:
1) Wicket 9.x - add the new method and use it. Just as in the PR but remove the deprecations. Also make it clear in the javadoc of the new method that it is there temporarily for 9.x
2) Wicket 10.x - just change the return type to `Component`

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigatorLabel.java
martin-g: What bothers me here is that we don't do this anywhere else. It just feels inconsistent.

reiern70: I can revert this if we do the model object for navigation class public.

reiern70: reverted this change. It is enough for my use case to have **LabelModelObject** public instead of private.

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigationToolbar.java
reiern70: I have done this for this branch

reiern70: I will create PR for wicket 10.x

martin-g (approved):
Please create a JIRA ticket

reiern70:
> Please create a JIRA ticket

https://issues.apache.org/jira/browse/WICKET-6935

reiern70:
@martin-g squashed all commits into one

PR #482, opened 2021-11-18
[WICKET-6935] several minor datatable improvements

reiern70:
Corresponding changes to 

https://github.com/apache/wicket/pull/481

But for mater. As agreed here we just change return type of factory method for label.

### DIFF
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigationToolbar.java
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigationToolbar.java
@@ -17,8 +17,10 @@
 package org.apache.wicket.extensions.markup.html.repeater.data.table;
 
 import org.apache.wicket.AttributeModifier;
+import org.apache.wicket.Component;
 import org.apache.wicket.markup.html.WebComponent;
 import org.apache.wicket.markup.html.WebMarkupContainer;
+import org.apache.wicket.markup.html.navigation.paging.IPageableItems;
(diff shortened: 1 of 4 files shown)


wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigationToolbar.java
```diff
@@ -73,16 +68,16 @@ protected PagingNavigator newPagingNavigator(final String navigatorId,
...
 	 * @param navigatorId
 	 *            component id navigator label should be created with
 	 * @param table
-	 *            dataview used by datatable
+	 *            DataTable used by datatable
 	 * @return navigator label that will be used to navigate the data table
 	 * 
 	 */
-	protected WebComponent newNavigatorLabel(final String navigatorId, final DataTable<?, ?> table)
+	protected Component newNavigatorLabel(final String navigatorId, final DataTable<?, ?> table)
```
martin-g, line 80: Please do the same for `org.apache.wicket.extensions.ajax.markup.html.AjaxEditableChoiceLabel#newLabel()`

martin-g (approved):

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/NavigationToolbar.java
reiern70: Done

reiern70:
@martin-g many thanks for review. Landed in master

PR #485, opened 2021-11-22
Fix spelling in user guide

froque:
- [ ] This should be backported to versions 7, 8 and 9


reiern70 (approved):
LGTM

martin-g:
Downported it to 8.x and 9.x. It does not apply for 7.x!

PR #488, opened 2021-12-23
[WICKET-6943] Save info about Locale direction LTR/RTL to Session metadata

solomax:
I would like to introduce the way to determine Locale direction LTR/RTL
this is useful to set some specific attributes
And to load RTL specific JS/CSS

will be useful for ex. here: https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/942 :)))

UPD: i'll create JIRA if this PR will pass the review :)


svenmeier:
I see that the value is derived from the session's locale. What is the benefit of caching it on the request? Is this a performance optimization?
If Wicket itself doesn't use this function, I don't see much reasons to add this.

solomax:
Thanks for the review @svenmeier :)

`IS_RTL` is saved to session metadata as soon as Locale for the session is set
Maybe it would be possible to create sort of extension point

Benefit of caching `RTL-state`  per session is not to check it for every request
To support RTL in `wicket-bootstrap` I need to add special JS file to every page

svenmeier:
Ah, the comment says `/** records if pages have been unlocked for the current request */`
I've just read the last two words and assumed you want to keep that in the request metadata.

I still don't see why we need to cache this separately from the locale.

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -126,6 +127,11 @@
 	{
 		private static final long serialVersionUID = 1L;
 	};
+	/** records if pages have been unlocked for the current request */
```
martin-g, line 130: this javadoc is wrong. it seems to be a copy from somewhere else

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -126,6 +127,11 @@
...
+	public static final MetaDataKey<Boolean> IS_RTL = new MetaDataKey<>()
```
martin-g, line 131: My suggestion in the Wicket-Bootstrap issue to use the metadata was good if this functionality was in W-B repo, because there there is no way to add a new field.
If we are going to add this functionality to wicket-core then we could use a normal field.

martin-g: In addition to using a field we should also add a non-final getter method. This way an application can define its own logic for deciding whether to use RTL or not.

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+package org.apache.wicket.core.util.lang;
+
+import java.util.Locale;
+import java.util.regex.Pattern;
+
+import org.apache.wicket.util.lang.Args;
+
+public class LocaleUtils {
```
martin-g, line 24: IMO, this class could be inlined in Session.

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * taken from BidiUtils
+	 *
+	 * A regular expression for matching right-to-left language codes. See
+	 * {@link #isRtlLanguage} for the design.
+	 */
+	private static final Pattern RtlLocalesRe = Pattern.compile("^(ar|dv|he|iw|fa|nqo|ps|sd|ug|ur|yi|.*[-_](Arab|Hebr|Thaa|Nkoo|Tfng))"
```
martin-g, line 31: The constant name should follow Java lang conventions, i.e. all upper-case - `RTL_LOCALES_RE`

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
```diff
@@ -0,0 +1,67 @@
...
+					+ "(?!.*[-_](Latn|Cyrl)($|-|_))($|-|_)");
+
+	private LocaleUtils() {
+		// utility class
+	}
+
+	/**
+	 * Check if a BCP 47 / III language code indicates an RTL language, i.e.
+	 * either: - a language code explicitly specifying one of the right-to-left
+	 * scripts, e.g. "az-Arab", or
```
martin-g, line 41: a`r`-Arab ?!

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
```diff
@@ -0,0 +1,67 @@
...
+	 * <p>
+	 * - a language code specifying one of the languages normally written in a
+	 * right-to-left script, e.g. "fa" (Farsi), except ones explicitly
+	 * specifying Latin or Cyrillic script (which are the usual LTR
+	 * alternatives).
+	 * <p>
+	 * The list of right-to-left scripts appears in the 100-199 range in
+	 * http://www.unicode.org/iso15924/iso15924-num.html, of which Arabic and
```
martin-g, line 49: use `<a href="...">`

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
```diff
@@ -0,0 +1,67 @@
...
+	 * Hebrew are by far the most widely used. We also recognize Thaana, N'Ko,
+	 * and Tifinagh, which also have significant modern usage. The rest (Syriac,
+	 * Samaritan, Mandaic, etc.) seem to have extremely limited or no modern
+	 * usage and are not recognized. The languages usually written in a
+	 * right-to-left script are taken as those with Suppress-Script:
+	 * Hebr|Arab|Thaa|Nkoo|Tfng in
+	 * http://www.iana.org/assignments/language-subtag-registry, as well as
```
martin-g, line 56: `<a href="...">`

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
martin-g: I guess `az-Arab` should be positive, since the `-Arab` part is matching.
Unit tests will help maintain this Regex in the future!

wicket-core/src/main/java/org/apache/wicket/Session.java
solomax: My bad, will fix ASAP :)

martin-g:
I find this useful and I'm +1 to add it to wicket-core! But we have to create a JIRA ticket for it for the changelog!

wicket-core/src/main/java/org/apache/wicket/Session.java
solomax: should be fixed

wicket-core/src/main/java/org/apache/wicket/Session.java
solomax: done

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
solomax: As I understand the comment it is correct:
The locale is RTL IF
- suffix states it is RTL: `az-Arab` -> `az` == `LTR` but `Arab` == `RTL` so everything is `RTL`
- the language is `RTL` for ex. `fa`
-

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
solomax: done

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
solomax: done

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
solomax: removed

wicket-core/src/main/java/org/apache/wicket/core/util/lang/LocaleUtils.java
solomax: done

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -130,6 +132,15 @@
 	/** Name of session attribute under which this session is stored */
 	public static final String SESSION_ATTRIBUTE_NAME = "session";
 
+	/**
+	 * taken from BidiUtils
```
martin-g, line 136: Maybe expand with "Apache Commons XYZ BidiUtils" or even a link to the Git repository ?

wicket-core/src/main/java/org/apache/wicket/Session.java
```diff
@@ -184,6 +195,36 @@ public static Session get()
...
+	 * right-to-left script are taken as those with 
+	 * <a href="http://www.iana.org/assignments/language-subtag-registry">Suppress-Script</a>:
+	 * Hebr|Arab|Thaa|Nkoo|Tfng, as well as
+	 * Sindhi (sd) and Uyghur (ug). The presence of other subtags of the
+	 * language code, e.g. regions like EG (Egypt), is ignored.
+	 *
+	 * @param locale - locale to check
+	 * @return <code>true</code> in case passed locale is right-to-left
+	 */
+	public static boolean isRtlLanguage(final Locale locale) {
```
martin-g, line 223: Any reason this method to be `public` ?
IMO we should start with `private` until there is a good reason to be part of the API

martin-g (commented):
Maybe we should replace `RTL` with `right-to-left` in the javadocs to be more clear.
Also several times it is said `locale is RTL`. IMO it is more correct to say that the locale's language is RTL

martin-g:
Also, basic unit tests are still missing.

wicket-core/src/main/java/org/apache/wicket/Session.java
solomax: I'd love to, but the sources seems to be unavailable :(

This SO article https://stackoverflow.com/questions/4192382/is-there-any-way-to-detect-an-rtl-language-in-java
has link to googlecode which is dead :(

I'll change the wording to be `Google Closure Templates BidiUtils` :)

solomax: done

wicket-core/src/main/java/org/apache/wicket/Session.java
solomax: method made private

martin-g (approved):

wicket-core/src/main/java/org/apache/wicket/Session.java
solomax: @martin-g can this method be made public? We are using it for utility purposes :)

martin-g: OK!

PR #489, opened 2021-12-27
Remove Apache collections4 as a runtime dependency

asvanberg:
For such a huge dependency it was barely used and in many instances was completely unnecessary. The biggest change would be how `MarkupContainer` children are stored.

### DIFF
--- a/wicket-core/pom.xml
+++ b/wicket-core/pom.xml
@@ -65,6 +65,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-collections4</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-lang3</artifactId>
 			<scope>test</scope>
 		</dependency>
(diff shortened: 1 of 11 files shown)


papegaaij:
This change will impose a huge performance impact on removing children. The complexity will change from O(1) to O(n) for a removal. This was the very reason why we used the LinkedMap from commons collections: it maintains a double linked list

martin-g:
I agree with @papegaaij that the new implementation of `previousKey()` will be slower and this will affect applications with big component trees.
Many of the other suggested changes could be made! But since we will have to keep collections4 in the classpath I wonder whether there is a need.

In the past Wicket (versions 1.0/1.1) have solved such problems by copying classes like `LinkedMap` in wicket-util. But this may lead to copying many classes - all directly used by LinkedMap.

asvanberg:
Are there any benchmarks that can be run to measure the performance difference? Or do you have a scenario in mind with "big component trees" where a lot of component removals are happening?

papegaaij:
Unfortunately, we don't have any benchmarks in place. I once tried to get JMH working, but it gave a lot of issues in Eclipse. In this case, the scenario isn't that complicated: create a component with about 10000 children and remove them again. You'll notice the difference in performance. Removing all components is not a very common operation, but removing (or replacing) a few is. It's very unfortunate that LinkedHashMap does not expose methods for bidirectional traversal. It does maintain a double linked list, but the 'before' reference is internal.

martin-g:
Closing as "Won't do" due to the reasons explained above.

PR #491, opened 2021-12-30
WICKET-6941 Add visible attribute to AbstractColumn to hide/show columns in DataTable/DataGridView

RLStokes:
Not sure about the change to the interface ICellPopulator whether it is allowed in a minor version update.


wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/grid/ICellPopulator.java
```diff
@@ -71,4 +71,8 @@
 	 */
 	void populateItem(final Item<ICellPopulator<T>> cellItem, final String componentId,
 		final IModel<T> rowModel);
+		
+	boolean isVisible();
```
reiern70, line 75: I would prefer this to be be 

```
default boolean isVisible() {
    return true;
}

```

and no setter.

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/grid/ICellPopulator.java
```diff
@@ -71,4 +71,8 @@
...
+	
+	void setVisible(boolean visible);
```
reiern70, line 77: As said I would prefer this to be only read-only

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/grid/PropertyPopulator.java
```diff
@@ -74,4 +75,16 @@ public void populateItem(final Item<ICellPopulator<T>> cellItem, final String co
 	{
 		cellItem.add(new Label(componentId, new PropertyModel<>(rowModel, property)));
 	}
+
+	@Override
+	public boolean isVisible()
```
reiern70, line 80: See my comment above.

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/AbstractColumn.java
```diff
@@ -37,6 +37,8 @@
 
 	private final S sortProperty;
 
+	private boolean visible = true;
```
reiern70, line 40: See comment about read only

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/table/HeadersToolbar.java
```diff
@@ -71,7 +71,10 @@
 
 				for (IColumn<T, S> column : table.getColumns())
 				{
-					columnsModels.add(Model.of(column));
+					if (column.isVisible())
```
reiern70, line 74: I'm not sure this is the only involved toolbar that needs to be fixed... Did you run

`mvn clean install`

?

reiern70 (commented):
Not sure I understand the use case for this. A table which dynamically can  have  more or less columns? I  would not mind this change is code is made read only with a default method in interface.

wicket-extensions/src/main/java/org/apache/wicket/extensions/markup/html/repeater/data/grid/ICellPopulator.java
reiern70: Also JavaDoc.

reiern70:
And thanks for your work :-)

martin-g:
Close/reopen to re-trigger the build checks.

PR #492, opened 2022-01-17
WICKET-6945 Allow onConfigure to hide forms

mat128:
Currently, MultipartFormComponentListener evaluates form visibility
before the configuration phase. This prevents code in onConfigure from
modifying the visibility of forms.

This ensures the listener evaluates the visibility during
onAfterRespond, which happens after the configuration phase.

The existing test case was modified to ensure modifications to the
visibility during the onConfigure phase is properly evaluated.

Closes: WICKET-6945


martin-g (approved):

martin-g:
Thank you, @mat128 !

PR #493, opened 2022-01-19
WICKET-6946 Document limitations of overriding isVisible/isEnabled

mat128:
Document limitations of overriding isVisible/isEnabled

This PR also improve various sections around the edited paragraphs. I can split them in different commits if need be.

Also, I'm aware of the whole override vs non-override debate and this PR only serves as documenting (some of) the pitfalls of doing so.


svenmeier:
Many thanks, that's a welcomed improvement.

bitstorm (approved):

bitstorm:
Merged

PR #497, opened 2022-02-01
WICKET-6952 Performance improvements for `Strings.isEmpty()`

theigl:
This PR adds some micro-improvements for `Strings.isEmpty`.

`String.trim` is one of the hottest methods in my application. Most calls come from `Strings.isEmpty`, either called directly or via `Args.notEmpty`. It is called thousands of times during each request, so even minor improvements should make sense.

This PR applies two optimizations:

1. It adds overloaded methods for `Strings.isEmpty` and `Args.notEmpty` that take a `String` argument. The compiler seems to better inline the method when the exact parameter type is known at compile time. This results in about 5% performance improvement in most of my tests.
2. Before calling `String.trim`, the first character of the string is checked for whitespace. If it isn't whitespace, we can avoid calling `String.trim`. This optimization makes sense for the new method, but *especially* for the existing method with CharSequence parameter. One CharSequence frequently passed to `isEmpty` is `AppendingStringBuffer`. Before calling `trim` we have to convert it into a string with `toString`. In the case of `AppendingStringBuffer`, this *always* allocates a new string with the entire contents of the buffer.

```
Benchmark                            Mode  Cnt          Score         Error  Units
StringsBenchmark.bufferFastIsEmpty  thrpt   15    4706409.686 ±   41865.767  ops/s
StringsBenchmark.bufferIsEmpty      thrpt   15    4001127.728 ±   14525.221  ops/s
StringsBenchmark.fastIsEmpty        thrpt   15  139232012.995 ± 1757089.132  ops/s
StringsBenchmark.isEmpty            thrpt   15  130471674.653 ± 1604850.726  ops/s
```

The new method is about 15% faster for `AppendingStringBuffer` and about 5-10% faster for `String`.

Notes:

1. The only case where these optimization will probably result in minor performance loss is for strings that actually start with whitespace. In that case we check the first character and then call trim. This shouldn't be a problem though because in 99.9% of cases the method is called with a null string, an empty string, or a "normal" string.
2. `string.charAt(0) <= ' '` is taken directly from `String.trim`, so there should be no hidden change in behavior.

Benchmark:

```java

import org.apache.wicket.util.string.AppendingStringBuffer;
import org.apache.wicket.util.string.Strings;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.infra.Blackhole;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;
import org.openjdk.jmh.runner.options.TimeValue;

import java.util.UUID;

@State(Scope.Thread)
public class StringsBenchmark {

	private final String s = UUID.randomUUID().toString();
	private final CharSequence c = UUID.randomUUID().toString();
	private final AppendingStringBuffer buffer = new AppendingStringBuffer(UUID.randomUUID().toString().repeat(10));

	@Benchmark
	public void bufferIsEmpty(Blackhole blackhole) {
		blackhole.consume(Strings.isEmpty(buffer));
	}

	@Benchmark
	public void bufferFastIsEmpty(Blackhole blackhole) {
		blackhole.consume(isEmpty(buffer));
	}

	@Benchmark
	public void isEmpty(Blackhole blackhole) {
		blackhole.consume(Strings.isEmpty(c));
	}

	@Benchmark
	public void fastIsEmpty(Blackhole blackhole) {
		blackhole.consume(isEmpty(s));
	}

	private static boolean isEmpty(final CharSequence string)
	{
		return string == null || string.length() == 0 ||
				(string.charAt(0) <= ' ' && string.toString().trim().isEmpty());
	}

	private static boolean isEmpty(final String string)
	{
		return string == null || string.isEmpty() ||
				(string.charAt(0) <= ' ' && string.trim().isEmpty());
	}

	public static void main(String[] args) throws RunnerException {
		final Options opt = new OptionsBuilder()
				.include(".*" + StringsBenchmark.class.getSimpleName() + ".*")
				.warmupIterations(3)
				.warmupTime(TimeValue.seconds(3))
				.measurementIterations(3)
				.measurementTime(TimeValue.seconds(3))
				.threads(1)
				.forks(3)
				.build();
		new Runner(opt).run();
	}
}
```

### DIFF
--- a/wicket-util/src/main/java/org/apache/wicket/util/lang/Args.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/lang/Args.java
@@ -66,6 +66,28 @@ public class Args
 		return argument;
 	}
 
+	/**
+	 * Checks argument is not empty (not null and has a non-whitespace character)
+	 *
+	 * Note: This method overloads {@link #notEmpty(CharSequence, String)} for performance reasons.
+	 * 
+	 * @param argument
+	 *            the argument to check for emptiness
+	 * @param name
+	 *            the name to use in the error message
+	 * @return The {@code argument} parameter if not empty
+	 * @throws IllegalArgumentException
+	 *             when the passed {@code argument} is empty
+	 */
+	public static String notEmpty(final String argument, final String name)
+	{
+		if (Strings.isEmpty(argument))
+		{
+			throw new IllegalArgumentException("Argument '" + name + "' may not be null or empty.");
+		}
+		return argument;
+	}
+
 
 	/**
 	 * 
--- a/wicket-util/src/main/java/org/apache/wicket/util/string/Strings.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/string/Strings.java
@@ -531,8 +531,26 @@ public final class Strings
 	 */
 	public static boolean isEmpty(final CharSequence string)
 	{
-		return (string == null) || (string.length() == 0) ||
-			(string.toString().trim().length() == 0);
+		return string == null || string.length() == 0 ||
+			(string.charAt(0) <= ' ' && string.toString().trim().isEmpty());
+	}
+
+	/**
+	 * Checks whether the <code>string</code> is considered empty. Empty means that the string may
+	 * contain whitespace, but no visible characters.
+	 *
+	 * "\n\t " is considered empty, while " a" is not.
+	 * 
+	 * Note: This method overloads {@link #isEmpty(CharSequence)} for performance reasons.
+	 *
+	 * @param string
+	 *            The string
+	 * @return True if the string is null or ""
+	 */
+	public static boolean isEmpty(final String string)
+	{
+		return string == null || string.isEmpty() ||
+			(string.charAt(0) <= ' ' && string.trim().isEmpty());
 	}
 
 	/**
--- a/wicket-util/src/test/java/org/apache/wicket/util/string/StringsTest.java
+++ b/wicket-util/src/test/java/org/apache/wicket/util/string/StringsTest.java
@@ -24,8 +24,6 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.lang.reflect.Field;
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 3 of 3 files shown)


svenmeier (approved):
+1

wicket-util/src/main/java/org/apache/wicket/util/lang/Args.java
```diff
@@ -66,6 +66,26 @@
...
+	 * Checks argument is not empty (not null and has a non-whitespace character)
+	 *
+	 * @param argument
+	 *            the argument to check for emptiness
+	 * @param name
+	 *            the name to use in the error message
+	 * @return The {@code argument} parameter if not empty
+	 * @throws IllegalArgumentException
+	 *             when the passed {@code argument} is empty
+	 */
```
martin-g, line 79: please add a comment why there are two "identical" methods.
in few months/years someone will wonder what is going on

wicket-util/src/main/java/org/apache/wicket/util/string/Strings.java
```diff
@@ -531,8 +531,24 @@ else if (aChar == 'f')
...
+	/**
+	 * Checks whether the <code>string</code> is considered empty. Empty means that the string may
+	 * contain whitespace, but no visible characters.
+	 *
+	 * "\n\t " is considered empty, while " a" is not.
+	 *
+	 * @param string
+	 *            The string
+	 * @return True if the string is null or ""
+	 */
```
martin-g, line 547: Same - please add a comment

martin-g (approved):

martin-g:
It would be good to have the benchmark code for future reference - either here as a comment or somewhere in the source tree

martin-g:
And please create a JIRA issue - for the changelog
Thanks!

theigl:
Right, I'll add the JavaDoc and the ticket tomorrow!

> It would be good to have the benchmark code for future reference - either here as a comment or somewhere in the source tree

I suggest we add a separate Maven module `wicket-benchmarks`, where we can add this and future benchmarks. I have such a module on a branch, but had problems with the enforcer plugin complaining about the plugins I added to run the benchmarks from the command line. I need to play around with the plugin definitions and will create a separate PR if I can get it working. The main advantage is that you can checkout the sources on a (cloud) server and run the benchmarks there. For some performance optimizations in Kryo, I had benchmarks running for hours over night, to get the error rates as low as possible.

theigl:
OK my previous benchmark was flawed because I created a new buffer for every run. The new `isEmpty` method for `AppendingStringBuffer´ is actually 40-50x faster for medium-sized buffers.

```
Benchmark                            Mode  Cnt           Score         Error  Units
StringsBenchmark.bufferFastIsEmpty  thrpt   15  1918461027.579 ± 4949234.208  ops/s
StringsBenchmark.bufferIsEmpty      thrpt   15    42567269.762 ±   40091.373  ops/s
StringsBenchmark.fastIsEmpty        thrpt   15  1910306184.208 ± 4067311.978  ops/s
StringsBenchmark.isEmpty            thrpt   15  1804225517.032 ± 1951669.787  ops/s
```

wicket-util/src/main/java/org/apache/wicket/util/lang/Args.java
theigl: Should I simply add: 

> This method overloads {@link #isEmpty(CharSequence)} for performance reasons.

Or should I speculate on why this is faster?

theigl:
https://issues.apache.org/jira/browse/WICKET-6952

theigl:
```
Benchmark                            Mode  Cnt           Score         Error  Units
StringsBenchmark.bufferFastIsEmpty  thrpt   15  1918461027.579 ± 4949234.208  ops/s
StringsBenchmark.bufferIsEmpty      thrpt   15    42567269.762 ±   40091.373  ops/s
```

wicket-util/src/main/java/org/apache/wicket/util/lang/Args.java
martin-g: `This method overloads {@link #isEmpty(CharSequence)} for performance reasons.` is enough.

martin-g (approved):

PR #498, opened 2022-02-05
Improvement/reiern70/wicket 6954 ping pong keep alive

reiern70:
@martin-g 

This work is not ready. I have just created a PR to get your early advice regarding the approach. 

How does it work?

- Via setting some timer process can be started that will use a thread pool to ping remote clients
- Clients will react to theses pings and terminate connection with some time-out
- The same timer thing check connections at server side and if they have not being alive for some time connection is terminated
- On receiving pong messages connection is marked as alive and optionally a message is delivered to page holding the connection.
- There is another flag that determines if client will try to reconnect to server (we maybe might need make this different from normal connection and generate some event telling page "Hey, connection was just re-stablished"). Maybe we need also some event when connection is terminated because of inactivity?


reiern70:
closing this PR:

1- It seems there is no way to listen at client side for pings
2- web sockets for master seems to be broken. 

I will open a new PR to 9.x (which seems to be working)

martin-g:
@reiern70 Did you mean to close this PR ? Your last comment suggest so.

reiern70:
> @reiern70 Did you mean to close this PR ? Your last comment suggest so.

Yes... But forgot to do so. WEB-sockets examples iare not working fro me master (even without my changes).

PR #504, opened 2022-03-22
Guarantee order of listener notifications

papegaaij:
`ListenerCollection` has a note that states that the order is not guaranteed. This seems strange as the collection is simply a list and it even has a method to notify the listeners in reverse order. I suggest to change the comment and actually guarantee the order of notification to be in match with the order the listeners were added.

### DIFF
--- a/wicket-util/src/main/java/org/apache/wicket/util/listener/ListenerCollection.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/listener/ListenerCollection.java
@@ -28,7 +28,8 @@ import org.slf4j.LoggerFactory;
 /**
  * Represents a collection of listeners. Facilitates invocation of events on each listener.
  * <p>
- * NOTE: Ordering of listeners is not guaranteed and should not be relied upon
+ * Listeners will be invoked in the order added to the collection when using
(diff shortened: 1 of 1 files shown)


bitstorm (approved):

martin-g:
I remember @ivaynberg having a conversation with someone on the mailing lists about depending on order. His wise advice was to never rely on someone/something else to provide it for you. It is better to check the environment/context/conditions and act as necessary. 
I don't remember the details now. I just try to follow this advice in my work.

papegaaij:
I wouldn't recommend to depend on listeners being added in a specific order, especially when they are added automatically, for example via a service loader. However, if you add them yourself, you are in control and we can guarantee the order of execution. This change only indicates that the listeners are notified in the order they were added, which does make sense imho, also because the class already has a method to invoke them in reverse (the reverse of an unordered set doesn't exist). 

This came up when a developer at our company wanted to change a specific listener rather than added his own, because this comment stated he could not rely on their order. He specifically wanted his code to be executed before another, both listeners are added manually.

martin-g:
No!
Go ahead!

PR #507, opened 2022-04-04
WICKET-6965 Store application name to avoid accessing the session

theigl:
This PR stores the application name in a field when a websocket connection is opened, so we don't have to access the session during `onClose` and `onError`. When `onError` is called, the session may already be closed and request parameters can no longer be accessed.

https://issues.apache.org/jira/browse/WICKET-6965?focusedCommentId=17516488&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17516488


bitstorm (approved):
Looks good to me!

PR #508, opened 2022-04-05
[WICKET-6972] Add Resource key to be logged on Warning in Localizer.java

renoth:
Suggestion: It would be very helpful to log the Resource key as well to easier identify the problem. 

Is this be security-sensitive? Can this be exploited somehow?


martin-g (approved):

martin-g:
@renoth Thanks for the PR!
I think it is safe to log the key.

We also need a ticket in JIRA for the release changelog. Once you have a ticket id please update the PR title and the commit message with it. Thanks!

martin-g:
Thank you, @renoth !

PR #510, opened 2022-04-13
[WICKET-6971] NullPointerException in ModificationWatcher

try0:
Adds null check.

https://issues.apache.org/jira/browse/WICKET-6971
https://lists.apache.org/thread/v3hbwsp3qs0rsnbsj6jp34662hrbjcgk

### DIFF
--- a/wicket-util/src/main/java/org/apache/wicket/util/io/Connections.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/io/Connections.java
@@ -18,6 +18,7 @@ package org.apache.wicket.util.io;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.JarURLConnection;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


martin-g:
Thanks for the PR, @try0 !

I wonder whether we should try/catch the whole method body. 
Closing a resource is a best effort operation and it should not break the following operations.
Currently the method `throws IOException`, i.e. the errors should be propagated. But I don't see much value in this. IMO catching `Exception` and logging it as a WARNING would be better.

Let's hear some more opinions!

try0:
Hi @martin-g .
Thank you for your advice.

>IMO catching Exception and logging it as a WARNING would be better.

Right.
I agree this opinion.

But author of the Connections class also implement closeQuietly method.
So I think it's better to keep the throw exceptions in close method.

It appears that there are only two callers of the close method, closeQuietly and org.apache.wicket.core.util.resource.UrlResourceStream#lastModifiedTime().

https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/core/util/resource/UrlResourceStream.java#L246

And lastModifiedTime() logs exceptions and return null.
https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/core/util/resource/UrlResourceStream.java#L227

So,
```
InputStream inputStream = connection.getInputStream();
if (inputStream == null)
{
	throw new IOException("No stream.");
}
```
Is it better this way?
NullPointerException can be avoided and the code to be changed is minimal.

martin-g (approved):
Right! There is `#closeQuietly()` for the other case.

I'm OK with the current solution! 
Throwing IOException("No stream") will log a stacktrace but the developer/maintainer won't be able to do anything about it.

try0:
Thanks for the review and approving PR!

martin-g:
Thank you, @try0 !

PR #512, opened 2022-04-19
List.of() creates an immutable List instance, without a backing Array...

astange1:
…List, saving one Object[] array creation and saving ~12 bytes for each component in the hierarchy.   This ends up being a few KB for each of our pages, resulting in meaningful memory savings, a little less GC pressure and a little more memory fitting into the processor caches.

This is a minor change.   I expect I will be sending along other small improvements like this one for the Wicket code base if there is an interest in receiving them.   FYI   We have been a Wicket user since ~2008.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -2669,7 +2669,7 @@ public abstract class Component
 					if (response.wasRendered(behavior) == false)
 					{
 						behavior.renderHead(this, response);
-						List<IClusterable> pair = Arrays.asList(this, behavior);
+						List<IClusterable> pair = List.of(this, behavior);
 						response.markRendered(pair);
 					}
 				}


solomax (approved):

wicket-core/src/main/java/org/apache/wicket/Component.java
```diff
@@ -2669,7 +2669,7 @@ public void internalRenderHead(final HtmlHeaderContainer container)
 					if (response.wasRendered(behavior) == false)
```
theigl, line 2669: The change itself looks good to me, but I'm wondering if the code is currently working as expected. We mark a pair of Component+Behavior as rendered, but the check if a behavior was already rendered only tests for the behavior itself.

@martin-g: You originally introduced this to fix https://issues.apache.org/jira/browse/WICKET-4570 many years ago. Could it be that you forgot to adjust the `wasRendered` condition? Imho the current code will never return `true` for `response.wasRendered(behavior)`.

martin-g: It looks like a bug indeed!
But it is strange that it was uncovered for so many years.

astange1:
I just mistakenly pushed a 2nd commit into this pull request.    I was going to make this a 2nd pull request, but it can go here too.

This 2nd commit greatly reduces the overhead of hashCode() computation in the rendering cycle as well.    These hashcode computations, which make calls to Objects.hash() create a temporary Object[] array, autobox the boolean and integer values, and are expensive due to this overhead.     I ran some tests on JDK17 and JDK19 using JMH and these Object[] creations are not elided in after inlining and after Escape Analysis.    One of our simplest Ajax callbacks results in 140 hashcode calls to the Abstract class.   In total, these changes avoid the creation of about 11KB of garbage memory in the server _per_mouse_click.     This code also runs in ~1ns, compared to ~15ns for the original code (on a M1 Ultra Macbook, JDK19).     On older JDKs, like JDK11, the original code is much more expensive by comparison.

astange1:
@theigl I just saw your note.   I've been using git for some time, but new to guthub.   I'll have to take some time to figure out how to remove this 2nd commit and move it into a new PR.    That was my original intent, and didn't expect the 2nd commit to show up in this PR.

theigl:
@astange1: There is some related code that might also benefit from using `List.of`.  E.g.

https://github.com/apache/wicket/blob/c6d18e697b699723ca2cc3788e64b56f5929445e/wicket-core/src/main/java/org/apache/wicket/markup/head/JavaScriptContentHeaderItem.java#L75

And all other occurrences of `Arrays.asList` in `org.apache.wicket.markup.head`.

wicket-core/src/main/java/org/apache/wicket/Component.java
theigl: Should we fix this? Or should we get rid of the whole conditional and always call `behavior.renderHead` as we seem to have been doing for 10 years without issues. All tests pass in both solutions.

martin-g: +1 to fix it

astange1:
I would be happy to proceed and make those needed edits.   I did just that one for two reasons:
1) it shows up as being a somewhat hot in some of my profiles.
2) I figured I would do that one to start.   If that simple PR was accepted then I would gladly do more, and if not then I hadn't wasted too much of my time.   

Let me know how you would like to proceed.    I can add the changes in this PR, or submit a new one.

theigl:
> I figured I would do that one to start. If that simple PR was accepted then I would gladly do more, and if not then I hadn't wasted too much of my time.

That sounds good to me. Let's keep it simple. Also I just realized that `List.of` is not null-safe. So we probably can't use it as a universal replacement for header items anyway.

wicket-core/src/main/java/org/apache/wicket/Component.java
theigl: @martin-g: I pushed https://github.com/apache/wicket/pull/515 that fixes this and incorporates the change suggested in this PR.

PR #513, opened 2022-04-19
WICKET-6977 More efficient hashcode computations

astange1:
AbstractJavaScriptReferenceHeaderItem and children use a simple, but somewhat expensive, hashcode computation.   We can obtain the same result using much less time, and avoid all creation of garbage objects.   This is a substantial memory savings on complex pages with a large component hierarchy.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -2669,7 +2669,7 @@ public abstract class Component
 					if (response.wasRendered(behavior) == false)
 					{
 						behavior.renderHead(this, response);
-						List<IClusterable> pair = List.of(this, behavior);
+						List<IClusterable> pair = Arrays.asList(this, behavior);
 						response.markRendered(pair);
 					}
 				}
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -2669,7 +2669,7 @@ public abstract class Component
 					if (response.wasRendered(behavior) == false)
 					{
 						behavior.renderHead(this, response);
-						List<IClusterable> pair = Arrays.asList(this, behavior);
+						List<IClusterable> pair = List.of(this, behavior);
 						response.markRendered(pair);
 					}
 				}
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
@@ -134,6 +134,7 @@ public abstract class AbstractJavaScriptReferenceHeaderItem extends JavaScriptHe
 
 	@Override
 	public int hashCode() {
+		// Not using `Objects.hash` for performance reasons
 		int result = super.hashCode();
 		result = 31 * result + (async ? 1 : 0);
 		result = 31 * result + (defer ? 1 : 0);
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
@@ -135,6 +135,16 @@ public abstract class AbstractJavaScriptReferenceHeaderItem extends JavaScriptHe
 	@Override
 	public int hashCode()
 	{
-		return Objects.hash(async, defer, charset);
+		// original code was: return Objects.hash(async, defer, charset);
+		// This returns the same hash code value, but doesn't allocate an Object[3],
+		// which consumes 32 bytes, and doesn't need to autobox the two booleans.
+		// This method is called very often, especially when a containing Map or Set is resized,
+		// and this version will run faster and save a lot of memory.
+		// It is possible that Escape Analysis would elide the creation of the Object[],
+		// but we do not see that in our running applications with the latest JDK.
+		return 31*31*31 +
+				31*31*Boolean.hashCode(async) +
+				31*Boolean.hashCode(defer) +
+				((charset != null) ? charset.hashCode() : 0);
 	}
 }
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
@@ -135,16 +135,6 @@ public abstract class AbstractJavaScriptReferenceHeaderItem extends JavaScriptHe
 	@Override
 	public int hashCode()
 	{
-		// original code was: return Objects.hash(async, defer, charset);
-		// This returns the same hash code value, but doesn't allocate an Object[3],
-		// which consumes 32 bytes, and doesn't need to autobox the two booleans.
-		// This method is called very often, especially when a containing Map or Set is resized,
-		// and this version will run faster and save a lot of memory.
-		// It is possible that Escape Analysis would elide the creation of the Object[],
-		// but we do not see that in our running applications with the latest JDK.
-		return 31*31*31 +
-				31*31*Boolean.hashCode(async) +
-				31*Boolean.hashCode(defer) +
-				((charset != null) ? charset.hashCode() : 0);
+		return Objects.hash(async, defer, charset);
 	}
 }
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
@@ -135,16 +135,7 @@ public abstract class AbstractJavaScriptReferenceHeaderItem extends JavaScriptHe
 	@Override
 	public int hashCode()
 	{
-		// original code was: return Objects.hash(async, defer, charset);
-		// This returns the same hash code value, but doesn't allocate an Object[3],
-		// which consumes 32 bytes, and doesn't need to autobox the two booleans.
-		// This method is called very often, especially when a containing Map or Set is resized,
-		// and this version will run faster and save a lot of memory.
-		// It is possible that Escape Analysis would elide the creation of the Object[],
-		// but we do not see that in our running applications with the latest JDK.
-		return 31*31*31 +
-				31*31*Boolean.hashCode(async) +
-				31*Boolean.hashCode(defer) +
-				((charset != null) ? charset.hashCode() : 0);
+		return (charset != null ? charset.hashCode() << 2 : 0)
+				+ (async ? 2 : 0) + (defer ? 1 : 0);
 	}
 }
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
@@ -133,9 +133,11 @@ public abstract class AbstractJavaScriptReferenceHeaderItem extends JavaScriptHe
 	}
 
 	@Override
-	public int hashCode()
-	{
-		return (charset != null ? charset.hashCode() << 2 : 0)
-				+ (async ? 2 : 0) + (defer ? 1 : 0);
+	public int hashCode() {
+		int result = super.hashCode();
+		result = 31 * result + (async ? 1 : 0);
+		result = 31 * result + (defer ? 1 : 0);
+		result = 31 * result + (charset != null ? charset.hashCode() : 0);
+		return result;
 	}
 }
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
@@ -133,7 +133,7 @@ public abstract class AbstractJavaScriptReferenceHeaderItem extends JavaScriptHe
 	}
 
 	@Override
-	public int hashCode() 
+	public int hashCode()
 	{
 		// Not using `Objects.hash` for performance reasons
 		int result = super.hashCode();
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
@@ -133,7 +133,8 @@ public abstract class AbstractJavaScriptReferenceHeaderItem extends JavaScriptHe
 	}
 
 	@Override
-	public int hashCode() {
+	public int hashCode() 
+	{
 		// Not using `Objects.hash` for performance reasons
 		int result = super.hashCode();
 		result = 31 * result + (async ? 1 : 0);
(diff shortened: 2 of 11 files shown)


wicket-core/src/main/java/org/apache/wicket/Component.java
```diff
@@ -2669,7 +2669,7 @@ public void internalRenderHead(final HtmlHeaderContainer container)
 					if (response.wasRendered(behavior) == false)
 					{
 						behavior.renderHead(this, response);
-						List<IClusterable> pair = Arrays.asList(this, behavior);
```
theigl, line 2672: Please revert the commit from this PR.

wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
```diff
@@ -135,6 +135,16 @@ public boolean equals(Object o)
...
+		// This returns the same hash code value, but doesn't allocate an Object[3],
+		// which consumes 32 bytes, and doesn't need to autobox the two booleans.
+		// This method is called very often, especially when a containing Map or Set is resized,
+		// and this version will run faster and save a lot of memory.
+		// It is possible that Escape Analysis would elide the creation of the Object[],
+		// but we do not see that in our running applications with the latest JDK.
+		return 31*31*31 +
+				31*31*Boolean.hashCode(async) +
+				31*Boolean.hashCode(defer) +
+				((charset != null) ? charset.hashCode() : 0);
```
theigl, line 148: I find this quite hard to read. Can you please use IntelliJ or Eclipse to generate the code. IntelliJ looks like this:

```java
	@Override
	public int hashCode() 
	{
		int result = super.hashCode();
		result = 31 * result + (async ? 1 : 0);
		result = 31 * result + (defer ? 1 : 0);
		result = 31 * result + (charset != null ? charset.hashCode() : 0);
		return result;
	}
```

theigl, line 148: I find this quite hard to read. Can you please use IntelliJ or Eclipse to generate the code. IntelliJ looks like this:

```
	@Override
	public int hashCode() 
	{
		int result = super.hashCode();
		result = 31 * result + (async ? 1 : 0);
		result = 31 * result + (defer ? 1 : 0);
		result = 31 * result + (charset != null ? charset.hashCode() : 0);
		return result;
	}
```

wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
```diff
@@ -135,6 +135,16 @@ public boolean equals(Object o)
...
 	public int hashCode()
 	{
-		return Objects.hash(async, defer, charset);
+		// original code was: return Objects.hash(async, defer, charset);
+		// This returns the same hash code value, but doesn't allocate an Object[3],
+		// which consumes 32 bytes, and doesn't need to autobox the two booleans.
+		// This method is called very often, especially when a containing Map or Set is resized,
+		// and this version will run faster and save a lot of memory.
+		// It is possible that Escape Analysis would elide the creation of the Object[],
+		// but we do not see that in our running applications with the latest JDK.
```
theigl, line 144: This comment should go into the corresponding Jira ticket. A simple comment like this should be enough:

```
// Not using `Objects.hash` for performance reasons
```

theigl, line 144: This comment should go into the corresponding Jira ticket. A simple comment like this should be enough:

```
// Not using `Objects.hash` for performance reasons
```

theigl (changes requested):
If we touch the hashCode for `JavaScriptHeaderItem` and sub-classes, we should do the same for `CssHeaderItem`.

wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
astange1: I wrote it the original way to produce the exact same hashcode value as the original code, and to minimize the math being done.    Are we OK with the hashcode values being different, perhaps resulting in other changes due to hashing?

I will update the CSSHeaderItem classes, and also remove the comments.

theigl: > Are we OK with the hashcode values being different, perhaps resulting in other changes due to hashing?

Good question. My guess would be that it doesn't matter, since collecting header items and putting them in collections is always done on a single JVM in a single request. Does anyone else see a problem with changing the hash codes?

solomax: `hash` method from here https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html can also be used :)

wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
```diff
@@ -135,6 +135,7 @@ public boolean equals(Object o)
 	@Override
 	public int hashCode()
 	{
-		return Objects.hash(async, defer, charset);
+		return (charset != null ? charset.hashCode() << 2 : 0)
```
theigl, line 138: Please use an auto-generated hash-code for this method as well. Doing everything consistently, makes it easier to simply regenerate the hash-code should a new field be added.

wicket-core/src/main/java/org/apache/wicket/markup/head/CssReferenceHeaderItem.java
```diff
@@ -154,7 +154,13 @@ private String getUrl()
 	@Override
 	public int hashCode()
 	{
-		return Objects.hash(super.hashCode(), reference, getMedia(), pageParameters, getRel());
+		String s;
```
theigl, line 157: I'm not a fan of re-using this `String s`. Did you IDE generate it this way?

wicket-core/src/main/java/org/apache/wicket/markup/head/CssReferenceHeaderItem.java
```diff
@@ -154,7 +154,13 @@ private String getUrl()
...
+		int result = super.hashCode();
+		result = 31*result + ((reference != null) ? reference.hashCode() : 0);
```
theigl, line 159: There should be whitespace around the multiplier. Please apply our standard code format.

theigl (changes requested):

theigl:
To give some more context, I ran the benchmarks myself. The overhead of `Objects.hash` is very significant:

```
Benchmark                                                Mode   Cnt          Score         Error   Units
Benchmarks.hashCodeNew                                   thrpt    9  134406237,465 ±  712957,234   ops/s
Benchmarks.hashCodeNew:·gc.alloc.rate.norm               thrpt    9         40,004 ±       0,001    B/op
Benchmarks.hashCodeNew:·gc.churn.G1_Eden_Space.norm      thrpt    9         40,115 ±       0,401    B/op
Benchmarks.hashCodeNew:·gc.churn.G1_Survivor_Space.norm  thrpt    9         ≈ 10⁻⁴                  B/op

Benchmarks.hashCodeOld                                   thrpt    9   29199877,793 ± 1543709,148   ops/s
Benchmarks.hashCodeOld:·gc.alloc.rate.norm               thrpt    9        120,015 ±       0,004    B/op
Benchmarks.hashCodeOld:·gc.churn.G1_Eden_Space.norm      thrpt    9        120,181 ±       1,556    B/op
Benchmarks.hashCodeOld:·gc.churn.G1_Survivor_Space.norm  thrpt    9          0,001 ±       0,001    B/op
```

The new hashcodes are **4x** faster and allocation is reduced significantly.

theigl:
@astange1: Given my benchmark results, we should definitely go ahead with these improvements.

I think we should apply them for *all* header items and in the most consistent way possible. I would vote for auto-generating all of the hashcodes without attempting any clever optimizations.

bitstorm:
> @astange1: Given my benchmark results, we should definitely go ahead with these improvements.
> 
> I think we should apply them for _all_ header items and in the most consistent way possible. I would vote for auto-generating all of the hashcodes without attempting any clever optimizations.

@theigl I think you've already been asked about it, but I'd like to know how do you create these micro-benchmarks. Do you use any particular framework?

theigl:
> @theigl I think you've already been asked about it, but I'd like to know how do you create these micro-benchmarks. Do you use any particular framework?

These are very simple JMH benchmarks. I already promised to create a `wicket-benchmarks` module where we can collect these benchmarks, but I haven't gotten around to it yet.

This specific benchmark looks like this:

```java
import org.apache.wicket.markup.head.JavaScriptReferenceHeaderItem;
import org.apache.wicket.markup.head.JavaScriptReferenceHeaderItemNew;
import org.apache.wicket.request.mapper.parameter.PageParameters;
import org.apache.wicket.resource.aggregator.ResourceReferenceA;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.infra.Blackhole;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;
import org.openjdk.jmh.runner.options.TimeValue;

@State(value = Scope.Thread)
public class HeaderItemBenchmarks {

    private final JavaScriptReferenceHeaderItem item = new JavaScriptReferenceHeaderItem(new ResourceReferenceA(), new PageParameters(), "anyId");
    private final JavaScriptReferenceHeaderItemNew item2 = new JavaScriptReferenceHeaderItemNew(new ResourceReferenceA(), new PageParameters(), "anyId");

    @Benchmark
    public void hashCodeOld(Blackhole blackhole)
    {
        blackhole.consume(item.hashCode());
    }

    @Benchmark
    public void hashCodeNew(Blackhole blackhole)
    {
        blackhole.consume(item2.hashCode());
    }

    public static void main(String[] args) throws RunnerException
    {
        final Options opt = new OptionsBuilder()
                .include(".*" + HeaderItemBenchmarks.class.getSimpleName() + ".*.hash.*")
                .addProfiler("gc")
                .warmupIterations(3)
                .warmupTime(TimeValue.seconds(3))
                .measurementIterations(3)
                .measurementTime(TimeValue.seconds(3))
                .threads(1)
                .forks(3)
                .build();
        new Runner(opt).run();
    }
}
```

astange1:
Hello, I believe I have made all the code formatting chnages that were requested.  I also made an additional update to the hashcode computation for a ResourceReference class, which should result in improved JMH performance.   

@theigl   The remaining memory allocation in the posted JMH results should be resolved now.   Can you kindly re-run the results with this updated code to confirm?    My JMH setup is running on a custom JDK19 build, and it would be a pain for me to set up here for Wicket.

FYI.  Wicket-utils has an Objects.hashCode() implementation that should be removed in favor of using the one in the JDK.   It is used by the Model class.

wicket-core/src/main/java/org/apache/wicket/markup/head/CssUrlReferenceHeaderItem.java
```diff
@@ -97,7 +97,12 @@ public String toString()
 	@Override
 	public int hashCode()
 	{
-		return Objects.hash(super.hashCode(), url, getMedia(), getRel());
+		String s;
+		int result = super.hashCode();
+		result = 31*result + ((url != null) ? url.hashCode() : 0);
+		result = 31*result + (((s = getMedia()) != null) ? s.hashCode() : 0);
+		result = 31*result + (((s = getRel()) != null) ? s.hashCode() : 0);
+		return result;
```
theigl, line 105: Please remove the `s` variable and fix the formatting.

wicket-core/src/main/java/org/apache/wicket/markup/head/JavaScriptContentHeaderItem.java
```diff
@@ -94,6 +94,8 @@ public boolean equals(Object o)
 	@Override
 	public int hashCode()
 	{
-		return Objects.hash(super.hashCode(), javaScript);
+		int result = super.hashCode();
+		result = 31*result + ((javaScript != null) ? javaScript.hashCode() : 0);
```
theigl, line 98: Please adjust the formatting.

theigl (changes requested):
There are two more minor formatting/style issues.

And we should probably do the same optimization for `OnEventHeaderItem` and `MetaDataHeaderItem`. Both allocate heavily at the moment.

If my benchmarks are correct, we can leave all calls to `Objects.hash` with a single string argument. Those do not seem to benefit from a manual hashcode. Did you see the same result in your tests?

astange1:
I just pushed changed to address the formatting and the two additional files.

Hotspot is able to ellide the OIbject[] creation for calls to Objects.hash() with a single object, so those instances will not benefit from rewriting.   It might be better to use Objects.hashCode() though, as the Object[] ellision probably doesn't happen until the C2 compiler is able to fully optimize the code.

theigl:
Please create a ticket for this issue in our [Jira](https://issues.apache.org/jira/browse/WICKET) with your original observations, so we have the documentation and a reference for the release notes.

I think we should add a brief comment for all the new hashCode implementations, so that doesn't get changed back accidentally in the future. Something like:

```// Not using `Objects.hash` for performance reasons```

Otherwise, this looks good to me now.

**Comment by astange1** [comment]
See https://issues.apache.org/jira/browse/WICKET-6977
I will add comments to the updated hashCode methods tomorrow.

**Comment by astange1** [comment]
I'm sorry for the delay.   I got a little busy with some other obligations.   I believe I have added comments to all the impacted hashCode methods.

I believe this work is now complete.   Please let me know if I missed something or if there is some other addition needed for this change.

File: wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java [hunk 9]
```diff
@@ -135,6 +135,11 @@ public boolean equals(Object o)
 	@Override
 	public int hashCode()
 	{
-		return Objects.hash(async, defer, charset);
+		// Not using `Objects.hash` for performance reasons
+		int result = super.hashCode();
```
theigl, line 139: I just noticed some inconsistencies:

You added `super.hashCode` here, but not to `AbstractCssReferenceHeaderItem`. We should probably keep those two classes in sync.

And the bigger issue: The contracts for equals and hashCode now don't align. If we add `super.hashCode` (which imho we should, because the super class contains `Objects.hash(markupId)`), we should also add `markupId` to the equals contract.

@martin-g: Do you have any idea, why `AbstractJavaScriptReferenceHeaderItem` and `AbstractCssReferenceHeaderItem` do not currently user `super.equals` and `super.hashCode`. Do we intentionally ignore the `markupId` field in the super class? Your original commit for these implementations was 47aa1461d58b9a5660fc11032db952ce4df36a33.

theigl (commented):
@astange1: I found some minor inconsistencies and will take over clarifying this and polishing the PR.

Thank you very much for your contribution!

Did you identify further allocation hotspots or do your three PRs address the most important issues?

wicket-core/src/main/java/org/apache/wicket/markup/head/AbstractJavaScriptReferenceHeaderItem.java
theigl: If we are unsure about this, I'll simply remove the `super.hashCode` from the new implementation to maintain the current behavior.

martin-g: I don't remember now but I guess those were generated by my IDE.
I agree that super should be called!

theigl (approved):

martin-g (approved):

theigl:
@astange1: This has been merged and backported to 9.x Thanks again!

If you identify more allocation hotspots in the future, please feel free to create more PRs.

PR #514, opened 2022-04-20
Avoid generating so many synthetic closing tags.

astange1:
We have observed the generation of many trivial instances of these synthetic closing tags.   One random page I loaded on our production system generated 10,600 instances, with each instance being about 48 bytes in size.   This ~500KB of generated garbage objects, for each page load, produces substantial GC pressure on active servers with non-trivial page complexity. 

I did not have any examples were the namespace was not-null, so I ignored that case and maintained the original code.
I also avoided using String instances to avoid the char[] deflation to Latin1, and then inflating the bytes back into chars when appending to another AppendingStringBuffer.   The method returns a CharSequence, which should limit some code attempting to mutate these instances accidentally.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/ComponentTag.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/ComponentTag.java
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.wicket.Component;
 import org.apache.wicket.MarkupContainer;
@@ -125,6 +126,8 @@ public class ComponentTag extends MarkupElement
 	private Map<String, Object> userData;
 
 	private IAutoComponentFactory autoComponentFactory;
+
+	private static final Map<String,AppendingStringBuffer> tagMap = new ConcurrentHashMap<>();
 
 	/**
 	 * Automatically create a XmlTag, assign the name and the type, and construct a ComponentTag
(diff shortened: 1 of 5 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/ComponentTag.java
```diff
@@ -126,6 +127,8 @@ public class ComponentTag extends MarkupElement
 
 	private IAutoComponentFactory autoComponentFactory;
 
+	private static final Map<String,AppendingStringBuffer> tagMap = new ConcurrentHashMap<>();
```
theigl, line 130: I feel a bit uneasy about using an unbounded map. Since Wicket might be used to parse arbitrary XML, we have no way of knowing how many distinct tags users may have in their application. This will probably be fine for 99.9% of use cases, but it might cause issues for some users.

Possible alternatives: 

- Create a cache map of all known HTML5 tags
- Use a bounded map. We currently only seem to have a `MostRecentlyUsedMap`, that is not thread-safe.

Does anyone else have better ideas?

I'm very interested in this optimization because GC pressure under high load is the main problem in our application.

papegaaij: Why not go for a much easier approach: keep the method as is and mark it deprecated in wicket 9.x and add a new method `writeSyntheticCloseTag(Response)` that writes the close tag directly to the `Response`. That would remove all the object allocations and simplify the code a lot. The solution with a cache seems overly complicated and might introduce congestion with lookups and updates.

astange1: @theigl I appreciate your concern about an unbounded map.   I toyed with the idea of pre-populating the map with a small set of frequently used tags.  That's a small change and gets 99% of the benefit.   I wrote the change this way as it was trivial and gets the conversation started.

I will tweak the code slightly in the case when a buffer is still allocated to size the internal character array properly and to be less wasteful.


@papegaaij Note that the original allocating code runs in twice the time of the map lookups, and has a greater probability of contention caused by GC pauses than any possible contention in the map (ConcurrentHashMap only contends on structural updates, which will be very rare after the few lines of HTML are rendered in the application).

papegaaij: @astange1 the original code is not very optimal indeed, but I don't see why this method needs to return a closing tag, when it's only called from 2 sites, and both sites immediately append the result to a response. Simply skipping the whole return value and appending the closing tag to the response avoids any allocations, does not need a global synchronized store, avoids a hash lookup and is a lot easier to understand. Even though this method is strictly speaking part of the public API of wicket, it is not meant to be used externally. Changing the method to better suit the needs is ok, as long as we keep the old method for strict API compatibility in 9.x. We can drop it in 10.

astange1: Excellent.   I figured there must have been a better way of doing this.

PR #516, opened 2022-04-20
WICKET-6976 Avoid allocations when writing synthetic close tag

theigl:
Alternative implementation of #514 as suggested by @papegaaij.

https://issues.apache.org/jira/projects/WICKET/issues/WICKET-6976


papegaaij (approved):
Yes, this is exactly what I meant. Good work!

PR #517, opened 2022-04-20
WICKET-6963 Singleton markup sourcing strategy (Take 2)

theigl:
This is my second attempt at using a singleton PanelMarkupSourcingStrategy for panels and borders.

My initial PR #503 had to be reverted, because I overlooked the non-final field `AssociatedMarkupSourcingStrategy#noMoreWicketHeadTagsAllowed` resulting in concurrency issues.

The field is questionable and the code related to it full of TODO and Why? comments. This PR gets rid of it. We have nearly two dozen tests for this code in HeaderSectionTest and the only result of deleting the field is a different error message in one of these tests.

The error message for the test changes from:
 
> \<wicket:head> tags are only allowed before \<body>, \</head>, \<wicket:panel> etc. tag

to the following message thrown from HtmlHeaderResolver: 

> Mis-placed \<wicket:head>. \<wicket:head> must be outside of \<wicket:panel>, \<wicket:border>, and \<wicket:extend>

Since the flag is used for validation only, this change cannot break existing code.

https://issues.apache.org/jira/browse/WICKET-6963

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/panel/PanelMarkupSourcingStrategy.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/panel/PanelMarkupSourcingStrategy.java
@@ -33,8 +33,27 @@ import org.apache.wicket.markup.MarkupStream;
  */
 public class PanelMarkupSourcingStrategy extends AssociatedMarkupSourcingStrategy
 {
+	private static final PanelMarkupSourcingStrategy PANEL_INSTANCE = new PanelMarkupSourcingStrategy(
+		false);
+	private static final PanelMarkupSourcingStrategy BORDER_INSTANCE = new PanelMarkupSourcingStrategy(
+		true);
+
 	// False for Panel and true for Border components.
 	private final boolean allowWicketComponentsInBodyMarkup;
+
+	/**
+	 * @param allowWicketComponentsInBodyMarkup
+	 *            {@code false} for Panel and {@code true} for Border components. If Panel then the
+	 *            body markup should only contain raw markup, which is ignored (removed), but no
+	 *            Wicket Component. With Border components, the body markup will be associated with
+	 *            the Body Component.
+	 * 
+	 * @return A singleton of the strategy
+	 */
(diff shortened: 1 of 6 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/html/panel/PanelMarkupSourcingStrategy.java
```diff
@@ -33,9 +33,28 @@
  */
 public class PanelMarkupSourcingStrategy extends AssociatedMarkupSourcingStrategy
 {
+	private static final PanelMarkupSourcingStrategy PANEL_INSTANCE = new PanelMarkupSourcingStrategy(
+		false);
+	private static final PanelMarkupSourcingStrategy BORDER_INSTANCE = new PanelMarkupSourcingStrategy(
+		true);
+
 	// False for Panel and true for Border components.
 	private final boolean allowWicketComponentsInBodyMarkup;
```
martin-g, line 42: Is it a good moment to re-work this boolean to an enum ?
With an enum it would be possible to have a third/forth/... state if needed. And it is more clear than `true`/`false`

theigl: @martin-g: I experimentally refactored this in https://github.com/theigl/wicket/pull/1. Could you take a look? I couldn't think of a good name for the new enum. I'm not sure this refactoring is really worth it.

martin-g: > I couldn't think of a good name for the new enum.

`Type` ?

martin-g: I'd drop the usage of the `boolean` in the enum. Just compare the variants.

> I'm not sure this refactoring is really worth it.

I think it is a good idea, but I am not keen on it.
Thanks for trying it anyway!

PR #521, opened 2022-05-02
Reduce the Object creation in toString() calls to ComponentInfo 

astange1:
The toString() methods use temporary StringBuilders when they are not needed.   This produces additional garbage objects, requiring more GC work, and is also computationally slower.   String concatenation has been indyfy'd in newer JDKs, so  avoid using the classic idiom of allocating a StringBuffer in this simple case.   (JEP 280 landed in JDK9).

These changes should incrementally reduce object creation and slightly improve application responsiveness.


astange1:
I realize that I messed up the PageComponentInfo.java method.   I need to give this some more thought.   
I am closing this PR and will resubmit it after some more thought.

PR #522, opened 2022-05-02
WICKET-6979 Cut back slightly on some String instance creation in PageInfo and friends.

astange1:
These are some simple improvements that reduces unneeded String allocation for some Wicket objects, and sizes a 
StringBuilder sensibly to avoid a resize operation on every use.

I suspect there are some other patterns like this in the code.   These are the ones that stood out in our applications.

### DIFF
--- a/wicket-request/src/main/java/org/apache/wicket/request/mapper/info/PageComponentInfo.java
+++ b/wicket-request/src/main/java/org/apache/wicket/request/mapper/info/PageComponentInfo.java
@@ -35,7 +35,7 @@ public class PageComponentInfo
 
 	/**
 	 * Construct.
-	 * 
+	 *
 	 * @param pageInfo
 	 * @param componentInfo
 	 */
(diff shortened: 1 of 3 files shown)


wicket-request/src/main/java/org/apache/wicket/request/mapper/info/PageComponentInfo.java
```diff
@@ -69,23 +69,18 @@ public ComponentInfo getComponentInfo()
...
-		if (pageInfo != null)
-		{
-			result.append(pageInfo.toString());
-		}
+		String result = pageInfo.toString();
 		if (componentInfo != null)
 		{
-			result.append(SEPARATOR);
-			result.append(componentInfo);
+			result = result + SEPARATOR + componentInfo;
```
martin-g, line 75: The check `if (pageInfo != null)` is no more there.

astange1: The constructor for the class ensures that this final field can not be null.

martin-g (approved):

theigl (approved):
Please create a ticket for this and add prefix the PR name with the ticket ID.

Otherwise LGTM.

PR #523, opened 2022-05-05
Reduce object creation and wasted memory in Url toString() methods

astange1:
The processing for toString() in the Url class is a bit expensive.     I think we can do better by using String concatenation, avoiding copying  byte[]/char[] arrays, and sizing StringBuilders effectively.

More could be done to improve this, but these changes are not too aggressive.   Ideally a common StringBuilder would be passed around, like the charset, to accumulate the String as processing is done.

### DIFF
--- a/wicket-request/src/main/java/org/apache/wicket/request/Url.java
+++ b/wicket-request/src/main/java/org/apache/wicket/request/Url.java
@@ -1039,9 +1039,12 @@ public class Url implements Serializable
 		public String toString(final Charset charset)
 		{
 			String value = getValue();
-			if (Strings.isEmpty(value)) {
+			if (Strings.isEmpty(value))
+			{
 				return encodeParameter(getName(), charset);
-			} else {
+			}
+			else
+			{
 				return encodeParameter(getName(), charset) + "=" + encodeParameter(value, charset);
 			}
 		}
--- a/wicket-request/src/main/java/org/apache/wicket/request/Url.java
+++ b/wicket-request/src/main/java/org/apache/wicket/request/Url.java
@@ -801,7 +801,7 @@ public class Url implements Serializable
 					StringMode.FULL.name() + " mode because it has a `..` segment: " + toString());
 			}
 
-			if (!path.toString().startsWith("/"))
+			if (!path.isEmpty() && !(path.charAt(0) == '/'))
 			{
 				result.append('/');
 			}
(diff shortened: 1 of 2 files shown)


wicket-request/src/main/java/org/apache/wicket/request/Url.java
```diff
@@ -747,8 +747,26 @@ public String toString()
...
+				return path.toString();
+			}
+			else {
+				return path + "?" + queryString;
+			}
+		}
+
+		// fall through into the traditional code path
+
+		StringBuilder result = new StringBuilder(64;
```
martin-g, line 769: missing closing `)`

wicket-request/src/main/java/org/apache/wicket/request/Url.java
```diff
@@ -781,22 +799,19 @@ else if (Strings.isEmpty(protocol) && Strings.isEmpty(host) == false)
 					StringMode.FULL.name() + " mode because it has a `..` segment: " + toString());
 			}
 
-			if (!path.startsWith("/"))
+			if (!path.toString().startsWith("/"))
```
martin-g, line 802: can we use `java.lang.CharSequence#charAt(int)` instead ?

wicket-request/src/main/java/org/apache/wicket/request/Url.java
```diff
@@ -1021,14 +1036,11 @@ public String toString()
 		 */
 		public String toString(final Charset charset)
 		{
-			StringBuilder result = new StringBuilder();
-			result.append(encodeParameter(getName(), charset));
-			if (!Strings.isEmpty(getValue()))
-			{
-				result.append('=');
-				result.append(encodeParameter(getValue(), charset));
+			if (Strings.isEmpty(getValue())) {
```
martin-g, line 1039: cache the value of `getValue()` as a local variable and reuse it

wicket-request/src/main/java/org/apache/wicket/request/Url.java
```diff
@@ -747,8 +747,26 @@ public String toString()
...
+		// this method is rarely called with StringMode == FULL.
+
+		final CharSequence path = getPathInternal(charset);
+		final String queryString = getQueryString(charset);
+		String _fragment = getFragment();
+
+		// short circuit all the processing in the most common cases
+		if (StringMode.FULL != mode && Strings.isEmpty(_fragment))
+		{
+			if (queryString == null) {
```
solomax, line 759: opening `{` should be on new line :)

wicket-request/src/main/java/org/apache/wicket/request/Url.java
```diff
@@ -1021,14 +1038,12 @@ public String toString()
...
-			StringBuilder result = new StringBuilder();
-			result.append(encodeParameter(getName(), charset));
-			if (!Strings.isEmpty(getValue()))
-			{
-				result.append('=');
-				result.append(encodeParameter(getValue(), charset));
+			String value = getValue();
+			if (Strings.isEmpty(value)) {
+				return encodeParameter(getName(), charset);
+			} else {
```
solomax, line 1044: `{`, `}` on the new line (here and later in the code) :)

astange1:
I'm curious why this PR is being closed without merging the changes?

solomax:
this issue https://issues.apache.org/jira/browse/WICKET-6992 look similar

PR #528, opened 2022-06-03
Make the servlet API a runtime provided dependency when using JPMS.

asvanberg:
When building a custom runtime for a Wicket application a servlet container will have to be required which will pull in the dependency. Symmetrical with how it works when using Maven to build a WAR file that is deployed to a stand-alone application server.

### DIFF
--- a/wicket-core/src/main/java/module-info.java
+++ b/wicket-core/src/main/java/module-info.java
@@ -25,7 +25,7 @@ module org.apache.wicket.core {
     requires org.apache.commons.collections4;
     requires commons.fileupload;
     requires org.slf4j;
-    requires javax.servlet.api;
+    requires static javax.servlet.api;
     requires org.danekja.jdk.serializable.functional;
     requires com.github.openjson;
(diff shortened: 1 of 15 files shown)


svenmeier:
I'm curious about how is that supposed to work? There are many code places requiring HttpServletRequest, doesn't this break when javax.servlet isn't available?

asvanberg:
The module is still available at compile time. This is just bringing it to parity with how regular WAR deployment works. When building a WAR file the servlet specification is not included in the WAR file but instead provided by the application server (servlet container, like Tomcat). Similarly if you're building a trimmed down runtime image of a Wicket application an embedded application server will have to be required which would in turn pull in the servlet specification and implementation.

bitstorm (approved):
If I understand it right 'static' for JPMS is comparable to scope 'required' for Maven. So this PR looks correct to me.

bitstorm:
@martin-g have you had the chance to look at this proposed PR? Looks good to me.

martin-g (approved):

PR #532, opened 2022-09-02
WICKET-7002 Remove blocking code from Application#get/setMetaData (10.x)

dashorst:
By replacing the datastructure for Application class's metaData field
with a ConcurrentHashMap we can remove the synchronization on the
application instance and speed up lookups of the metadata. For at least
one application this will remove 50% of all blocking calls.


theigl:
I think this change makes a lot of sense. I just debugged my app and this synchronized method is hit about 100 times per request on my larger pages.

PR #540, opened 2022-09-29
[WICKET-6999] Update bundle plugin to support slf4j [1.7,3)

mattrpav:


mattrpav:
At issue is the bundle plugin runs different at various maven phases. When combined with the enforcer plugin to restrict config to an execution, the bundle plugin definition will get chatty and redundant.

Thoughts?

mattrpav:
Commit f2edd61 fixes bundle plugin execution phase+goal to generate correct MANIFEST and allow for enforcer plugin to be maintained.

mattrpav:
wicket-request:

```
Export-Package = 
	org.apache.wicket.request;version=10.0.0,
	org.apache.wicket.request.flow;version=10.0.0,
	org.apache.wicket.request.handler;version=10.0.0,
	org.apache.wicket.request.handler.logger;version=10.0.0,
	org.apache.wicket.request.http;version=10.0.0,
	org.apache.wicket.request.http.flow;version=10.0.0,
	org.apache.wicket.request.http.handler;version=10.0.0,
	org.apache.wicket.request.mapper;version=10.0.0,
	org.apache.wicket.request.mapper.info;version=10.0.0,
	org.apache.wicket.request.mapper.parameter;version=10.0.0,
	org.apache.wicket.request.parameter;version=10.0.0
Import-Package = 
	org.slf4j;version="[1.7,3)",
	jakarta.servlet.http;version="[5.0,6)",
	org.apache.commons.collections4;version="[4.4,5)",
	org.apache.wicket.util;version="[10.0,11)",
	org.apache.wicket.util.encoding;version="[10.0,11)",
	org.apache.wicket.util.io;version="[10.0,11)",
	org.apache.wicket.util.lang;version="[10.0,11)",
	org.apache.wicket.util.string;version="[10.0,11)",
	org.apache.wicket.util.time;version="[10.0,11)"
```

mattrpav:
NOTE: Generate export packages list for wicket-core

```
% find . -name "*.java" -exec grep "^package" {} \; | cut -f2 -d" " | sort -u
```

mattrpav:
I've removed the DynamicImport-Package: * rule. I think that should only be needed by the application wars to support modular pages and allow Wicket to instantiate those classes from the war's classpath.

mattrpav:
> > I've removed the DynamicImport-Package: * rule. I think that should only be needed by the application wars to support modular pages and allow Wicket to instantiate those classes from the war's classpath.
> 
> I think it would be better not to remove DynamicImport-Package: *, because of deserialization problems in OSGi env (Wicket need to know all classes serialized inside components from other bundles). It can be achieve in different way, but I think it requires more changes in the framework.

@dstoch the actual _application_ bundle would need to do that as they require. Further backing evidence is the Wicket 8.x bundles do not have 'DynamicImport-Package: *' specified.

dstoch:
> @dstoch the actual _application_ bundle would need to do that as they require. Further backing evidence is the Wicket 8.x bundles do not have 'DynamicImport-Package: *' specified.

MANIFESTs are broken since Wicket 8.x, so it is not a good evidence ;).

mattrpav:
> MANIFESTs are broken since Wicket 8.x, so it is not a good evidence ;).

What is broken? We've been running unmodified bundles on Wicket 8.x for years.

dstoch:
I described a problem in this issue description in JIRA. Eg. without exporting org.apache.wicket.markup.html.internal package is not possible to use Wicket in OSGi. Do you really use it in OSGi without problems? I wonder how is this possible, maybe you don't use the code which requires abstractions from this packages?

mattrpav:
> I described a problem in this issue description in JIRA. Eg. without exporting org.apache.wicket.markup.html.internal package is not possible to use Wicket in OSGi. Do you really use it in OSGi without problems? I wonder how is this possible, maybe you don't use the code which requires abstractions from this packages?

Correct. We do not extend anything from that package. If you see in the above, the new OSGi wiring includes exporting that package from wicket-core.

mattrpav:
> I described a problem in this issue description in JIRA. Eg. without exporting org.apache.wicket.markup.html.internal package is not possible to use Wicket in OSGi. Do you really use it in OSGi without problems? I wonder how is this possible, maybe you don't use the code which requires abstractions from this packages?

Keep in mind-- if you need to extend packages that have not been exported you can create a fragment bundle to attach your extension classes. Generally, things like 'internal' are _not_ exported since they are not part of a proper API or SPI.

mattrpav:
> > Generally, things like 'internal' are not exported since they are not part of a proper API or SPI.
> 
> In OSGi world it is indeed. But as I wrote in description in Wicket they are part of API (maybe these packages should not have "internal" word in name).

Agreed, Wicket's API+SPI definition and encapsulation could be better. The API+SPI should be separated into its own jar and only include interfaces and POJO classes. I created https://issues.apache.org/jira/browse/WICKET-6855 to open discussion on this, but I didn't articulate the use case in a meaningful way.

bitstorm (approved):
Thank you for your PR! When it'ok for you I will proceed merging it.

mattrpav:
Let me get a couple more done. For sure we should have the web socket bundles in good shape

pom.xml
```diff
@@ -891,18 +894,19 @@
...
-								<goal>manifest</goal>
+								<goal>bundle</goal>
 							</goals>
 							<configuration>
 								<instructions>
-									<Export-Package>*</Export-Package>
-									<Import-Package>org.apache.wicket*,
-										org.junit.jupiter*</Import-Package>
-									<DynamicImport-Package>*</DynamicImport-Package>
+									<Require-Capability>osgi.ee;filter:="(&amp;(osgi.ee=JavaSE)(version=11.0))"</Require-Capability>
```
martin-g, line 904: This probably should be version 17.0 for `master` and `11` for wicket-9.x

pom.xml
```diff
@@ -891,18 +894,19 @@
...
+									<Export-Package>${osgi.export.package}</Export-Package>
+									<Import-Package>${osgi.import.package}</Import-Package>
+									<DynamicImport-Package>${osgi.dynamicimport.package}</DynamicImport-Package>
 									<_nouses>true</_nouses>
+									<!-- _noee>true</_noee -->
```
martin-g, line 909: This could be removed, right ?

mattrpav: > OsgiClashingPackagesTest.collectProjectPackages() is failing.

I'll look into that.

> <Require-Capability>osgi.ee;filter:="(&amp;(osgi.ee=JavaSE)(version=11.0))"</Require-Capability>
> This probably should be version 17.0 for master and 11 for wicket-9.x

Does Wicket master have _runtime_ require on JDK 17?

> 
> This could be removed, right ?

This flag would prevent the Require-Capability header from being generated.

mattrpav: For Wicket, I think it is safe to remove the Require-Capability header altogether. (ie.. use the _noee = true flag.

Web app users are going to need to know full stack to align JDK + Karaf + Pax Web + Wicket at all times. This header only provides a pre-wiring check to fast fail.

martin-g: > Does Wicket master have _runtime_ require on JDK 17?

Yes!

mattrpav:
New feature to ease installs on Apache Karaf

```
karaf@root()> repo-add mvn:org.apache.wicket/wicket/10.0.0-M1-SNAPSHOT/xml/features
Adding feature url mvn:org.apache.wicket/wicket/10.0.0-M1-SNAPSHOT/xml/features
karaf@root()> feature:list | grep -i wicket
wicket-core                       │ 10.0.0.M1-SNAPSHOT │          │ Uninstalled │ org.apache.wicket.wicket-10.0.0-M1-SNAPSHOT │
karaf@root()> feature:info wicket-core
Feature wicket-core 10.0.0.M1-SNAPSHOT
Feature has no configuration
Feature has no configuration files
Feature has no dependencies.
Feature contains followed bundles:
  mvn:com.github.openjson/openjson/1.0.12
  mvn:commons-io/commons-io/2.11.0
  mvn:commons-fileupload/commons-fileupload/2.0-SNAPSHOT
  mvn:org.apache.commons/commons-collections4/4.4
  mvn:org.danekja/jdk-serializable-functional/1.9.0
  mvn:org.apache.wicket/wicket-util/10.0.0-M1-SNAPSHOT
  mvn:org.apache.wicket/wicket-request/10.0.0-M1-SNAPSHOT
  mvn:org.apache.wicket/wicket-core/10.0.0-M1-SNAPSHOT
  mvn:org.apache.wicket/wicket-auth-roles/10.0.0-M1-SNAPSHOT
  mvn:org.apache.wicket/wicket-devutils/10.0.0-M1-SNAPSHOT
  mvn:org.apache.wicket/wicket-extensions/10.0.0-M1-SNAPSHOT
  mvn:org.apache.wicket/wicket-jmx/10.0.0-M1-SNAPSHOT
Feature has no conditionals.
```

wicket/src/main/filtered-resources/features.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
martin-g, line 1: Could we add some XML comments here about the purpose of this file ?
From the PR comments I understand it is Apache Karaf thingy, but it would be nice to have some documentation, maybe with a link to Karaf docs about features.

martin-g (approved):

mattrpav:
Updated user guide w/ Apache Karaf feature info

wicket-feature-docs

martin-g (approved):

PR #541, opened 2022-10-06
WICKET-7008 toString() should reflect the actual variable name

tenbeitelh:


bitstorm:
@tenbeitelh thank you!

tenbeitelh:
@bitstorm with pleasure

PR #546, opened 2022-10-26
WICKET-7011 Swap assertion arguments to conform to expected vs actual

timtebeek:
https://issues.apache.org/jira/browse/WICKET-7011
https://github.com/openrewrite/rewrite-testing-frameworks/pull/277

### DIFF
--- a/wicket-extensions/src/test/java/org/apache/wicket/extensions/markup/html/repeater/data/table/DataTableTest.java
+++ b/wicket-extensions/src/test/java/org/apache/wicket/extensions/markup/html/repeater/data/table/DataTableTest.java
@@ -66,15 +66,15 @@ public class DataTableTest extends WicketTestCase
 		int index = document.indexOf("<thead");
 		assertTrue(index != -1, "Expected at least on <thead>");
 		index = document.indexOf("<thead", index + 1);
-		assertEquals(index, -1, "There must be only one <thead>");
+		assertEquals(-1, index, "There must be only one <thead>");
 
 		index = document.indexOf("<tbody");
 		assertTrue(index != -1, "Expected at least on <tbody>");
 		index = document.indexOf("<tbody", index + 1);
-		assertEquals(index, -1, "There must be only one <tbody>");
+		assertEquals(-1, index, "There must be only one <tbody>");
 
 		index = document.indexOf("<caption", index + 1);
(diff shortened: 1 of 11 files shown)


timtebeek:
Might have found an issue in `AppendingStringBuffer`; [the equals method](https://github.com/apache/wicket/blob/master/wicket-util/src/main/java/org/apache/wicket/util/string/AppendingStringBuffer.java#L1659) does not return the same when comparing String with AppendingStringBuffer, as when you compare AppendingStringBuffer with String.

When you change:
```diff
diff --git a/wicket-util/src/test/java/org/apache/wicket/util/string/AppendingStringBufferTest.java b/wicket-util/src/test/java/org/apache/wicket/util/string/AppendingStringBufferTest.java
index 7eedb46fed..e1ac4b0883 100644
--- a/wicket-util/src/test/java/org/apache/wicket/util/string/AppendingStringBufferTest.java
+++ b/wicket-util/src/test/java/org/apache/wicket/util/string/AppendingStringBufferTest.java
@@ -67,7 +67,7 @@ public class AppendingStringBufferTest
                AppendingStringBuffer asb = new AppendingStringBuffer("123456789");
                StringBuilder sb = new StringBuilder("123456789");
                assertEquals(asb, sb);
-               assertEquals(asb, "123456789");
+               assertEquals("123456789", asb);
 
                sb = new StringBuilder("01234567890");
                assertNotEquals(asb, sb);
```

You get:
```
[ERROR] Failures: 
[ERROR]   AppendingStringBufferTest.equalsToCharSequence:70
  expected: java.lang.String@52cb4f50<123456789>
  but was: org.apache.wicket.util.string.AppendingStringBuffer@25a5c7db<123456789>
```
Might be worth creating a separate issue for.

martin-g (approved):

martin-g:
> Might be worth creating a separate issue for.

Is it possible to do anything about it at our side ? 
AppendingStringBuffer#equals() has logic to compare against any `CharSequence`.
`java.lang.String` (Java 17) does not have such logic.
Maybe we should open an issue at OpenJDK ?!

martin-g:
I'm fine with this solution!

timtebeek:
Thanks for the quick review! I've kept the `AppendingStringBufferTest` changes out of this PR for now, to keep things simple.

PR #547, opened 2022-11-02
InSessionPageStore throws IndexOutOfBoundsException

renoth:
I have the same problem like in https://issues.apache.org/jira/browse/WICKET-6966 when using Wicket 9.11.0 and all Settings in SessionPageStore at default values. The problem does only occur in DEVELOPMENT environment so far, but the settings regarding SessionStorage are the same in DEV and PROD.

Can we avoid this exception centrally by checking if the pages list is empty before trying an iterator?
Logging could be added to inform the user about an inconsistent state.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/pageStore/InSessionPageStore.java
+++ b/wicket-core/src/main/java/org/apache/wicket/pageStore/InSessionPageStore.java
@@ -24,8 +24,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.function.Supplier;
 
-import jakarta.servlet.http.HttpSession;
-
 import org.apache.wicket.MetaDataKey;
 import org.apache.wicket.Session;
 import org.apache.wicket.WicketRuntimeException;
@@ -34,6 +32,8 @@ import org.apache.wicket.serialize.ISerializer;
 import org.apache.wicket.util.lang.Args;
 import org.apache.wicket.util.lang.Bytes;
 import org.apache.wicket.util.lang.Classes;
+
+import jakarta.servlet.http.HttpSession;
 
 /**
  * A store keeping a configurable maximum of pages in the session.
@@ -216,6 +216,11 @@ public class InSessionPageStore implements IPageStore
 
 		public synchronized IManageablePage remove(int pageId)
 		{
+			if (pages == null || pages.size() == 0) {
+				// Logging?
+				return null;
+			}
+
 			Iterator<IManageablePage> iterator = pages.iterator();
 			while (iterator.hasNext())
 			{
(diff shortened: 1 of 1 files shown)


martin-g:
Calling `List.iterator()` on an empty list won't lead to IndexOutOfBoundsException.
Is the list is `null` then the exception would be a `NullPointerException`.

Please create a new ticket in JIRA and paste the exception stack trace.

renoth:
yes you are right, The size should not be negative.

the stacktrace is 

```
java.lang.IndexOutOfBoundsException: Index: 0, Size: -1
    at java.util.LinkedList.checkPositionIndex(LinkedList.java:564) ~[?:?]
    at java.util.LinkedList.listIterator(LinkedList.java:871) ~[?:?]
    at java.util.AbstractList.listIterator(AbstractList.java:311) ~[?:?]
    at java.util.AbstractSequentialList.iterator(AbstractSequentialList.java:238) ~[?:?]
    at org.apache.wicket.pageStore.InSessionPageStore$SessionData.remove(InSessionPageStore.java:219) ~[wicket-core-9.11.0.jar:9.11.0]
    at org.apache.wicket.pageStore.InSessionPageStore$SessionData.add(InSessionPageStore.java:205) ~[wicket-core-9.11.0.jar:9.11.0]
    at org.apache.wicket.pageStore.InSessionPageStore$CountLimitedData.add(InSessionPageStore.java:313) ~[wicket-core-9.11.0.jar:9.11.0]
    at org.apache.wicket.pageStore.InSessionPageStore.addPage(InSessionPageStore.java:127) ~[wicket-core-9.11.0.jar:9.11.0]
    at org.apache.wicket.pageStore.CachingPageStore.addPage(CachingPageStore.java:71) ~[wicket-core-9.11.0.jar:9.11.0]
    at org.apache.wicket.pageStore.RequestPageStore.detach(RequestPageStore.java:114) ~[wicket-core-9.11.0.jar:9.11.0]
    at org.apache.wicket.page.PageManager.detach(PageManager.java:91) ~[wicket-core-9.11.0.jar:9.11.0]
    at org.apache.wicket.Application$2.onDetach(Application.java:1582) ~[wicket-core-9.11.0.jar:9.11.0]
```

martin-g:
According to https://stackoverflow.com/a/60976421/497381 it should be caused by concurrent removals.
But all methods which manipulate `pages` are `synchronized` ...
Maybe [#writeObject()](https://github.com/apache/wicket/blob/acc979b3bae398bc695b7558f9cb49c73a818867/wicket-core/src/main/java/org/apache/wicket/pageStore/InSessionPageStore.java#L264) breaks it somehow ...

renoth:
The error only occurs in our DEVELOPMENT staging environment but not in TEST or PRODUCTION, maybe that has something to do with it. Debugbar maybe?

PR #550, opened 2022-12-25
[WICKET-7021] Update OSGi imports for needed packages

mattrpav:


bitstorm (approved):
I'm not much into OSGi stuff but looks good to me

bitstorm:
@mattrpav do you think we might need something similar for master branch with new  Jakarta EE packages?

mattrpav:
> @mattrpav do you think we might need something similar for master branch with new Jakarta EE packages?

Yep. The packages should map 1:1 (if not too much has changed in Wicket 10)

I’ll try to take a look this week. I don’t have an OSGi+Jakarta runtime to test yet. That work is underway in the upstream projects (karaf and pax-web)

PR #564, opened 2023-03-24
WICKET-7036: MarkupFactory supports IMarkupFilters

hosea:
Wicket-7036: Support IMarkupFilters within MarkupFactory.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/MarkupFactory.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/MarkupFactory.java
@@ -17,6 +17,9 @@
 package org.apache.wicket.markup;
 
 import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.wicket.Application;
 import org.apache.wicket.MarkupContainer;
@@ -60,6 +63,8 @@ public class MarkupFactory
 	{
 		return Application.get().getMarkupSettings().getMarkupFactory();
 	}
+
+	private final LinkedHashMap<IMarkupFilter, Class<? extends IMarkupFilter>> additionalMarkupFilters = new LinkedHashMap<>();
 
 	/**
 	 * Construct.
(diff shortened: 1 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/MarkupFactory.java
```diff
@@ -61,6 +64,8 @@ public static MarkupFactory get()
 		return Application.get().getMarkupSettings().getMarkupFactory();
 	}
 
+	private final LinkedHashMap<IMarkupFilter, Class<? extends IMarkupFilter>> additionalMarkupFilters = new LinkedHashMap<>();
```
solomax, line 67: IMO it would be better to change `private final LinkedHashMap` -> `private final Map`

theigl:
Is this really necessary?

The Javadoc for `MarkupFactory` reads:

```java
* MarkupParser's can be extended by means of {@link IMarkupFilter}. You can add your own filter
* as follows:
* 
* <pre>
*    public MyMarkupFactory {
*      ...
*      public MarkupParser newMarkupParser(final MarkupResourceStream resource) {
*         MarkupParser parser = super.newMarkupParser(resource);
*         parser.add(new MyFilter());
*         return parser;
*      }
*    }
* </pre>
```

martin-g:
I agree with @theigl that one could easily extend MarkupFactory and achieve the requirement.

hosea:
Hi @theigl , @martin-g ,

I agree with you, that one could easily extend MarkupFactory and achieve the requirement. That's in fact, what we do now. I just want to explain, why it would be helpful to have such extensions:
I have about 20 Wicket-Applications developed by several teams sharing several libraries of wicket-components. These "central" libraries use IInitializers. So the code #Application.getMarkupSettings().setMarkupFactory(new AriaAwareMarkupFactory()) is now located in one of the libraries. But if I want to add processing of MarkupFilters in another library, then I cannot use this initialization-code anymore: two Initializers setting a new MarkupFacatory => only one wins and the other functionality is lost. This means, that all 20 Applications have to implement a new MarkupFactory that implements both functionalities, or I have to add a dependency between libraries. Both solutions are not nice.
Adding to possibility to add IMarkupFilters without replacing the instance of the MarkupFactory makes initialization much easier in such a case.

martin-g:
Thanks for the explanation, @hosea !

I can see how the code in this PR would help for your case! 
But no one ever asked for this so far and it looks like a rather unique case.

How about adding these new methods to `AriaAwareMarkupFactory` or to a new class, e.g. `CustomMarkupFactory`, as parent of AriaAwareMF, that is part of your `wicket-components` library ? Then each library and application can check what's the type of the current MarkupFactory and either just add filters or set a new MarkupFactory.

PR #566, opened 2023-04-02
[WICKET-7037] add SameSite attribute to cookie.

reiern70:
This seems to fix the FF warning.

### DIFF
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/wicket-ajaxdownload.js
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/wicket-ajaxdownload.js
@@ -28,7 +28,7 @@
 	Wicket.AjaxDownload = {
 		initiate : function(settings) {
 			document.cookie = settings.name +
-				'=;path=/;Max-Age=0;expires=Thu, 01 Jan 1970 00:00:01 GMT';
+				'=;path=/;Max-Age=0;expires=Thu, 01 Jan 1970 00:00:01 GMT' + settings.sameSite;
 			var notifyServer = function(result) {
 				settings.attributes.ep = settings.attributes.ep || {};
 				settings.attributes.ep.result = result;
--- a/wicket-core/src/main/java/org/apache/wicket/util/cookies/CookieDefaults.java
+++ b/wicket-core/src/main/java/org/apache/wicket/util/cookies/CookieDefaults.java
@@ -28,6 +28,29 @@ public class CookieDefaults implements IClusterable
 {
 	private static final long serialVersionUID = 1L;
 
+	public enum SameSite
+	{
+		/**
+		 * Cookies will be sent in all contexts, i.e. in responses to both
+		 * first-party and cross-site requests. If SameSite=None is set,
+		 * the cookie Secure attribute must also be set (or the cookie will be blocked).
+		 */
+		None,
+
+		/**
+		 * Cookies will only be sent in a first-party context and not be sent
+		 * along with requests initiated by third party websites.
+		 */
+		Strict,
+
+		/**
+		 * Cookies are not sent on normal cross-site subrequests (for example to load
+		 * images or frames into a third party site), but are sent when a user is
+		 * navigating to the origin site (i.e., when following a link).
+		 */
+		Lax
+	}
+
 	/** Max age that the component will be persisted in seconds. */
 	private int maxAge = 3600 * 24 * 30; // 30 days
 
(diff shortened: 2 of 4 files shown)


wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/wicket-ajaxdownload.js
```diff
@@ -28,7 +28,7 @@
 	Wicket.AjaxDownload = {
 		initiate : function(settings) {
 			document.cookie = settings.name +
-				'=;path=/;Max-Age=0;expires=Thu, 01 Jan 1970 00:00:01 GMT';
+				'=;path=/;Max-Age=0;expires=Thu, 01 Jan 1970 00:00:01 GMT; SameSite=None; Secure';
```
solomax, line 31: I'm not sure regarding this change

One of use-cases of our application is "Run inside `iframe`"
This mode requires modification of `webapps/_ctx_/META-INF/context.xml`
And setting

`<CookieProcessor sameSiteCookies="None" />` (Lax by default)

So _maybe_ Servlet container can handle this?

reiern70: > I'm not sure regarding this change
> 
> One of use-cases of our application is "Run inside `iframe`" This mode requires modification of `webapps/_ctx_/META-INF/context.xml` And setting
> 
> `<CookieProcessor sameSiteCookies="None" />` (Lax by default)
> 
> So _maybe_ Servlet container can handle this?

I don't think so becase this is used by server side cookie processor like in code bellow 

image

to add the SameSite thing to server side generated cookies. Maybe the we should make this configurable by passing some paramter to the JavaScrip layer?

reiern70: Code above is from

image

martin-g: I just created https://issues.apache.org/jira/browse/WICKET-7038
Servlet API 6.x has the setter.
But this is only for Wicket 10.x

martin-g:
From the JIRA description:
```
Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. 
```

Do we want this cookie to be available to third parties ?!

solomax:
It might be required to set cookie to `none` in case wicket application opened in `iframe`
We had troubles with login inside iframe, so tomcat config was corrected

reiern70:
> 

I have no idea... Af far as I udnerestand from code this is only used in order to trigger donwload finished event.  Correct? We are not using that event in our application but we get that warning form client side when cookie is "deleted" by seeting a date in the past. Thus even if it is not inpacting us we would preffer this not to apear in borwser "logs".

Another thing that I'm wondering and I will try to investigate is if we do not have a "memory leak" with this cookie after server side is notified donwload finished.

We are using a lot AjaxDownload in our app as "normal" download causes web socket disconnections on FF

reiern70:
> It might be required to set cookie to `none` in case wicket application opened in `iframe` We had troubles with login inside iframe, so tomcat config was corrected

What changes do you want me to do? Make this configurable? Anyways we ca'nt use the setCookie because then this will depend on how user configures its application server

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/wicket-ajaxdownload.js
reiern70: A cookie is just some special header... Maybe this is teh approach we should follow here so that we can have the same at client side and server side. Because if we use setCookie then, at least in case fo tomcat, application sever will use its own machiney to write SameSite attibuture (e.g. see Rfc6265CookieProcessor)

reiern70:
> @reiern70 my only concern is to avoid breakage of currently working applications with this change :( I can't test your code using my use case right now :( I need to set-up the whole chain of applications (but don't have enough time :((( )
> 
> If you sure your change is safe - I will be happy to approve it :))

@solomax 

As it is right now it will for certain break some application. As stated before I don't see a clear way to fix this on 9.x master as Cookie class offers no setters for SameSite and this is added, at least in tomcat, via some impememntation CookieProcessor inetrface that transform cookies into a cookie header. I will see what I can do otherwise I close this PR

reiern70:
@solomax I have used the idea @martin-g used for master and partilly ported to 9.x in a way that is backward compatible (or so I hope)

martin-g (approved):

solomax (approved):

PR #567, opened 2023-04-04
[WICKET-7037] add SameSite attribute to download finished cookie

reiern70:


solomax:
@reiern70 Would it be possible to add XML configuration (like `wicket.mode`) ?

reiern70:
> @reiern70 Would it be possible to add XML configuration (like `wicket.mode`) ?



> 



> @reiern70 Would it be possible to add XML configuration (like `wicket.mode`) ?

Not following you... You meam something like

image

reiern70:
> Yes, exactly so this settings can be set without code modification and recompilation :)

Shall this be part of some application settings?

martin-g (approved):

solomax:
> Shall this be part of some application settings?

In our case init parameter of servlet context will be enough :))

martin-g:
> In our case init parameter of servlet context will be enough :))

If Wicket won't set this value into some **Settings bean then there is nothing to do here. The application can set the extra init-param in its web.xml/annotation and read it from the ServletContext.

reiern70:
> > In our case init parameter of servlet context will be enough :))
> 
> If Wicket won't set this value into some **Settings bean then there is nothing to do here. The application can set the extra init-param in its web.xml/annotation and read it from the ServletContext.

Yes... but then for all the AjaxDonwloadBehaviors you will need to:

1. Read this parameter 
2. Adapt your code to do `ajaxDonwloadBehavior.setAttribute("SameSite", "What you read from parameter");

@solomax 

You can allways roll you own AjaxDonwloadBehavior and do

```
class MyAjaxDonwloadBehavior extends AjaxDonwloadBehavior {

     MyAjaxDonwloadBehavior(IResource resource) {
       super (resource);
       setSameSite("read it from somewhere");
     }
}
```

But I stilll see the value of fixing some site without tocuhing your code (mind that for wicket 9.x you still will need to configure application server to decorate cookies).

reiern70:
@martin-g Should I add such configuration option?

reiern70:
> > You can allways roll you own AjaxDonwloadBehavior and do
> 
> Good point! Thanks :)

But I still see a valid use case for what you requested for existing applications:

1- You have browser warnings
2- You upgrade wicket + add some new system property to "your deployment"
3- redeploy and warnings are gone 

If this metits to have more configuration I don't know.

solomax:
Yes, this would be ideal
But Custom AjaxDownloadBehavior is something we can workaround with until Wicket-10 :)

reiern70:
> Yes, this would be ideal But Custom AjaxDownloadBehavior is something we can workaround with until Wicket-10 :)

For 9.x we need

https://github.com/apache/wicket/pull/566

PR #569, opened 2023-04-05
[WICKET-7028] do CSP related header generation both at onRequestHandl…

reiern70:
do CSP related header generation both at onRequestHandlerResolved and onRequestHandlerExecuted

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/coep/CrossOriginEmbedderPolicyRequestCycleListener.java
+++ b/wicket-core/src/main/java/org/apache/wicket/coep/CrossOriginEmbedderPolicyRequestCycleListener.java
@@ -36,7 +36,7 @@ import javax.servlet.http.HttpServletRequest;
  * <code>Cross-Origin-Embedder-Policy</code>. The header is not set for the paths that are exempted
  * from COEP. The only valid value of COEP is <code>require-corp</code>, so if the listener is
  * enabled the policy value will be specified as so.
- *
+ * <p>
  * COEP prevents a document from loading any non-same-origin resources which don't explicitly grant
  * the document permission to be loaded. Using COEP and COOP together allows developers to safely
  * use powerful features such as <code>SharedArrayBuffer</code>,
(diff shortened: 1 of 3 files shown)


reiern70:
@svenmeier ?

svenmeier:
Can't check ATM, but don't we have to check for duplicated headers then?

I wonder whether we should go with a completely different solution:
Aren't the headers specific to the page? We could add a behavior to the page (via Component instantiation/Initialization listener) and thus always add the header when the page is rendered.

reiern70:
> Not beautiful, but it seems to work.
> 
> Maybe let #onRequestHandlerResolved() and #onRequestHandlerExecuted() both delegate to a private method #protect() ... the way the one calls the other now I find confusing.

This sounds like a goog idea. I will do and push those changes and merge. After that I can try to explore the other idea you proposed (that seems less fragile that doing it this way). Wondering if we should also add some tests covering actual generation of headers for these corrner cases.

wicket-core/src/main/java/org/apache/wicket/coep/CrossOriginEmbedderPolicyRequestCycleListener.java
```diff
@@ -36,15 +36,15 @@
  * <code>Cross-Origin-Embedder-Policy</code>. The header is not set for the paths that are exempted
  * from COEP. The only valid value of COEP is <code>require-corp</code>, so if the listener is
  * enabled the policy value will be specified as so.
- *
+ * <p>
```
solomax, line 39: I believe, this one will produce JavaDoc warning :(

reiern70: :-( I had a warining locally... And this fixed it. DO we have some IJ settings I could use as default for wicket?

reiern70: > you are right :) no specific warning at this line, sorry for the noise :(

No problem. Everything that makes wicket code better is welcomed :-)

> 
> I'm not using IJ, I believe checkstyle and javadoc maven plugins can be configured to be more strict

Oh! I didn't check the result of those. Just that code build with maven clean install on terminal in my linux box.

At work we have the same "warnings war" because some people use IJ and other use eclipse and we use gradle instead of  maven and it is very diffcult to get all the parties agree about what a warning is :-(

PR #572, opened 2023-04-10
WICKET-7041 Introduce `LazyStringResponse` to avoid unnecessary allocations

theigl:
This PR avoids unnecessary allocation of `AppendingStringBuffer` for a `StringResponse` that is never written to.

In our case, 99% of the allocations in `Component#internalRenderHead` can be avoided in this way.

I'm not sure if we should add `LazyStringResponse` as a separate class, or simply make the existing `StringResponse` lazy. It is used in a lot of places, so we might avoid more allocations if we changed the default implementation.

See https://issues.apache.org/jira/browse/WICKET-7041

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/response/LazyStringResponse.java
+++ b/wicket-core/src/main/java/org/apache/wicket/response/LazyStringResponse.java
@@ -1,0 +1,116 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.response;
+
+import org.apache.wicket.request.Response;
+import org.apache.wicket.util.string.AppendingStringBuffer;
+
+
+/**
+ * Response object that writes to an AppendingStringBuffer. This class is functionally equivalent to
+ * {@link StringResponse}, but defers creating the buffer until it is needed.
+ * 
+ * @author Thomas Heigl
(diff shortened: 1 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/response/LazyStringResponse.java
```diff
@@ -0,0 +1,103 @@
...
+		getOut().append(string);
+	}
+
+	/**
+	 * @see Response#reset()
+	 */
+	@Override
+	public void reset()
+	{
+		getOut().clear();
```
solomax, line 51: Is it necessary to create buffer if you only need to reset it?

Maybe this code can be changed to:
`if (out != null) out.reset() ?`

If so `getOut()` method can be dropped :)

theigl: Ha, i just realized the same thing and pushed it before reading your comment ;)

wicket-core/src/main/java/org/apache/wicket/response/LazyStringResponse.java
```diff
@@ -0,0 +1,101 @@
...
+
+	/**
+	 * @see Response#write(CharSequence)
+	 */
+	@Override
+	public void write(final CharSequence string)
+	{
+		if (out == null)
+		{
+			out = new AppendingStringBuffer(128);
```
martin-g, line 44: `128` should be a constructor parameter. The default constructor should pass `128` to another ctor that accepts the initial size

wicket-core/src/main/java/org/apache/wicket/response/LazyStringResponse.java
```diff
@@ -0,0 +1,101 @@
...
+package org.apache.wicket.response;
+
+import org.apache.wicket.request.Response;
+import org.apache.wicket.util.string.AppendingStringBuffer;
+
+
+/**
+ * Response object that writes to a StringWriter. If the StringResponse is later converted to a
+ * String via toString(), the output which was written to the StringResponse will be returned as a
+ * String.
```
martin-g, line 26: Copy/pasted javadoc from StringResponse.
Please add something about its uniqueness.

wicket-core/src/main/java/org/apache/wicket/response/LazyStringResponse.java
theigl: Done

reiern70 (approved):
Nice!

bitstorm:
> > @chrisco484 I believe this one will definitely be merged to 9.x
> 
> @solomax Any earlier versions like 7 or 8 ? (Optimistic I know ;) )

If it's just (as it seems...) a cherry pick we might port this back to Wicket 8.

PR #576, opened 2023-04-14
[WICKET-7033] implementation of: 1) an upload field that allows uploa…

reiern70:
…ding files into to a mounted resource, 2) a reusable mounted resource that process uploads and 3) adapting upload progress bar to work in such use case


reiern70:
This is the correspoding for https://github.com/apache/wicket/pull/571 but to master

reiern70:
Branch name is worng but PR is correct :-)

bitstorm (approved):

PR #578, opened 2023-04-16
WICKET-7046 Avoid buffering in AjaxRequestHandler when no response filters active

theigl:
We currently buffer ajax responses twice, first wrapping them in a `StringResponse` and then in an `AppendingStringBuffer`. This whole wrapping is really only necessary when the application has one or more `IResponseFilter`. By default, Wicket applications running in deployment mode *do not have* any response filters.

This PR adjusts `AjaxRequestHandler#respond` to write the page update directly to the response if no filters are active. This should drastically reduce allocations for large ajax responses.

https://issues.apache.org/jira/browse/WICKET-7046


solomax (approved):
Lgtm :)

reiern70 (approved):
LGTM!

PR #579, opened 2023-04-16
WICKET-7047 Improve initial buffer capacity for multiline-markup

theigl:
This PR initializes the string buffer for Strings#toMultilineMarkup to a capacity that should avoid resizing the buffer in most cases.

https://issues.apache.org/jira/browse/WICKET-7047

### DIFF
--- a/wicket-util/src/main/java/org/apache/wicket/util/string/Strings.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/string/Strings.java
@@ -1088,12 +1088,15 @@ public final class Strings
 		{
 			return null;
 		}
-
-		final AppendingStringBuffer buffer = new AppendingStringBuffer();
+		
+		final int len = s.length();
+		
+		// allocate a buffer that is 10% larger than the original string to account for markup
+		final AppendingStringBuffer buffer = new AppendingStringBuffer((int) (len * 1.1) + 16);
 		int newlineCount = 0;
 
 		buffer.append("<p>");
-		for (int i = 0; i < s.length(); i++)
+		for (int i = 0; i < len; i++)
 		{
 			final char c = s.charAt(i);
 


wicket-util/src/main/java/org/apache/wicket/util/string/Strings.java
```diff
@@ -1089,7 +1089,7 @@ public static CharSequence toMultilineMarkup(final CharSequence s)
 			return null;
 		}
 
-		final AppendingStringBuffer buffer = new AppendingStringBuffer();
+		final AppendingStringBuffer buffer = new AppendingStringBuffer((int) (s.length() * 1.1));
```
solomax, line 1092: This looks like black magic :)
Maybe comment and/or meaningful constant name should be added? :)

theigl: We have the exact same logic in `Strings.escapeMarkup`:

https://github.com/apache/wicket/blob/d51883862a448e26f8d973a24756ea7e03259de7/wicket-util/src/main/java/org/apache/wicket/util/string/Strings.java#L305

A constant would have to be something like `ESTIMATED_OVERHEAD_OF_MULTILINE_MARKUP`. I think a comment would be better.

reiern70: Would it make sense to have this "magic" in some setting?

theigl: I don't think so. The additional 10% should be enough to accommodate markup and escaping overhead in most cases. And if it isn't enough, the buffer is resized once. This is *much* better than resizing the buffer almost every time as it is now. Imho there is no need to configure or optimize this further.

reiern70: > I don't think so. The additional 10% should be enough to accommodate markup and escaping overhead in most cases. And if it isn't enough, the buffer is resized once. This is _much_ better than resizing the buffer almost every time as it is now. Imho there is no need to configure or optimize this further.

Yes... I after I wrote my comment I looked at soucre code and I realized the 10% more should be enough. Maybe what we can have is a TenPercentMoreAppendingStringBuffer class?

reiern70 (approved):
LGTM!

wicket-util/src/main/java/org/apache/wicket/util/string/Strings.java
theigl: I just realized that this has an adverse effect on short strings where the overhead of `<p></p>` is significantly more than 10%. It makes sense to set a minimum capacity for the buffer. Either via `Math.max((int)(len * 1.1), 16)` or by always adding the default capacity of 16 to the length.

solomax (approved):
I would add `+ 16` to comment :)

martin-g (approved):

PR #583, opened 2023-04-18
OSGI import fixes

akramshuja:
Added additional missing OSGI imports after comparing with 9.11.0 Manifest.mf.

Found the issue after logging into our application as we encrypt our url. It resulted in ClassNotFoundException for javax.crypto.*

### DIFF
--- a/wicket-core/pom.xml
+++ b/wicket-core/pom.xml
@@ -143,7 +143,7 @@ org.apache.wicket.util.tester;-noimport:=true,
 org.apache.wicket.validation;-noimport:=true,
 org.apache.wicket.validation.validator;-noimport:=true
  		</osgi.export.package> 
-                <osgi.import.package>!java*,!kotlin*,!sun.nio.ch,!com.sun.crypto.provider,org.slf4j*;version="[1.7,3)",javax.servlet,javax.servlet.http,*</osgi.import.package>
+                <osgi.import.package>!java*,!kotlin*,!sun.nio.ch,!com.sun.crypto.provider,org.slf4j*;version="[1.7,3)",javax.servlet,javax.servlet.http,javax.crypto,javax.crypto.spec,javax.imageio,javax.security.auth,javax.xml.transform,javax.xml.transform.stream,*</osgi.import.package>
         </properties>
 	<dependencies>
 		<dependency>
--- a/wicket-util/pom.xml
+++ b/wicket-util/pom.xml
@@ -29,7 +29,7 @@
 
 	<properties>
 		<osgi.export.package>org.apache.wicket.util*;-noimport:=true</osgi.export.package>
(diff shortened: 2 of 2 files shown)


martin-g:
@mattrpav Could you please take a look at the changes when you have a second ? Thanks!

mattrpav:
@akramshuja where or how does *wicket* use these imports? If your app uses wicket and these imports, your app needs to import them, not the wicket libraries.

akramshuja:
1. We use Apache Felix
2. We are on Java 11

mattrpav:
> 1. We use Apache Felix
> 2. We are on Java 11

How does Apache Felix configure the list of packages from JDK to register to all bundles? ie. bundle0? 

In Apache Karaf, there is an etc/jre.properties that lists the packages to export from the JDK.

akramshuja:
Not an OSGI expert here, but
"In Apache Felix, the boot delegation list is used to configure the list of packages from the JDK that should be registered to all bundles, including bundle0.

The boot delegation list is a comma-separated list of package names that are delegated to the parent class loader, which is typically the class loader of the JVM. When a bundle requests a class that is not found in its own class loader, the OSGi framework searches the boot delegation list for the package containing that class."

mattrpav:
@akramshuja you would change the config value for 'org.osgi.framework.bootdelegation' in your Felix runtime to expose the packages you added to this PR, vs updating the Wicket bundles.

akramshuja:
No, I would not make any changes to 'org.osgi.framework.bootdelegation'. We maintain export.packages that contains a List of packages to be exported by the OSGI framework that will be made available to all bundles for import. If adding there solves our purpose, I think we are good to go.

akramshuja:
Just verified it. Doesn't work that way. I think it was the same case with javax.servlet which has been addressed in one of the pull requests.(javax.servlet is required by WicketFilter, although we had this in our export.packages, the Tomcat startup would still throw error mentioning java.servlet not found by wicket.core).

mattrpav:
javax.servlet is different from the others. javax.servlet was never part of the JDK. These others (crypto, xml, etc) were or are still part of the JDK.

akramshuja:
Verified by adding the packages to org.osgi.framework.bootdelegation property in our app, it works now.
Having said that, I would also like to bring your attention to org.wicketstuff:wicket-datetime": "9.13.0". I see its manifest is having - **java.io, java.lang, java.lang.invoke, java.text, java.util, java.util.function** under import-packages that is causing the bundle to fail during startup.
These were not there until 9.12.0

martin-g:
Shall we close this PR ?
There is nothing to do, right ?

martin-g:
About WicketStuff: there were no OSGi related changes in https://github.com/wicketstuff/core/commits/wicketstuff-core-9.13.0/wicket-datetime-parent/wicket-datetime nor in the parent poms.
If there is such need then please open a PR ! Thank you in advance!

akramshuja:
Yes, this can be closed.

martin-g:
Thank you both!

PR #587, opened 2023-04-26
WICKET-6966, WICKET-7013: prevent unsynchronized access to pages

dr0ps:
InSessionPageStore.SessionData#pages is a LinkedList. LinkedList is not thread-safe and keeps a rather complex state as it has to keep track of first and last element and the list size. The combination leads to some fragility. For instance, if multiple threads concurrently remove the same element, both operations will succeed (and remove the same element) but the individually stored list size will get reduced twice. In the case of WICKET-6966 and WICKET-7013 this has happened. The exception message states that the current list size is -1. 
InSessionPageStore.SessionData#pages is only accessed in InSessionPageStore.SessionData itself and InSessionPageStore.CountLimitedData. Most accesses are in synchronized instance methods. Unfortunately InSessionPageStore.SessionData#writeObject(OutputStream) contains multiple accesses of the pages list without any synchronization. Therefore should serialization occur while e.G. new page instances are being created this might corrupt the state of the LinkedList. 
This merge request removes the last unsynchronized calls.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/pageStore/InSessionPageStore.java
+++ b/wicket-core/src/main/java/org/apache/wicket/pageStore/InSessionPageStore.java
@@ -261,7 +261,7 @@ public class InSessionPageStore implements IPageStore
 		/**
 		 * Serialize pages before writing to output.
 		 */
-		private void writeObject(final ObjectOutputStream output) throws IOException
+		private synchronized void writeObject(final ObjectOutputStream output) throws IOException
 		{
 			// handle non-serialized pages
 			for (int p = 0; p < pages.size(); p++)
--- a/wicket-core/src/test/java/org/apache/wicket/pageStore/InSessionPageStoreConcurrencyTest.java
+++ b/wicket-core/src/test/java/org/apache/wicket/pageStore/InSessionPageStoreConcurrencyTest.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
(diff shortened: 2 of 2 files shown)


dr0ps:
It would be possible to prevent the list corruption by switching it to a synchronized collection. However, writeObject does first get the list size, then it for-loops over the list (btw an expensive operation on a LinkedList) and it individually removes the pages. It is not hard to imagine that the pages list might end up in all kinds of states if concurrent changes can occur.

theigl:
Since all other methods in the class are already synchronized, I think your approach to fixing this is correct.

It would be great to have a test for this, but I don't know how hard it is to simulate the failure in a multi-threaded test.

dr0ps:
I added a test case that pretty reliably fails without synchronization. I would not make this test part of the Wicket distribution but it helps showing the issue.

renoth:
You could try to make it 100% reliable with executorservice and Threadpool like here for a similar Testcase:
https://github.com/apache/wicket/blob/5d46ca58cc835b2ac925c14190260c6253db2c3e/wicket-spring/src/test/java/org/apache/wicket/spring/injection/bytebuddy/ParallelInjectionTest.java#L115

dr0ps:
> You could try to make it 100% reliable with executorservice and Threadpool like here for a similar Testcase:
> 
> https://github.com/apache/wicket/blob/5d46ca58cc835b2ac925c14190260c6253db2c3e/wicket-spring/src/test/java/org/apache/wicket/spring/injection/bytebuddy/ParallelInjectionTest.java#L115

That is just as reliable as my test. AbstractExecutorService#invokeAll uses a loop to execute each tests. In no way it guarantees parallel execution as in the end this is very much depending on the underlying OS and hardware. So for consistency reasons I think my test should look like the ParallelInjectionTest. But both tests do not actually prove the absence of a problem.

renoth:
If it does not improve the reliability, disregard my comment, i was thinking the code would look cleaner that way, nevermind.

theigl (approved):

martin-g (approved):

martin-g:
Thank you, @dr0ps !

PR #591, opened 2023-05-31
Dependencies are excluded from final jar

solomax:


wicket-commons-fileupload/pom.xml
```diff
@@ -317,13 +317,9 @@
...
         <artifactId>maven-bundle-plugin</artifactId>
-        <dependencies>
-          <dependency>
-            <groupId>biz.aQute.bnd</groupId>
-            <artifactId>biz.aQute.bndlib</artifactId>
-            <version>6.4.0</version>
-          </dependency>
-        </dependencies>
+        <configuration>
+          <excludeDependencies>true</excludeDependencies>
```
martin-g, line 321: This looks like something we should do in the `dependencyManagement` in the parent pom.
I wonder why this didn't cause issues with the other `wicket-` Maven modules before ?!

solomax: To be fair: I have no idea :(

I never use bundle plugin :(

solomax: This is the original maven output:

```
INFO] --- bundle:5.1.8:bundle (default-bundle) @ wicket-commons-fileupload ---
[WARNING] Parameter 'localRepository' is deprecated core expression; Avoid use of ArtifactRepository type. If you need access to local repository, switch to '${repositorySystemSession}' expression and get LRM from it instead.
[WARNING] Bundle org.apache.wicket:wicket-commons-fileupload:jar:10.0.0-M1-SNAPSHOT : Split package, multiple jars provide the same package:
Use Import/Export Package directive -split-package:=(merge-first|merge-last|error|first) to get rid of this warning
Package found in   [Jar:., Jar:jakarta.servlet-api]
Class path         [Jar:., Jar:commons-io, Jar:jakarta.servlet-api, Jar:slf4j-api]
[INFO] Building bundle: /home/solomax/work/wicket/wicket/wicket-commons-fileupload/target/wicket-commons-fileupload-10.0.0-M1-SNAPSHOT.jar
[INFO] Writing manifest: /home/solomax/work/wicket/wicket/wicket-commons-fileupload/target/classes/META-INF/MANIFEST.MF
[INFO] 

```

martin-g: OSGi again ...
@mattrpav Do you have any recommendations ?

mattrpav: I’ll check

mattrpav (commented):
This looks good, we should not override the version of bndlib used by maven-bundle-plugin

PR #592, opened 2023-06-05
Improvement/ernestosemedt/wicket 7059

reiern70:


theigl:
In my project, we usually do this when we don't want a page to be serialized after an ajax request:

```java
private void untouchPage(Page page) {
  if (Session.exists()) {
     // Do not re-serialize page
     Session.get().getPageManager().untouchPage(page);
  }
}
```

Did you consider this approach as well?

reiern70:
> In my project, we usually do this when we don't want a page to be serialized after an ajax request:
> 
> ```java
> private void untouchPage(Page page) {
>   if (Session.exists()) {
>      // Do not re-serialize page
>      Session.get().getPageManager().untouchPage(page);
>   }
> }
> ```
> 
> Did you consider this approach as well?

No. You do this after each AJAX request?

I just want to be able to mark some requests as "do not generate" a page serialization.

theigl:
> No. You do this after each AJAX request?
>
> I just want to be able to mark some requests as "do not generate" a page serialization.

Yes, after each ajax request that does not change the stage of the page.

PR #594, opened 2023-06-29
Update Application_cs.properties

vrozkovec:
Replace incorrect "greater than" with "at least"

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Application_cs.properties
+++ b/wicket-core/src/main/java/org/apache/wicket/Application_cs.properties
@@ -16,7 +16,7 @@ Required=pole '${label}' je povinn\u00e9.
 IConverter='${label}' nen\u00ed validn\u00ed ${type}.
 
 RangeValidator.range=${label} mus\u00ed b\u00fdt mezi ${minimum} a ${maximum}.
-RangeValidator.minimum='${label}' mus\u00ed b\u00fdt v\u011bt\u0161\u00ed ne\u017e ${minimum}.
+RangeValidator.minimum='${label}' mus\u00ed b\u00fdt alespo\u0148 ${minimum}.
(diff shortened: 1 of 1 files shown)


reiern70 (approved):
Approving this is an act of faith :-)

solomax:
I would convert this file to XML format, so edit would be less painful :))

martin-g (approved):

solomax (approved):

martin-g:
> Approving this is an act of faith :-)

One could use https://examples9x.wicket.apache.org/unicodeconverter/ to see the real letters.

martin-g:
Do we need a JIRA ticket too ?
The changes are really small, so I am OK without ticket.

vrozkovec:
I can convert the file to Application_cs.utf8.properties, should I make another PR? Or is there a reason why you keep using encoded version of .properties?
If needed/wanted, I could convert all .properties to .utf8.properties.

martin-g:
+1 to convert all .properties to .utf8.properties !
But please create a ticket in JIRA for the changelog!

vrozkovec:
Created: https://issues.apache.org/jira/browse/WICKET-7063
In what branch should I do the change? Will you cherry-pick it to other branches? 
And should I also rename all files to contain .utf8. in the filename, even if there are no encoded entities in the file, so it is future-proof?

martin-g:
> In what branch should I do the change?

`master`

> Will you cherry-pick it to other branches?

If requested - yes!


> And should I also rename all files to contain .utf8. in the filename, even if there are no encoded entities in the file, so it is future-proof?

Yes, please!

vrozkovec:
@martin-g done: #596 
PR #596 contain also identical factual fix of Czech `RangeValidator.minimum` resource key as this PR.

martin-g:
Fixed with #596

PR #599, opened 2023-08-17
[WICKET-7067] WebSocketResponse does not allow setting a header thus,…

reiern70:
WebSocketResponse does not allow setting a header thus,  disableCaching fails with an UnsupportedOperationException

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/DefaultExceptionMapper.java
+++ b/wicket-core/src/main/java/org/apache/wicket/DefaultExceptionMapper.java
@@ -54,10 +54,11 @@ public class DefaultExceptionMapper implements IExceptionMapper
 		try
 		{
 			Response response = RequestCycle.get().getResponse();
-			if (response instanceof WebResponse)
+			// WICKET-7067 WebSocketResponse doesn't allow setting headers
+			if (response instanceof WebResponse && ((WebResponse) response).isHeaderSupported())
 			{
 				// we don't want to cache an exceptional reply in the browser
-				((WebResponse)response).disableCaching();
+				((WebResponse) response).disableCaching();
 			}
 			return internalMap(e);
 		}


wicket-core/src/main/java/org/apache/wicket/DefaultExceptionMapper.java
```diff
@@ -56,8 +56,15 @@ public IRequestHandler map(Exception e)
 			Response response = RequestCycle.get().getResponse();
 			if (response instanceof WebResponse)
 			{
-				// we don't want to cache an exceptional reply in the browser
-				((WebResponse)response).disableCaching();
+				try
```
reiern70, line 59: @martin-g I'm not sure thsi is the best way to fix this. Other way would be have silent implementations in 

image

But that  sound not correct.

martin-g: IMO a better way would be to check with [WebResponse.isHeaderSupported()](https://github.com/apache/wicket/blob/a15fbea9bea4e40684997f151e8f8957e0e33e3f/wicket-request/src/main/java/org/apache/wicket/request/http/WebResponse.java#L63) before calling `disableCaching()`.

reiern70: > IMO a better way would be to check with [WebResponse.isHeaderSupported()](https://github.com/apache/wicket/blob/a15fbea9bea4e40684997f151e8f8957e0e33e3f/wicket-request/src/main/java/org/apache/wicket/request/http/WebResponse.java#L63) before calling `disableCaching()`.

Oh! That is way better. I somehow missed this method!

martin-g (approved):

reiern70:
@martin-g Many thanks for review! Merging and cherrypicking to master

PR #600, opened 2023-08-19
[WICKET-7068] Tree styles are updated to be RTL-friendly

solomax:
This one can be tested at https://examples9x.wicket.apache.org/tree by setting `dir="rtl"` attribute for `<html>` tag

I'll create JIRA if this PR will be considered useful :)


reiern70:
@solomax Maybe create a second example page with dir="rtl"

solomax:
I'm not sure what would be the best way in this case :(

It might be some sort of dropdown (so every page can be checked in RTL), but this might be useless due to all contents is in English ....

martin-g (approved):
It looks OK to me but I am not the best person to review CSS code!

martin-g:
@solomax Should this be ported to `master` ?

solomax:
> @solomax Should this be ported to `master` ?

Unfortunately while cherry-picking and testing I found minified CSS is broken :(

please see https://github.com/apache/wicket/pull/602

This one should be additionally fixed by merging #602 or reverted .... :((

PR #601, opened 2023-08-24
[WICKET-7069] Refactor `assertTrue(equals())` using `assertEquals`

Taher-Ghaleb:
I am working on research that investigates test smell refactoring in which we identify alternative implementations of test cases, study how commonly used these refactorings are, and assess how acceptable they are in practice.

The first smell is when inappropriate assertions are used, while there exist better alternatives. For example, in [throws Exception using assertThrows](https://github.com/apache/wicket/commit/d667bafbb8081b87fc6dc764f7fc895b066b96d7), I refactored `assertTrue(img.getName().equals("img"));` using `assertEquals(img.getName(), "img");`.

The second smell is when exception handling can alternatively be implemented using assertion rather than annotation. For example, in [ApplicationSettingsTest.java](https://github.com/apache/wicket/commit/d667bafbb8081b87fc6dc764f7fc895b066b96d7#diff-33c5fca8a9e134cb8422f5a47c446cc03ecdb043fcc8c70a780ee69c0aba789e), I used `assertThrows(Exception.class, () -> {...});` instead of `throws Exception`.

While there could be several cases like this, we aim in this pull request to get your feedback on these particular test smells and their refactorings. Thanks in advance for your input.

### DIFF
--- a/wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
+++ b/wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
@@ -133,12 +133,12 @@ final class MarkupParserTest extends WicketTestCase
 		log.info("tok(4)=" + tokens.get(4));
 		log.info("tok(5)=" + tokens.get(5));
 
-		assertEquals("This is a test ", tokens.get(0));
+		assertEquals("This is a test ", tokens.get(0).toString());
 
 		final ComponentTag a = (ComponentTag)tokens.get(1);
 
 		assertEquals(9, a.getAttributes().getInt("componentName:id"));
-		assertEquals(" <b>bold</b> ", tokens.get(2));
+		assertEquals(" <b>bold</b> ", tokens.get(2).toString());
 
 		final ComponentTag b = (ComponentTag)tokens.get(3);
 
@@ -147,7 +147,7 @@ final class MarkupParserTest extends WicketTestCase
 		final ComponentTag closeA = (ComponentTag)tokens.get(5);
 
 		assertEquals("a", closeA.getName());
-		assertEquals(" of the emergency broadcasting system", tokens.get(6));
+		assertEquals(" of the emergency broadcasting system", tokens.get(6).toString());
 	}
 
 	/**
--- a/wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
+++ b/wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
@@ -67,7 +67,7 @@ final class MarkupParserTest extends WicketTestCase
 		final ComponentTag aOpen = (ComponentTag)markupStream.next();
 
 		log.info("", aOpen);
-		assertTrue(aOpen.getName().equals("a"));
+		assertEquals(aOpen.getName(), "a");
 		assertEquals("foo.html", aOpen.getAttributes().getString("href"));
 
 		markupStream.next();
@@ -75,7 +75,7 @@ final class MarkupParserTest extends WicketTestCase
 		final ComponentTag boldOpen = (ComponentTag)markupStream.next();
 
 		log.info("", boldOpen);
-		assertTrue(boldOpen.getName().equals("b"));
+		assertEquals(boldOpen.getName(), "b");
 		assertEquals(TagType.OPEN, boldOpen.getType());
 
 		markupStream.next();
@@ -83,7 +83,7 @@ final class MarkupParserTest extends WicketTestCase
 		final ComponentTag boldClose = (ComponentTag)markupStream.next();
 
 		log.info("", boldClose);
-		assertTrue(boldClose.getName().equals("b"));
+		assertEquals(boldClose.getName(), "b");
 		assertEquals(TagType.CLOSE, boldClose.getType());
 
 		markupStream.next();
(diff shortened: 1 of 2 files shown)


solomax:
Hello @Taher-Ghaleb 

please revert your commit labeled `[Refactor throws Exception using assertThrows](https://github.com/apache/wicket/pull/601/commits/d667bafbb8081b87fc6dc764f7fc895b066b96d7)`

It breaks the build, because:

If tests is defined as follows:
```
@Test
	void testExceptionOnMissingResourceDefaultValue() throws Exception
```

It doesn't mean it expects the `Exception` to be thrown
It only states some method **might** throw `Exception` so we should add `try-catch` OR `throws Exception` to satisfy compiler :)

wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
```diff
@@ -67,31 +67,31 @@ void tagParsing() throws Exception
 		final ComponentTag aOpen = (ComponentTag)markupStream.next();
 
 		log.info("", aOpen);
-		assertTrue(aOpen.getName().equals("a"));
+		assertEquals(aOpen.getName(), "a");
```
solomax, line 70: These are valid changes :))))

solomax (changes requested):

Taher-Ghaleb:
Thanks @solomax for your feedback.

I must have misunderstood the actual objective of these cases. I will revert that commit shortly.

In our research, we have actually encountered test cases where developers refactor exception handling operations using the `assertThrows` assertion (e.g., instead of using `@Test(expected)`). Therefore, we wanted to understand whether such refactoring is commonly acceptable in practice by developers and whether there are certain cases where it is not preferred by submitting such pull requests, but I mistakenly chose test cases with different behavior. Thanks.

wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
Taher-Ghaleb: What is your opinion about this refactoring, please? Why do you think there are many cases like `assertTrue(x.equals(y))` and `assertFalse(x.equals(y))` or `assertTrue(!x.equals(y))`, though `assertEquals` and `assertNotEquals` are more readable/intuitive and can provide better error messages in case of failures?

solomax:
> where developers refactor exception handling operations using the assertThrows assertion (e.g., instead of using @Test(expected))

AFAIK `@Test(expected)` is **JUnit-4** way of checking exception was thrown, while `assertThrows` is **JUnit-5** way :)))

for ex. see here: https://www.baeldung.com/junit-5-migration#1-annotations

wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
solomax: I can't say for sure

I guess someone just overlooked _the right way_ :)))

Taher-Ghaleb:
Should I change the title and description of the pull request to reflect the eventual changes performed?

wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
```diff
@@ -67,31 +67,31 @@ void tagParsing() throws Exception
...
 		assertEquals("foo.html", aOpen.getAttributes().getString("href"));
 
 		markupStream.next();
 
 		final ComponentTag boldOpen = (ComponentTag)markupStream.next();
 
 		log.info("", boldOpen);
-		assertTrue(boldOpen.getName().equals("b"));
+		assertEquals(boldOpen.getName(), "b");
```
solomax, line 78: Please do another change:

`assertEquals` defined as follows: `assertEquals(expected, actual)` so these 2 arguments need to be swapped

Please ensure the order is correct for all asserts you have changed :)

solomax (changes requested):

solomax:
> Should I change the title and description of the pull request to reflect the eventual changes performed?

Yes, please :))
Additionally JIRA issue need to be created here https://issues.apache.org/jira/projects/WICKET

I can take care of both steps :)

wicket-core/src/test/java/org/apache/wicket/markup/MarkupParserTest.java
```diff
@@ -133,12 +133,12 @@ final void test1() throws Exception
 		log.info("tok(4)=" + tokens.get(4));
 		log.info("tok(5)=" + tokens.get(5));
 
-		assertTrue(tokens.get(0).equals("This is a test "));
+		assertEquals("This is a test ", tokens.get(0));
```
solomax, line 136: `tokens.get(0)` is of type `RawMarkup` https://github.com/apache/wicket/blob/master/wicket-core/src/main/java/org/apache/wicket/markup/RawMarkup.java

so `tokens.get(0).equals(_some_string_)` works due to `equals` implementation

this assert need to be rewritten as 
`assertEquals("This is a test ", tokens.get(0).toCharSequence());`
OR
`assertEquals("This is a test ", tokens.get(0).toString());`

also other 2 asserts with `tokens.get()`

solomax (changes requested):

solomax (approved):

martin-g (approved):

solomax:
Thanks for the contribution @Taher-Ghaleb !

PR #608, opened 2023-09-11
Add dependabot config for updating Maven and NPM dependencies regularly

martin-g:


solomax:
This one produce outstanding amount of email :(

martin-g:
This is one time only. From now on it will be a separate PR for each new update, i.e. few PRs per week.

PR #652, opened 2023-09-18
WICKET-7075: Only support headers on uncommitted response

jsteenbeeke:
CrossOriginOpenerPolicyRequestCycleListener attempts to write headers to a response that is already committed. It contains logic to prevent setting of headers when a response does not support them, but the implementation for ServletWebResponse always returns true.

This PR modifies ServletWebResponse to check if the underlying HttpServletResponse has already been committed.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/protocol/http/servlet/ServletWebResponse.java
+++ b/wicket-core/src/main/java/org/apache/wicket/protocol/http/servlet/ServletWebResponse.java
@@ -91,7 +91,7 @@ public class ServletWebResponse extends WebResponse
 	@Override
 	public boolean isHeaderSupported()
 	{
-		return true;
+		return !httpServletResponse.isCommitted();
 	}
 	
 	@Override


solomax (approved):

martin-g (changes requested):
We need a JIRA ticket for the changelog

jsteenbeeke:
@martin-g certainly, though I don't believe I have an account for the ASF JIRA yet so this may take a bit

martin-g:
We/I can help with the creation of the ticket if you like!
But 9.15 has been just released so we have time until the next release :-)

jsteenbeeke:
I've only run into this problem with the milestone 1 release for version 10 when run on Jetty 12 though, so the 9.x releases are somewhat less relevant to my situation I suppose.

So thanks for the offer, but it's probably best if I create the ticket myself so I include all relevant background information.

jsteenbeeke:
I've created https://issues.apache.org/jira/browse/WICKET-7075. I'll amend the commit to include the ticket number

martin-g:
Thank you, @jsteenbeeke !

PR #653, opened 2023-09-19
WICKET-7072 Move AbstractLicenseHeaderTestCase from wicket-util src/main to src/test

martin-g:


solomax:
Maybe this can be replaced with https://creadur.apache.org/rat/apache-rat-plugin/ ?

martin-g:
> Maybe this can be replaced with https://creadur.apache.org/rat/apache-rat-plugin/ ?

Yes, it can!

PR #655, opened 2023-09-19
[WICKET-7074] write AJAX responses to a temporary buffer so that is A…

reiern70:
write AJAX responses to a temporary buffer so that is AJAX fails in rendering phase response is not polluted with an invalid XML

### DIFF
--- a/wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
+++ b/wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
@@ -34,20 +34,28 @@ import org.apache.wicket.Component;
 import org.apache.wicket.MarkupContainer;
 import org.apache.wicket.MockPageWithLink;
 import org.apache.wicket.MockPageWithLinkAndComponent;
+import org.apache.wicket.Page;
+import org.apache.wicket.RuntimeConfigurationType;
 import org.apache.wicket.ajax.markup.html.AjaxLink;
 import org.apache.wicket.event.IEvent;
 import org.apache.wicket.markup.IMarkupResourceStreamProvider;
 import org.apache.wicket.markup.html.WebComponent;
 import org.apache.wicket.markup.html.WebPage;
+import org.apache.wicket.markup.html.basic.Label;
+import org.apache.wicket.mock.MockApplication;
 import org.apache.wicket.request.IRequestHandler;
 import org.apache.wicket.request.cycle.IRequestCycleListener;
 import org.apache.wicket.request.cycle.RequestCycle;
+import org.apache.wicket.response.filter.IResponseFilter;
+import org.apache.wicket.settings.RequestCycleSettings;
 import org.apache.wicket.util.encoding.UrlEncoder;
 import org.apache.wicket.util.resource.IResourceStream;
 import org.apache.wicket.util.resource.StringResourceStream;
 import org.apache.wicket.util.tester.DiffUtil;
 import org.apache.wicket.util.tester.WicketTestCase;
+import org.apache.wicket.util.tester.WicketTester;
 import org.apache.wicket.util.time.Instants;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
```diff
@@ -306,6 +312,94 @@ void ajaxRedirectSetsNoCachingHeaders()
...
+		@Override
+		public Class<? extends Page> getHomePage()
+		{
+			return Wicket7074.class;
+		}
+
+
+		@Override
+		public RuntimeConfigurationType getConfigurationType() {
+			// this ise needed so that no filters are added
```
martin-g, line 327:
```suggestion
			// this is needed so that no filters are added
```

martin-g: Can we use more stable way to make sure that there are no filters ? E.g. code that removes all registered filters.
Otherwise if later a filter is added to the default ones in DEPLOYMENT mode then this test will silently change its behavior and do nothing.

wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
```diff
@@ -306,6 +312,94 @@ void ajaxRedirectSetsNoCachingHeaders()
...
+		public IResourceStream getMarkupResourceStream(MarkupContainer container,
+													   Class<?> containerClass)
+		{
+			return new StringResourceStream(
+					"<html><body><a wicket:id='action'>link1</a><br/><br/><br/><span wicket:id='label'>Link2</span></body></html>");
+		}
+	}
+
+	@Test
+	void ajaxResposeIsEmptyIfExceptionIsProducedDuringRenderingAndNoFiltersAreSet() {
```
martin-g, line 390:
```suggestion
	void ajaxResponseIsEmptyIfExceptionIsProducedDuringRenderingAndNoFiltersAreSet() {
```

wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
```diff
@@ -306,6 +312,94 @@ void ajaxRedirectSetsNoCachingHeaders()
...
+		Wicket7074Application application = new Wicket7074Application();
+		WicketTester tester = newWicketTester(application);
+		// page renders normally first time
+		tester.startPage(new Wicket7074());
+		// click on action to produce render phase error
+		assertThrows(WeirdException.class, () -> {
+			tester.clickLink("action");
+		});
+		// AJAX response is written now into a buffer => response should empty
```
martin-g, line 399: Why the buffer is not flushed ?
I'd expect that the Ajax response is still a valid XML with a redirect to the error page.

wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
```diff
@@ -306,6 +312,94 @@ void ajaxRedirectSetsNoCachingHeaders()
...
+			return new StringResourceStream(
+					"<html><body><a wicket:id='action'>link1</a><br/><br/><br/><span wicket:id='label'>Link2</span></body></html>");
+		}
+	}
+
+	@Test
+	void ajaxResposeIsEmptyIfExceptionIsProducedDuringRenderingAndNoFiltersAreSet() {
+		Wicket7074Application application = new Wicket7074Application();
+		WicketTester tester = newWicketTester(application);
+		// page renders normally first time
```
martin-g, line 393:
```suggestion
		// the page renders normally the first time
```

wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
martin-g, line 399:
```suggestion
		// AJAX response is written now into a buffer => response should be empty
```

martin-g (changes requested):

wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
reiern70: @martin-g This is true but 

image

Thus the only way I see is add a clearResponseFilters method. Or do you have a better idea?

reiern70:
@martin-g I have implemented your suggestions. Additionally I have created 

https://github.com/apache/wicket/pull/658

Where you can see this problem in action + explanations on how this happens.

wicket-core/src/main/java/org/apache/wicket/settings/RequestCycleSettings.java
```diff
@@ -244,6 +244,16 @@ public List<IResponseFilter> getResponseFilters()
 		}
 	}
 
+	/**
+	 * Allows to clear response filters.
+	 */
+	public void clearResponseFilters()
```
martin-g, line 250: Instead of adding new public API methods I'd suggest to use custom RequestCycleSettings class that does nothing at org.apache.wicket.settings.RequestCycleSettings#addResponseFilter()

wicket-core/src/test/java/org/apache/wicket/ajax/AjaxRequestHandlerTest.java
martin-g: https://github.com/apache/wicket/pull/655/commits/a5d7c973fbf1f906ea1b6ab81cdde02c4cdf3ffc#r1332582655

wicket-core/src/main/java/org/apache/wicket/settings/RequestCycleSettings.java
reiern70: Done but the factory method needs to be non final?

reiern70: https://github.com/apache/wicket/pull/655/commits/f5edc91f45e7f9f4968b82604b2f7c4b14b77038

PR #666, opened 2023-09-29
WICKET-7072 wicket tester and core tests without test jar

gluser1357:
Just before the weekend, here is the new PR for the updated branch which seems to run now for Eclipse and VSCode.

See latest comments and test results on WICKET-7072.

Thank you for testing and reviewing in different environments.


bitstorm (approved):
I'm generally happy with this PR, it's a good improvement for  the JPMS support of Wicket. I think I've found some other JPMS  related problems working on this PR but they probably deserve a separate issue. For example, resources from package org.apache.wicket from wicket-core are not visible from external modules, so in a fully JPMS compliant project we might have problems with simply using core components (like Form). I think we should add 'opens' clause to info-module.java, but as I said this should a work for a separate issue.

Forgot to mention WICKET-7072 for Jira

wicket-cdi/src/main/java/module-info.java
```diff
@@ -24,6 +24,7 @@
     requires org.apache.wicket.request;
     requires org.apache.wicket.core;
     requires org.slf4j;
+    requires static jakarta.servlet;
```
martin-g, line 27: This change is not related to this commit.
I reverted it once and everything was working.
Now it is re-introduced silently again.

martin-g (changes requested):

wicket-cdi/src/main/java/module-info.java
gluser1357: Please see my latest comment in WICKET-7072 for leaving this change.

gluser1357: > Maven Surefire plugin is only for executing the tests via Maven. It does not affect the compilation phase!

I agree, sorry for mixing up compile and surefire plugins.

> In addition: the src/test/java don't have module-info.java, i.e. they are not Java modules!

IMO (esp. after reading https://stackoverflow.com/questions/46613214/java-9-maven-junit-does-test-code-need-module-info-java-of-its-own-and-wher), /src/main/java/module-info.java is used for compiling both /src/main/java and /src/test/java, but compilation of test code seems to be different in Eclipse and Maven, leading to different results (Maven compiles test sources, but Eclipse doesn't do in this case). I think more research is needed here. It would be great if Eclipse would follow Maven in this use case. Meanwhile we could keep requires static jakarta.servlet which would prevent Eclipse developers from manually fixing the build path (via exclude module-info.java, see also my comment in WICKET-7072).

martin-g: > Meanwhile we could keep requires static jakarta.servlet which would prevent Eclipse developers from manually fixing the build path

Fully agree on this! 
I just want to propose an improvement:
- either mention the reason for the change in the commit message
- or even better - use a separate commit for it. Mixing it in the same commit with the wicket-util:test-jar one is not good

solomax:
Just have tested this branch in VSCode

there no errors :))

Hooray! :)

wicket-cdi/src/main/java/module-info.java
gluser1357: > Fully agree on this!

Ok, great!

> I just want to propose an improvement:

Thank you for this proposal. I'm unsure how to do this practically:
- revert the change, commit, redo the change with appropriate comment and commit

or

- revert the change, commit, wait until PR is merged into master (with commits squashed), open another PR for WICKET-7072, redo the change with appropriate comment and commit?

gluser1357:
> Just have tested this branch in VSCode
> 
> there no errors :))
> 
> Hooray! :)

Great, thanks for testing!

gluser1357:
There is one question left from previous version #659:

>>> wicket-core/src/main/java/org/apache/wicket/protocol/http/mock/MockHttpSession.java

>> I think this class should be moved to wicket-tester

What about also moving package _org/apache/wicket/protocol/http/mock_ to wicket-tester in this PR?

It would solve the remaining issue (see comments in WICKET-7072) "_And related, There is one issue in MockHttpSession.invalidate() that I couldn't solve, I think it's uncritical: attributes.get("wicket:" + BaseWicketTester.TestFilterConfig.class.getName())_"

wicket-core/src/main/java/org/apache/wicket/protocol/http/mock/MockHttpSession.java
```diff
@@ -146,7 +145,8 @@ public ServletContext getServletContext()
 	@Override
 	public void invalidate()
 	{
-		Session session = (Session) attributes.get("wicket:" + BaseWicketTester.TestFilterConfig.class.getName() + ":session");
+		Session session = (Session) attributes.get("wicket:" + "BaseWicketTester.TestFilterConfig" + ":session"); // TODO
```
gluser1357, line 148: Copied from previous PR #659:

>>> wicket-core/src/main/java/org/apache/wicket/protocol/http/mock/MockHttpSession.java

>> I think this class should be moved to wicket-tester

Ok, we could do that (I also thought about moving the whole  _package _org/apache/wicket/protocol/http/mock_ to wicket-tester but org.apache.wicket.core.util.string.ComponentRenderer in /src/main/java makes use of org.apache.wicket.protocol.http.mock.MockServletContext, so that won’t work).

Should we move only MockHttpSession to wicket-tester then? If not, how to rewrite this line in MockHttpSession.invalidate()?

bitstorm: If we want to move the entire package I don't see many alternatives,,,maybe the best option first glance could be move it into wicket-util module (i.e org.apache.wicket.util.protocol.http.mock ). Waiting for better ideas :-)

gluser1357: I don't think that moving the package from wicket-core to wicket-util would have advantages.

My initial assumption was that mocking stuff should be only test scoped, but it currently isn't (as ComponentRenderer shows). IMO we should leave it as it is (possibly requiring a better fix in MockHttpSession.java, see diff) OR move just MockHttpSession.java to wicket-tester?

martin-g: I think these classes should stay in -core.

gluser1357: Ok, and what about _Session session = (Session) attributes.get("wicket:" + BaseWicketTester.TestFilterConfig.class.getName() + ":session");_? Where is this relevant?

wicket-core/pom.xml
```diff
@@ -207,6 +179,24 @@ org.apache.wicket.validation.validator;-noimport:=true
 				<groupId>org.primefaces.extensions</groupId>
 				<artifactId>resources-optimizer-maven-plugin</artifactId>
 			</plugin>
+			<!--
```
solomax, line 182: @gluser1357 I believe commented-out code need to be dropped

wicket-util/pom.xml
```diff
@@ -47,22 +47,6 @@
 		<dependency>
 			<groupId>org.junit.jupiter</groupId>
```
solomax, line 48: This dependency is global (set in parent pom), maybe it can be dropped here?

wicket-core/pom.xml
gluser1357: Yes, thanks. Code is removed.

wicket-util/pom.xml
gluser1357: Yes, thanks. Code is removed.

wicket-cdi/src/main/java/module-info.java
gluser1357: @martin-g: Looking at some commits in wicket master it looks like that commits of a PR are squashed, and the comments of these commits are merged and used as comment of the squashed commit. Because of this I decided for the first option (reverted the change, commit, redo, added again and commit) and applied it.

wicket-core/src/main/java/org/apache/wicket/protocol/http/mock/MockHttpSession.java
gluser1357: This seems to be the only open issue of this PR. Could someone please take a closer look at this?

(I just don't see any reference where this attribute "wicket:[class]:session" is set. Related could be AbstractPersistentPageStore.KEY_PREFIX = "wicket:", but I dont see any other ":session" in the code.)

martin-g:
I'll take a new look soon!

wicket-core/src/main/java/org/apache/wicket/protocol/http/mock/MockHttpSession.java
martin-g: See org.apache.wicket.session.HttpSessionStore#getSessionAttributePrefix

martin-g:
@gluser1357 Why most of the commits in this PR are by user @sre ? What is your relation ?

martin-g:
org.apache.wicket.core.util.license.ApacheLicenceHeaderTest does nothing in wicket-core-tests module.
Issues:
1) it does not override `before()` and `licenseHeaders()`
2) even when 1) is fixed it does not check the sources of `wicket-core` module

gluser1357:
> @gluser1357 Why most of the commits in this PR are by user @sre ? What is your relation ?

My fault. 

> org.apache.wicket.core.util.license.ApacheLicenceHeaderTest does nothing in wicket-core-tests module. Issues:
> 
>     1. it does not override `before()` and `licenseHeaders()`
> 
>     2. even when 1) is fixed it does not check the sources of `wicket-core` module
> 
> 
> **Update**: Fixed

Thanks for fixing, I obviously forgot the two overrides in this class.

Are sources of wicket-core checked now, or is something to do here?

sre:
> @gluser1357 Why most of the commits in this PR are by user @sre ? What is your relation ?

None of them are from me. You see 'sre', because that has been used as the full name for git when authoring the commits. Github just shows the name of the commit, since there is no account connected to the mapchart.com mail address used by the 'sre' in this pull request.

martin-g:
> Are sources of wicket-core checked now, or is something to do here?

https://github.com/apache/wicket/pull/666/commits/ceacaf29fd68a28e6c6ee1554fb324e5a37bb4e8#diff-cd87ffc1ff8683a3f9ae0fcc7055de1691b36452d962a438d32b72840d5f0c99R137-R151

martin-g:
Merging ...
More improvements could be made in a separate PR. This one is too big already!

gluser1357:
> > @gluser1357 Why most of the commits in this PR are by user @sre ? What is your relation ?
> 
> None of them are from me. You see 'sre', because that has been used as the full name for git when authoring the commits. Github just shows the name of the commit, since there is no account connected to the mapchart.com mail address used by the 'sre' in this pull request.

@sre: Ah, ok, I see, we are both using "sre" ;-) I really have to take care about that when commiting, sorry for this confusion. I wonder that my "sre" commits haven't been checked and rejected by Github :/

gluser1357:
> > Are sources of wicket-core checked now, or is something to do here?
> 
> [ceacaf2#diff-cd87ffc1ff8683a3f9ae0fcc7055de1691b36452d962a438d32b72840d5f0c99R137-R151](https://github.com/apache/wicket/commit/ceacaf29fd68a28e6c6ee1554fb324e5a37bb4e8#diff-cd87ffc1ff8683a3f9ae0fcc7055de1691b36452d962a438d32b72840d5f0c99R137-R151)

Ok, tests runs for me on wicket-core with this change, nothing to do anymore here.

wicket-core/src/main/java/org/apache/wicket/protocol/http/mock/MockHttpSession.java
gluser1357: > See org.apache.wicket.session.HttpSessionStore#getSessionAttributePrefix

Ok, I see you changed this
originally:  ... _attributes.get("wicket:" + BaseWicketTester.TestFilterConfig.class.getName() + ":session");_
to: ... _attributes.get("wicket:org.apache.wicket.util.tester.BaseWicketTester.TestFilterConfig:session");_

I wonder if that's correct since TestFilterConfig is an inner class?

A simple alternative would be to save TestFilterConfig as new file and adapt the line to _attributes.get("wicket:org.apache.wicket.util.tester.TestFilterConfig:session");_

gluser1357:
> Thanks for pushing this work, @gluser1357 !

> Thanks for pushing this work, @gluser1357 !

No problem. Thank you as well for your great support. 10.0.0-M2 master doesn't show any errors in Eclipse anymore, that's really great 👍

wicket-core/src/main/java/org/apache/wicket/protocol/http/mock/MockHttpSession.java
martin-g: Good catch!
Screenshot from 2023-10-12 16-18-31

martin-g: I fixed the bug!
The best would be to use a real class reference instead of hardcoding a String like now.

bitstorm:
@gluser1357  thank you indeed for your effort! 🙏🏻

PR #670, opened 2023-10-05
WICKET-7056 try-catch attempts to operate on invalid HTTP session and…

DavesMan:
… log them

1. The getHttpSession returns non-null value in case HTTP session exists and is valid and reference to it is stored.
2. Meanwhile the session is invalidated by another thread (another http request).
3. The IF condition evaluates still to true and it results in calling httpSession.getAttribute.
4. Because the session has been invalidated by another thread this call results in IllegalStateException

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/session/HttpSessionStore.java
+++ b/wicket-core/src/main/java/org/apache/wicket/session/HttpSessionStore.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CopyOnWriteArraySet;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


martin-g (commented):
Catching and ignoring (logging as DEBUG) all `IllegalStateException`s feels dirty to me.
Synchronizing on the current HttpSessionStore instance looks OK to me!

DavesMan:
I could not find any situable object to synchronize on.
The HttpSessionStore is one instance per application. Because at least org.apache.wicket.Session#exists() is called in every request,
 synchronizing on its instance would mean synchronizing all requests incoming to that application. We have about 1000 requests per second in morning /evening peaks. I think that synchronizing 1000 req/s. one by one would have serious P4M impact.
Synchronization on one single session would be the best solution. But web containers can return different HttpSession object instances per 1 single session.
I even cannot introduce some kind of a lock object and put it in the session as an attribute. Because while calling the getAttribute I could get an IllegalStateException because of invalid session.
I aknowledge that throwing an IllegalStateException as an indication of invalid session is not very nice but so is the J2EE spec.

martin-g:
Thanks for the explanation!
I wanted to re-open the PR but it seems you already deleted it.
I guess we will come with a similar solution ...

martin-g:
Thanks for restoring it!

DavesMan:
@martin-g I am afraid there is no ideal solution of this problem. It is bad by design of the specs. We can:

1. Sync on some object like the SessionStore and loose performance. Bad.
2. Log the `IllegalStateException` and continue (and hope everyone runs a container that is following the specs).
3. Figure out some snyc object related to the session itself which is not obtained via getAttribute.
4. Build a `ConcurrentHashMap` of locks, keyed by session ID strings inside the HttpSessionStore. Memory intensive and ugly.

martin-g:
Thank you, @DavesMan !

DavesMan:
Great!

PR #682, opened 2023-10-16
[WICKET-7080] 1) make default events delivery machinery pluggable 2) …

reiern70:
1) make default events delivery machinery plug-gable 
2) allow to disabled sending events 
3) defines a reflection based machinery to deliver events


wicket-core/src/main/java/org/apache/wicket/event/ReflectionEventDispatcher.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.wicket.Component;
+import org.apache.wicket.IEventDispatcher;
+import org.apache.wicket.WicketRuntimeException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * IEventDispatcher that uses reflection in order to locate events methods.
+ */
+public class ReflectionEventDispatcher implements IEventDispatcher {
```
reiern70, line 29: Maybe we should add here a class to method map cache. As a similar project on wicket stuff have. One main difference is we don't have a component instantiation listener=> this will only happen at runtime when event is delivered.

wicket-core/src/main/java/org/apache/wicket/event/ReflectionEventDispatcher.java
```diff
@@ -0,0 +1,104 @@
...
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(ReflectionEventDispatcher.class);
+
+	/**
+	 * If set to true then only sinks annotated with @{@link EventAwareObject} will be injected.
+	 * Probably, this should be set to true in most applications and enforce the use of either @{@link EventAwareObject}
+	 * annotation or the marker interface @{@link IEventAwareObject}
+	 */
+	private final boolean restrictToEventAware;
```
reiern70, line 38: This flag should make event's processing faster and only marked classes will be scanned.

reiern70:
This PR still needs some more work and care

reiern70:
I have added more tests and allow to use more that one eventType

martin-g:
https://github.com/wicketstuff/core/tree/master/annotationeventdispatcher-parent

reiern70:
> https://github.com/wicketstuff/core/tree/master/annotationeventdispatcher-parent

@martin-g I'm aware of that work. This PR is 2 things

1. Make "default" event delivery plug-able. This means replacing it by something else (or disabling it)
2. Some simple event annotation based events delivery. Where you can get protected or private or public methods injected with an event, 
 
```
@OnEvent(XXX.class)
protected onXXX(IEvent<XXX>  event) {
}

```

This allows to still work with event and programmatically cancel it.

or
 
```
@OnEvent(XXX.class)
protected onXXX(XXX  payload) {
}

```
will both work. No restrictions on XXX are imposed. 

IMHO, 1. would be nice to have in wicket. As for 2. I can understand can be implemented in many different ways.

3. This approach does not require a component listener and only marked classes are scanned (by annotation or implementing an interface)

reiern70:
@martin-g My last commits adds caching

reiern70:
Ok, I have reverted the part related to adding a reflection based machinery. What remains is:

1. Allow to plug your own default "event  handler"
2. As a side effect it is possible to disable events delivery

wicket-core/src/main/java/org/apache/wicket/settings/FrameworkSettings.java
```diff
@@ -45,8 +45,28 @@
  */
 public class FrameworkSettings implements IEventDispatcher
 {
+	/**
+	 *  Does the standard delivery of events. Override and do nothing if you want to disable it.
```
martin-g, line 49: Since it is `private` it is not possible to `override` it

wicket-core/src/main/java/org/apache/wicket/settings/FrameworkSettings.java
```diff
@@ -68,7 +88,7 @@ public FrameworkSettings(final Application application)
 	 * Gets the Wicket version. The Wicket version is in the same format as the version element in
 	 * the pom.xml file (project descriptor). The version is generated by maven in the build/release
 	 * cycle and put in the /META-INF/MANIFEST.MF file located in the root folder of the Wicket jar.
-	 *
+	 * <p></p>
```
martin-g, line 91: Most probably you want the closing `</p>` to be at the end of the paragraph.

wicket-core/src/main/java/org/apache/wicket/settings/FrameworkSettings.java
```diff
@@ -129,26 +149,30 @@ public FrameworkSettings add(IEventDispatcher dispatcher)
 		return this;
 	}
 
+	/**
+	 * @return Returns <code>true</code> if there is at least one event dispatcher
+	 */
+	public final boolean hasAnyAnyEventDispatchers()
```
martin-g, line 155: one too many `any`

martin-g (commented):
Please add a test case with disabled default event dispatcher.

wicket-core/src/main/java/org/apache/wicket/settings/FrameworkSettings.java
reiern70: Or. create an  empty paragraph to separate from next IJ complains here

wicket-core/src/main/java/org/apache/wicket/settings/FrameworkSettings.java
reiern70: /**
	 *  Does the standard delivery of events. Events can be disabled at application level by setting a <code>null</code>
	 *  {@link IEventDispatcher} via {@link #setDefaultEventDispatcher(IEventDispatcher)} and setting  no additional
	 *  {@link IEventDispatcher}s (via {@link #add(IEventDispatcher)}).
	 */

wicket-core/src/main/java/org/apache/wicket/settings/FrameworkSettings.java
reiern70: Done.

wicket-core/src/main/java/org/apache/wicket/settings/FrameworkSettings.java
reiern70: Done

PR #683, opened 2023-10-19
WICKET-7083: Trigger client side validation when using SubmitLinks

asvanberg:
[WICKET-7083](https://issues.apache.org/jira/browse/WICKET-7083)

Instead of using jQuery's event triggering for the submit event, use the [browsers built-in functionality](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit) which, as a bonus, will also perform client side validation (such as the `required` attribute).

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js
@@ -2422,6 +2422,22 @@
 			},
 
 			/**
+			 * Submits the given form using, if available, standard form processing
+			 * including client-side validation and firing of SubmitEvent.
+			 * If it is not available, uses jQuery event triggering to submit the form
+			 * and send out the SubmitEvent.
+			 *
+			 * @param form {HTMLFormElement} form to submit
+			 */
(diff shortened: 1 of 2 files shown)


martin-g (commented):
According to my web search people had problem with this method on Safari.
According to https://caniuse.com/?search=requestSubmit and https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit it is well supported on all major browsers! (Safari 16+)

We already send information about the submitter, so there is no need to pass it as parameter!

The code change looks good to me!

But we need a JIRA ticket for the changelog!

asvanberg:
I checked caniuse.com and saw that it had had broad support for over a year so I figured that would be good enough. I considered checking if the method existed and if not, use the old jQuery trigger but decided against it. Easy enough to add back if necessary.

Is a JIRA ticket something I have to do?

martin-g:
To be on the safe side I'd suggest to introduce `Wicket.Event.requestSubmit(f)` function that checks for availability and uses it or falls back to the current code if not available.

martin-g (approved):

bitstorm (approved):

bitstorm:
Thanks @asvanberg !

PR #684, opened 2023-10-19
Easier to work with polymorphic values inside IModel

asvanberg:
Wicket currently leaves a lot to be desired when working with polymorphic datastructures.

Consider the following scenario:
```java
interface TextMatchingService {
  sealed interface Status {
    record NotSubmitted(Runnable submit) implements Status {}
    record Queued(Instant since, Optional<Instant> expectedFinish) implements Status {}
    record Analysed(float percentMatch) implements Status {}
    record Error(int errorCode, String humanReadableMessage) implements Status {}
  }
  Status getStatus(File file);
}
```
Now I want to have a panel that gets the status of a file and then, depending on the status, shows the different cases.

If I were to unpack the model and `switch` over the type in the constructor, when the value returned by the `IModel<Status>` changes due to the asynchronous nature of the text matching the hierarchy would be out of sync.

The proposed `poly` function would enable the following pattern for dealing with this case.

```java
public class TextMatchingPanel extends GenericPanel<File> {
  @Inject
  TextMatchingService textMatchingService;

  public TextMatchingPanel(String id, IModel<File> fileModel) {
    super(id, fileModel);
    IModel<Status> statusModel = fileModel.map(textMatchingService::getStatus);
    
    add(new NotSubmittedPanel("not_submitted", statusModel.poly(NotSubmitted.class));
    [add more specific status panels]
  }

  class NotSubmittedPanel extends Panel {
    NotSubmittedPanel(String id, IModel<NotSubmitted> notSubmittedModel) {
      super(id, notSubmittedModel);
      add(new Link<>("submit") {
        @Override
        public void onClick() {
          NotSubmitted notSubmitted = notSubmittedModel.getObject();
          Runnable submitForTextMatching = notSubmitted.submit();
          submitForTextMatching.run();
        }
      });
    }

    @Override
    protected void onConfigure() {
      super.onConfigure();
      setVisible(getDefaultModelObject() != null);
    }
  }

  [more status panels]
}
```

If there's a better way to do this, preferably using switch expressions (to get totality checking), that would be preferred but I've yet to come up with one.

Unpacking in the constructor and creating different component hierarchies would work if the page was stateless but then there would also be no need for a detachable `IModel` since nothing gets serialized.

### DIFF
--- a/wicket-core-tests/src/test/java/org/apache/wicket/model/IModelTest.java
+++ b/wicket-core-tests/src/test/java/org/apache/wicket/model/IModelTest.java
@@ -234,4 +234,42 @@ class IModelTest
 		assertNotNull(clone);
 		assertEquals("Some Name", clone.getObject());
 	}
+
+	sealed interface TextMatchingStatus
+	{
+		record NotSubmitted() implements TextMatchingStatus {}
+		record Queued() implements TextMatchingStatus {}
+		record Analysed(int matchingInPercent) implements TextMatchingStatus {}
+		record Error(int errorCode, String humanReadableMessage) implements TextMatchingStatus {}
+	}
+
+	@Test
+	void polymorphicModelWrongClass()
+	{
+		IModel<TextMatchingStatus> statusModel = LoadableDetachableModel.of(() ->
+				new TextMatchingStatus.Error(3, "File too big"));
+		IModel<TextMatchingStatus.Queued> poly = statusModel.poly(TextMatchingStatus.Queued.class);
+
+		assertNull(poly.getObject());
+	}
+
+	@Test
+	void polymorphicModelCorrectClass()
+	{
+		IModel<TextMatchingStatus> statusModel = LoadableDetachableModel.of(() ->
+				new TextMatchingStatus.Analysed(14));
+		IModel<TextMatchingStatus.Analysed> poly = statusModel.poly(TextMatchingStatus.Analysed.class);
+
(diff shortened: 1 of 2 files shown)


martin-g (commented):
I can see the usefulness of this new method!
I'm just not sure about its name. To me `as()` looks more clear!

We need a JIRA issue for the changelog!

asvanberg:
I agree about the name change, `as` is much more clear than `poly`.

No idea about the JIRA issue, is that something I have to do?

bitstorm:
@asvanberg thanks for your contribution! For JIRA you can request an  account here:

https://selfserve.apache.org/jira-account.html

then you can create a issue related to this pull request on your project page

https://issues.apache.org/jira/projects/WICKET/

asvanberg:
https://issues.apache.org/jira/browse/WICKET-7082

svenmeier:
What is the model object of `statusModel.poly(NotSubmitted.class)` if the status isn't a NotSubmitted but instance of another subclass?

asvanberg:
> What is the model object of `statusModel.poly(NotSubmitted.class)` if the status isn't a NotSubmitted but instance of another subclass?

`null`

svenmeier:
> `null`

Ah, should have taken a look at the code changes ;)

As a user of that API a null in that case seems surprising to me. But I'm ok with it.

martin-g (approved):

PR #708, opened 2023-11-22
WICKET-7087: AjaxLazyLoadPanelTester not available in 10.0.0-M2

martin-g:
Introduce wicket-extensions-tester module that currently provides only AjaxLazyLoadPanelTester

Add all **-tester modules to wicket-common-tests for better OSGi compatibility.


wicket-extensions-tester/src/test/java/org/apache/wicket/extensions/tester/AjaxLazyLoadPanelTesterTest.java
```diff
@@ -1,63 +1,65 @@
-/*
```
martin-g, line 1: `CRLF` endline has been changed to `LF`. Otherwise Git was refusing to move the files.

martin-g: Use https://github.com/apache/wicket/pull/708/files?diff=unified&w=1 to review without the whitespace changes

PR #758, opened 2024-01-04
WICKET-7090: add outputTimestamp property to enable reproducible builds

papegaaij:
In addition to reprocible builds, this should also result in timestamps being set on files in jars.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,6 +129,8 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
+		<project.build.outputTimestamp>2024-01-01T00:00:00Z</project.build.outputTimestamp>
+
 		<javadoc.additionalJOption />
 		<javadoc.jdk.apidocs.link>https://docs.oracle.com/en/java/javase/${java.specification.version}/docs/api/</javadoc.jdk.apidocs.link>
 
--- a/pom.xml
+++ b/pom.xml
@@ -916,9 +916,9 @@
 					<executions>
 						<execution>
 							<id>default-bundle</id>
-							<phase>package</phase>
+							<phase>process-classes</phase>
 							<goals>
-								<goal>bundle</goal>
+								<goal>manifest</goal>
 							</goals>
 							<configuration>
 								<instructions>
@@ -1060,6 +1060,7 @@
 					<version>${maven-jar-plugin.version}</version>
 					<configuration>
 						<archive>
+							<manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
 							<manifest>
 								<addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
 								<addDefaultImplementationEntries>true</addDefaultImplementationEntries>
--- a/wicket-auth-roles/pom.xml
+++ b/wicket-auth-roles/pom.xml
@@ -24,7 +24,7 @@
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 	<artifactId>wicket-auth-roles</artifactId>
-	<packaging>bundle</packaging>
+	<packaging>jar</packaging>
 	<name>Wicket Auth Roles</name>
 	<description>
 		Wicket Authorization Integration Based on roles, metadata and
--- a/wicket-bean-validation/pom.xml
+++ b/wicket-bean-validation/pom.xml
@@ -24,7 +24,7 @@
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 	<artifactId>wicket-bean-validation</artifactId>
-	<packaging>bundle</packaging>
+	<packaging>jar</packaging>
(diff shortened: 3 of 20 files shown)


papegaaij:
Ok, unfortunately, this does not seem to give the desired effect. The files in the jars still don't have timestamps.

wicket-auth-roles/pom.xml
```diff
@@ -24,7 +24,7 @@
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 	<artifactId>wicket-auth-roles</artifactId>
-	<packaging>bundle</packaging>
+	<packaging>jar</packaging>
```
martin-g, line 27: Does this break the OSGi support ?!
CC @mattrpav

martin-g:
In addition: one of the commits uses `TKH-7090` instead of `WICKET-7090`

wicket-auth-roles/pom.xml
papegaaij: As far as I know, it shouldn't. A jar is an OSGi bundle, when it's manifest contains OSGi related information, but it's still a jar. The documentation for the maven-bundle-plugin even describes this setup as a valid solution: https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#_adding_osgi_metadata_to_existing_projects_without_changing_the_packaging_type

papegaaij:
> In addition: one of the commits uses `TKH-7090` instead of `WICKET-7090`

Doh, TKH is the prefix we use for Topicus KeyHub :) I've fixed the commit.

wicket-auth-roles/pom.xml
mattrpav: Yes, type should be ‘bundle’ to activate the bundle plugin when generating the jar. 

ActiveMQ had changes to support reproducible builds and maintained the bundle plugin. 

I recommend we should know exactly what is the miss or change needed to support reproducible builds, and not just change the project type to jar.

papegaaij: @mattrpav the bundle plugin is activated by binding one of its goals to the maven execution phase (in this case `manifest` in `process-classes`). This will generate the MANIFEST.MF, which will then be used by the maven-jar-plugin. I can see the correct MANIFEST.MF in the produced jar.

The problem with the maven-bundle-plugin is that it generates jars without timestamps on its entries. This is valid (and considered reproducible), but very inconvenient. There is no way to change this when using packaging `bundle`. I think the produced jars are fine. They do contain the correct MANIFEST.MF. However, we do not use OSGi, so I have no way of testing it.

mattrpav: @papegaaij the bundle plugin can add any field to the MANIFEST.MF. What field are you looking for in the MANIFEST.MF?

pom.xml
```diff
@@ -129,6 +129,8 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
+		<project.build.outputTimestamp>2024-01-01T00:00:00Z</project.build.outputTimestamp>
```
mattrpav, line 132: Using a static value seems difficult to manage. I suggest looking at using the git commit timestamp. This would be automatic, it would change with ever commit and ensure reproducible builds get the same timestamp when rebuilt from any commit.

mattrpav (changes requested):

pom.xml
solomax: Maven-release-plugin updates this field during release .....

wicket-auth-roles/pom.xml
papegaaij: @mattrpav This change is to get correct timestamps on files (also called entries) in the jar files. The maven-bundle-plugin does not set these timestamps at all, causing them to revert to the jar-epoch (02-01-1980). The maven-jar-plugin honors `project.build.outputTimestamp`. If it also gives us reproducible builds, that's a bonus but not the reason for this change.

We do however not want to break OSGi support. However, none of the Wicket core developers uses OSGi, so this is very hard for us to test. With this change, the produced jars *should* still be valid OSGi bundles, but we have no good way to verify this. It be really great of you could check this.

As far as I know OSGi, the only thing that makes a normal jar an OSGi bundle is a set of entries in the MANIFEST.MF. These all seem unaffected by this change (I compared before and after), so I think the change is correct. But maybe you can point to something else that must be in place for an OSGi bundle?

pom.xml
papegaaij: Yes, the maven-release-plugin should update this field automatically during a release. We do need to verify this at the next release though.

papegaaij:
@martin-g are you ok with merging this change?

martin-g (approved):
LGTM!

Hopefully the change will be tested by an OSGi user during the vote for 10.0.0! 
If there is a regression we can work on a fix in 10.0.x!

mattrpav:
@martin-g @papegaaij I'll test the v10 release in OSGi and report any issues. Thanks.

papegaaij:
Thank you @mattrpav

PR #761, opened 2024-01-07
WICKET-7094: Add missing scheme values of CSP

renoth:
WICKET-7094


renoth:
See also https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources#scheme-source

martin-g (changes requested):
The change looks good but:
1) we will need a JIRA ticket for the changelog
2) it would be good to add some new unit tests

renoth:
Ticket is https://issues.apache.org/jira/browse/WICKET-7094

martin-g (approved):

martin-g:
Thank you for the contribution, @renoth !

PR #762, opened 2024-01-09
PartialPageUpdater could not update if different object references were used for its own page and the component to update

aminabromand:
fixed equality check of PartialPageUpdater to use equals(...) method instead of equals sign

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Page.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Page.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 
 import org.apache.wicket.authorization.UnauthorizedActionException;
@@ -1011,5 +1012,22 @@ public abstract class Page extends MarkupContainer
 	public final boolean wasRendered(Component component)
 	{
 		return renderedComponents != null && renderedComponents.contains(component);
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+		Page that = (Page) o;
+		return autoIndex == that.autoIndex && numericId == that.numericId;
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(autoIndex, numericId);
 	}
 }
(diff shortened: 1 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/page/PartialPageUpdate.java
```diff
@@ -567,7 +567,7 @@ public final void add(final Component component, final String markupId)
 				LOG.warn("Component '{}' not cannot be updated because it was already removed from page", component);
 				return;
 			}
-			else if (pageOfComponent != page) 
+			else if (!pageOfComponent.equals(page))
```
svenmeier, line 570: Do you have a usecase, why this change should be necessary?

The check is precisely there to check on the page instance.

aminabromand: Sorry, I have missed the most important part in the pull request.
So now I have added another commit containing the `equals()` and `hashCode()` methods for the Page.class.

Well my usecase is the following:
I have a `AbstractAjaxTimerBehavior` on my page, which runs every 500 milli secs.
This `AbstractAjaxTimerBehavior` pulls Buttons out of a registry.
The registry contains the Buttons together with an information, whether the button should be enabled or disabled.
The state of the button enabled/disabled can be updated via a websocket, that is connected to the server.
So when the `AbstractAjaxTimerBehavior` pulls the Buttons our of the registry, the buttons are assigned to the same page, but it is not the very same object in the heap. 
Thus the inequality check with `pageOfComponent != page` goes true, even though the page is the same.
This is why I would like to have the inequality check as `!pageOfComponent.equals(page)` with the `equals()` method checking on the immutal fields of the page.

Do you agree, that this is a sensible approach, or would you suggest otherwise?

wicket-core/src/main/java/org/apache/wicket/Page.java
```diff
@@ -1012,4 +1013,21 @@ public final boolean wasRendered(Component component)
...
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+		Page that = (Page) o;
+		return autoIndex == that.autoIndex && numericId == that.numericId;
```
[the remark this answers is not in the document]
aminabromand: you are right, that `autoIndex` is not good for the method as it can change.
But so does `renderCount`.
So I change the method to use `numericId` only.

wicket-core/src/main/java/org/apache/wicket/page/PartialPageUpdate.java
svenmeier: Changing #equals and #hashCode in such a central class can have severe negative consequence, that I would not like to risk.
Your idea with updating via component path sounds good to me.

aminabromand: ok, understood.
Anyhow, thanks for the support and quick responses! 👍

PR #768, opened 2024-01-15
WICKET-7096: add missing nonce to auto linked resources

sebthom:
https://issues.apache.org/jira/browse/WICKET-7096


martin-g (changes requested):
It would be nice to add a test case too

sebthom:
> It would be nice to add a test case too

done

martin-g (approved):

martin-g:
Thank you, @sebthom !

PR #781, opened 2024-02-09
WICKET-6890 Introduce strategy for class name debug output

theigl:
This PR introduces a strategy for outputting a container's class name for debugging purposes.

Up until 8bbf158ebc4d12f3b317c0d607200441c0767162, `ClassNameStrategy.HTML_COMMENT` was the default strategy. Since then  `ClassNameStrategy.TAG_ATTRIBUTE` has become the default and only option.

For applications that strip Wicket tags during development, this means that there is currently no more debug output.


wicket-core/src/main/java/org/apache/wicket/Page.java
```diff
@@ -827,9 +827,10 @@ public void component(final Component component, final IVisit<Void> visit)
...
 		}
 
+		//TODO WICKET-6890 What about TAG_ATTRIBUTE?
 		if (getApplication().getDebugSettings().isOutputMarkupContainerClassName())
 		{
-			String className = Classes.name(getClass());
+			final String className = Classes.name(getClass());
 			getResponse().write("<!-- Page Class ");
 			getResponse().write(className);
 			getResponse().write(" END -->\n");
```
theigl, line 836: What should we do here? Ignore the strategy and always write an HTML comment, or only write it when the `HTML_COMMENT` strategy is enabled?

wicket-jmx/src/main/java/org/apache/wicket/jmx/DebugSettingsMBean.java
```diff
@@ -71,6 +71,8 @@ public interface DebugSettingsMBean
 	 * Enables wrapping output of markup container in html comments that contain markup container's
 	 * class name. (Useful for determining which part of page belongs to which markup file).
 	 * 
+	 * //TODO WICKET-6890 JMX Probably does not work with enum, What do we do here?
```
theigl, line 74: @martin-g: Any idea what we should do about the JMX bean? Looking at all available beans, enums have been carefully avoided.

martin-g: One way is to use the enum name as a String.
Another way is to use MXBean instead of MBean.

martin-g:
@theigl Do you need help here ?

wicket-jmx/src/main/java/org/apache/wicket/jmx/DebugSettingsMBean.java
theigl: I went the "enum name as a String" way.

theigl:
> @theigl Do you need help here ?

Thanks a lot for asking @martin-g! I just found the time to continue working on this.

I improved the JavaDoc some more and added a method to set the strategy to the JMX bean.

martin-g (approved):

PR #794, opened 2024-02-20
NO-JIRA: fix possible Pattern collision with obj.value == {}

renoth:
No JIRA issue because this just fixes the logging pattern for an INFO message.
Also i applied the Wicket autoformatter to the class


wicket-tester/src/main/java/org/apache/wicket/util/tester/BaseWicketTester.java
```diff
@@ -2258,8 +2267,10 @@ public void debugComponentTrees(String filter)
 		{
 			if (obj.path.matches(".*" + filter + ".*"))
 			{
-				log.info("[{}{}] path\t" + obj.path + " \t" + obj.type + " \t[" + obj.value + "]",
-					obj.isEnabled ? "E" : "-", obj.isVisible ? "V" : "-");
+				var enabled = obj.isEnabled ? "E" : "-";
```
renoth, line 2270: the relevant change

martin-g (approved):

PR #798, opened 2024-02-25
Fix condition for simple top level navigation

papegaaij:


martin-g (commented):
Do we want to add some unit tests ?

papegaaij:
@martin-g The tests were already supposed to check this behavior, however WicketTester defaults to POST, which caused it not to trigger the error. I've updated the tests to use GET. Also, the test that was supposed to check the top level navigation did not trigger that specific code, because it used an incorrect sec-fetch-site header.

martin-g (approved):
Looks better now!
Maybe we should add a new test for `POST` too ?

papegaaij:
> Maybe we should add a new test for POST too ?
Yes, good idea. I've added the test.

PR #812, opened 2024-03-20
[WICKET-7101] The label associated with form component should always …

reiern70:
The label associated with form component should always be rendered when the form component is rendered via AJAX. This way we ensure the state of the label is consistent with form component's state. This introduces a new wicket attribute wicket:auto that controls that auto-label is bound to form component.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/form/AjaxFormComponentUpdatingBehavior.java
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/form/AjaxFormComponentUpdatingBehavior.java
@@ -25,6 +25,7 @@ import org.apache.wicket.ajax.AjaxEventBehavior;
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.ajax.attributes.AjaxRequestAttributes;
 import org.apache.wicket.ajax.attributes.AjaxRequestAttributes.Method;
+import org.apache.wicket.core.request.handler.IPartialPageRequestHandler;
 import org.apache.wicket.markup.html.form.FormComponent;
 import org.apache.wicket.markup.html.form.validation.IFormValidator;
 import org.apache.wicket.util.lang.Args;
@@ -157,7 +158,7 @@ public abstract class AjaxFormComponentUpdatingBehavior extends AjaxEventBehavio
 		{
 			onError(target, e);
 		}
-		formComponent.updateAutoLabels(target);
+		formComponent.updateAutoLabels((IPartialPageRequestHandler)target, false);
 	}
 
 	/**
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
@@ -36,6 +36,7 @@ import org.apache.wicket.Page;
 import org.apache.wicket.WicketRuntimeException;
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.behavior.Behavior;
+import org.apache.wicket.core.request.handler.IPartialPageRequestHandler;
 import org.apache.wicket.core.util.string.CssUtils;
 import org.apache.wicket.event.IEvent;
 import org.apache.wicket.markup.ComponentTag;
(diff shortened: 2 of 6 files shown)


wicket-core/src/test/java/org/apache/wicket/markup/html/form/AutoLabelTest.java
```diff
@@ -163,22 +167,80 @@ class MyTestPage2 extends TestPage
...
+	void labelTagClassesAreUpdatedWhenRelatedFormComponentIsUpdated()
+	{
+		class MyAjaxTestPage extends TestPage
+		{
+			private static final long serialVersionUID = 1L;
+
+			// we define a label that is auto-synchronized with component
+			MyAjaxTestPage()
+			{
+				super("<label wicket:for='t' wicket:auto='true'><span class='label-text'>field</span></label>");
```
reiern70, line 181: This introduces a new attribute wicket:auto="true" that essentially say label is always repainted when component is re-rendered this way label is always up to date.

If not set label behaves as before. Thus, this change should be backward compatible.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/FormComponent.java
```diff
@@ -301,6 +302,13 @@ private String getLabel()
 		}
 	}
 
+	@Override
+	protected void onBeforeRender()
+	{
+		// WICKET-7101 update related auto-label
+		getRequestCycle().find(IPartialPageRequestHandler.class).ifPresent(handler -> updateAutoLabels(handler, true));
```
reiern70, line 309: This will only succeed if the label have the `wicket:auto="true"` attribute.

reiern70:
> Maybe another possibility is to add an application setting toggling this behavior (and set it to disabled by default). This, way existing applications will not break.

Added some application setting allowing to toggle on and off the feature

wicket-core/src/main/java/org/apache/wicket/ajax/form/AjaxFormComponentUpdatingBehavior.java
```diff
@@ -157,7 +158,7 @@ protected final void onEvent(final AjaxRequestTarget target)
 		{
 			onError(target, e);
 		}
-		formComponent.updateAutoLabels(target);
+		formComponent.updateAutoLabels((IPartialPageRequestHandler)target, false);
```
martin-g, line 161: Is the casting really needed here ?

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
```diff
@@ -873,7 +874,7 @@ else if (hasError())
 				@Override
 				public void component(FormComponent<?> component, IVisit<Void> visit)
 				{
-					component.updateAutoLabels(target);
+					component.updateAutoLabels((IPartialPageRequestHandler)target, false);
```
martin-g, line 877: Is the casting needed ?

wicket-core/src/main/java/org/apache/wicket/markup/html/form/FormComponent.java
```diff
@@ -1558,9 +1566,9 @@ public final IValidatable<T> newValidatable()
 	 * Updates auto label css classes such as error/required during ajax updates when the labels may
 	 * not be directly repainted in the response.
 	 * 
-	 * @param target
+	 * @param target The {@link IPartialPageRequestHandler}
```
martin-g, line 1569: Please add `@param auto ...` too

wicket-core/src/main/java/org/apache/wicket/markup/html/form/FormComponent.java
```diff
@@ -1570,9 +1578,30 @@ public final void updateAutoLabels(AjaxRequestTarget target)
...
 
+		if (auto)
+		{
+			if (marker.isAuto())
+			{
+				marker.updateFrom(this, target);
+				return;
+			} else {
+				return;
+			}
```
martin-g, line 1589: This can be simplified to 
```
if (!marker.isAuto())
{
  return;  		
}
marker.updateFrom(this, target); 
```

Maybe even more.

martin-g (commented):
It would be nice to update the guide too.

wicket-core/src/main/java/org/apache/wicket/ajax/form/AjaxFormComponentUpdatingBehavior.java
reiern70: Not really.

martin-g:
Just edit the adoc files. Maven does the rest.
One more thing - please update wicket.xsd too

PR #824, opened 2024-03-28
[WICKET-7101] modify schema definition

reiern70:


wicket-core/src/main/resources/META-INF/wicket.xsd
```diff
@@ -102,6 +103,15 @@
 		</simpleType>
 	</attribute>
 
+	<attribute name="auto" type="boolean">
+		<annotation>
+			<documentation>Can be used in HTML <![CDATA[<label]]> to indicate the label will be automatically updated
```
martin-g, line 108:
```suggestion
			<documentation>The attribute can be used in HTML &lt;label&gt; to indicate the label will be automatically updated
```

reiern70: Hum... other elements use CDATA



Maybe change all of them too?

reiern70: Screenshot from 2024-03-28 15-48-36

martin-g: I updated the suggestion

PR #832, opened 2024-04-09
[WICKET-7102]

renoth:
Fix Error when Bytes Constructor is called with -1


reiern70 (approved):
LGTM! Thanks for contributing

bitstorm (approved):

PR #846, opened 2024-04-25
WICKET-7107 improved BaseWicketTester#WicketTesterServletWebResponse

dr0ps:
WICKET-7107 seems to not be an issue in Wicket core but rather a shortcoming of BaseWicketTester. WicketTesterServletWebResponse does implement IMetaDataBufferingWebResponse but fails to buffer header data. Headers are therefore lost in some cases. I took the provided test from WICKET-7107 and reduced it to the minimum required to demonstrate the issue. I also added the header buffering to BaseWicketTester#WicketTesterServletWebResponse. 
I believe that this is an improvement but it ultimately falls short of actually testing wicket core functionality. It only tests the test infrastructure.

### DIFF
--- a/wicket-core-tests/src/test/java/org/apache/wicket/csp/CSPRequestCycleListenerTest.java
+++ b/wicket-core-tests/src/test/java/org/apache/wicket/csp/CSPRequestCycleListenerTest.java
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.wicket.csp;
+
+
+import org.apache.wicket.Page;
+import org.apache.wicket.RestartResponseException;
+import org.apache.wicket.core.request.handler.PageProvider;
+import org.apache.wicket.core.request.handler.RenderPageRequestHandler.RedirectPolicy;
+import org.apache.wicket.markup.html.WebPage;
+import org.apache.wicket.protocol.http.WebApplication;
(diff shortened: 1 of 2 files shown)


dr0ps:
ResetResponseException$ResponseResettingDecorator#respond(RequestCycle) calls reset() on the response object. The current implementation in BaseWicketTester$WicketTesterServletWebResponse does nothing which is not representative of the implementation in BufferedWebResponse or ServletWebResponse. I therefore added the clearing of cookies and headers which makes the tests fail.
As the ResetResponseException then continues to call IRequestHandler#respond(RequestCycle) the onRequestHandlerResolved and onRequestHandlerExecuted in CSPRequestCycleListener will not be called. Therefore the CSP values are missing.

dr0ps:
I replaced CSPRequestCycleListener#onRequestHandlerResolved and CSPRequestCycleListener#onRequestHandlerExecuted with onUrlMapped which fixes the tests. Is the a good idea?

wicket-core/src/main/java/org/apache/wicket/csp/CSPRequestCycleListener.java
```diff
@@ -39,14 +40,7 @@ public CSPRequestCycleListener(ContentSecurityPolicySettings settings)
...
 	@Override
-	public void onRequestHandlerResolved(RequestCycle cycle, IRequestHandler handler)
-	{
-		// WICKET-7028- this is needed for redirect to buffer use case.
-		protect(cycle, handler);
-	}
-
-	@Override
-	public void onRequestHandlerExecuted(RequestCycle cycle, IRequestHandler handler)
+	public void onUrlMapped(RequestCycle cycle, IRequestHandler handler, Url url)
```
reiern70, line 43: I have tried this with 

https://issues.apache.org/jira/browse/WICKET-7028

sample application and this change seems to be Ok. But to be honest I'm not sure this is the way we should go.  See 

https://issues.apache.org/jira/browse/WICKET-7040

reiern70 (changes requested):
I'm not sure this is the way to go. This seems to fix a previous ISSUE 

https://issues.apache.org/jira/browse/WICKET-7028

But this I had to check by manual testing. IMHO, we should not invest more time on fixing corner cases and invest some time doing

https://issues.apache.org/jira/browse/WICKET-7040

wicket-core/src/main/java/org/apache/wicket/csp/CSPRequestCycleListener.java
dr0ps: WICKET-7040 is never going to work, I think. Due to ajax requests, components can be instantiated long after the original page has been created and long after the original CSP has been sent to the browser. Additional CPSs could be delivered to the client as meta tags but weakening the original CSP is not allowed ( https://www.w3.org/TR/CSP3/#multiple-policies ). Therefore the strictest possible set of policies has to be known before the page is rendered.

PR #861, opened 2024-05-09
WICKET-7110 Feat: add lock holding stack to CouldNotLockPageException

dashorst:
Adds the stack trace of the thread holding the lock of a page in case of
a timeout trying to acquire a lock. This is quite handy when trying to
figure out what caused the CouldNotLockPageException without having to
go sift through logs.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
+++ b/wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
@@ -133,8 +133,8 @@ public class DefaultPageLockManager implements IPageLockManager {
 		}
 		else
 		{
-			final String previousThreadName = previous != null ? previous.getThread().getName() : "N/A";
 			final Thread previousThread = previous != null ? previous.getThread() : null;
+			final String previousThreadName = previousThread != null ? previousThread.getName() : "N/A";
 			if (logger.isWarnEnabled())
 			{
 				logger.warn(
@@ -228,11 +228,11 @@ public class DefaultPageLockManager implements IPageLockManager {
 	}
 }
 
-class PageLockedException extends Exception
+static class PageLockedException extends Exception
 {
-	PageLockedException(Thread previousThread, int pageId) 
+	PageLockedException(Thread pageHoldingThread, int pageId) 
 	{
-		super("This thread " + previousThread.getName() + " holds the lock to page " + pageId);
-		setStackTrace(previousThread.getStackTrace());
+		super("This thread " + pageHoldingThread.getName() + " holds the lock to page " + pageId);
+		setStackTrace(pageHoldingThread.getStackTrace());
 	}
 }
--- a/wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
+++ b/wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
@@ -133,9 +133,10 @@ public class DefaultPageLockManager implements IPageLockManager {
 		}
 		else
 		{
+			final String previousThreadName = previous != null ? previous.getThread().getName() : "N/A";
+			final Thread previousThread = previous != null ? previous.getThread() : null;
 			if (logger.isWarnEnabled())
 			{
-				final String previousThreadName = previous != null ? previous.getThread().getName() : "N/A";
 				logger.warn(
 						"Thread '{}' failed to acquire lock to page with id '{}', attempted for {} out of allowed {}." +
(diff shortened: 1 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
```diff
@@ -222,3 +227,12 @@ Supplier<ConcurrentMap<Integer, PageAccessSynchronizer.PageLock>> getLocks()
 		return locks;
 	}
 }
+
+class PageLockedException extends Exception
```
martin-g, line 231:
```suggestion
static class PageLockedException extends Exception
```

wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
```diff
@@ -133,9 +133,10 @@ public void lockPage(int pageId) throws CouldNotLockPageException
 		}
 		else
 		{
+			final String previousThreadName = previous != null ? previous.getThread().getName() : "N/A";
+			final Thread previousThread = previous != null ? previous.getThread() : null;
```
martin-g, line 137: nit: swap these two lines and simplify the latter to: `final String previousThreadName = previousThread != null ? previousThread.getName() : "N/A";`

wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
```diff
@@ -222,3 +227,12 @@ Supplier<ConcurrentMap<Integer, PageAccessSynchronizer.PageLock>> getLocks()
 		return locks;
 	}
 }
+
+class PageLockedException extends Exception
+{
+	PageLockedException(Thread previousThread, int pageId)
```
martin-g, line 233:
```suggestion
	PageLockedException(Thread pageHoldingThread, int pageId) 
```

martin-g (approved):

wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
dashorst, line 137:
```suggestion
			final Thread previousThread = previous != null ? previous.getThread() : null;
			final String previousThreadName = previousThread != null ? previousThread.getName() : "N/A";
```

wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
```diff
@@ -222,3 +227,12 @@ Supplier<ConcurrentMap<Integer, PageAccessSynchronizer.PageLock>> getLocks()
 		return locks;
 	}
 }
+
+class PageLockedException extends Exception
+{
+	PageLockedException(Thread previousThread, int pageId) 
+	{
+		super("This thread " + previousThread.getName() + " holds the lock to page " + pageId);
+		setStackTrace(previousThread.getStackTrace());
```
dashorst, line 236:
```suggestion
		super("This thread " + pageHoldingThread.getName() + " holds the lock to page " + pageId);
		setStackTrace(pageHoldingThread.getStackTrace());
```

wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java
```diff
@@ -222,3 +227,12 @@ Supplier<ConcurrentMap<Integer, PageAccessSynchronizer.PageLock>> getLocks()
 		return locks;
 	}
 }
+
+static class PageLockedException extends Exception
+{
```
dashorst, line 232:
```suggestion
class PageLockedException extends Exception
{
```
@martin-g this is outside the surrounding class, so it is already static. The compiler also complains:

```
Error:  /home/runner/work/wicket/wicket/wicket-core/src/main/java/org/apache/wicket/page/DefaultPageLockManager.java:[231,8] modifier static not allowed here
```

martin-g: > this is outside the surrounding class

who does this ?! :-)
why not create its own class ? because you will need to add a license at the top ? :-)

dashorst: 🖐️ 

The exception has no reuse outside the DefaultPageLockManager, I thought about making it a local class, but then you get those enclosures in your stack traces, which I don't like. So this should get both requirements: no strange stack trace names and no reuse.

PR #862, opened 2024-05-11
feat: configurable transformer factory

klopfdreh:
Make the javax.xml.transform.TransformerFactory configurable so that features can be set.

### DIFF
--- a/wicket-util/src/main/java/org/apache/wicket/util/resource/XSLTResourceStream.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/resource/XSLTResourceStream.java
@@ -42,9 +42,6 @@ import org.apache.wicket.util.lang.Bytes;
  */
 public class XSLTResourceStream extends AbstractResourceStream
 {
-	/**
-	 * 
-	 */
 	private static final long serialVersionUID = 1L;
 	private final transient ByteArrayOutputStream out;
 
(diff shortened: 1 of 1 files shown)


wicket-util/src/main/java/org/apache/wicket/util/resource/XSLTResourceStream.java
```diff
@@ -77,7 +74,7 @@ public XSLTResourceStream(final IResourceStream xsltResource, final IResourceStr
 			javax.xml.transform.Result result = new javax.xml.transform.stream.StreamResult(out);
 
 			// create an instance of TransformerFactory
-			javax.xml.transform.TransformerFactory transFact = javax.xml.transform.TransformerFactory.newInstance();
+			javax.xml.transform.TransformerFactory transFact = getTransformerFactory();
```
martin-g, line 77: Calling an overwriteable method from the constructor is an anti-pattern - https://stackoverflow.com/questions/3404301/whats-wrong-with-overridable-method-calls-in-constructors

wicket-util/src/main/java/org/apache/wicket/util/resource/XSLTResourceStream.java
```diff
@@ -102,6 +99,16 @@ public XSLTResourceStream(final IResourceStream xsltResource, final IResourceStr
...
 	}
 
+	/**
+	 * Used to provide a configured {@link javax.xml.transform.TransformerFactory}.
+	 *
+	 * @return the transformer factory
+	 */
+	protected javax.xml.transform.TransformerFactory getTransformerFactory()
+	{
+		return javax.xml.transform.TransformerFactory.newInstance();
```
martin-g, line 109: IMO, if we are going to make this configurable by the user than we should use a more secure factory by default  and let the apps to go back to unsecure if needed.

wicket-util/src/main/java/org/apache/wicket/util/resource/XSLTResourceStream.java
klopfdreh: Then I would say that the whole processing in the constructor is not a good behavior as the factory needs to be configured in some cases. I am going to change the implementation so that the user can configure it in a different way.

klopfdreh:
As suggested with an additional constructor the default behavior is preserved and the user is able to configure the transformer factory by the new constructor.

martin-g (approved):

papegaaij:
Thank you @klopfdreh for addressing this. I had a week off. This change looks good to me.

PR #879, opened 2024-05-27
WICKET-7111: Fix/improve Application_el.utf8.properties: fix filename, replace xml…

mscoon:
File `application_el.ut8.properties` had wrong filename which was causing it to be ignored. Also fixed some comments within the file and added some properties that were defined in `application.utf8.properties`.


martin-g (approved):
We need a JIRA ticket too, for the changelog. Let me create it!

martin-g:
Thank you, @mscoon !

PR #894, opened 2024-06-18
WICKET-7112 Expand error messages in Component class

kbachl:
WICKET-7112

The error messages in the Component class have been augmented with more detailed information. These include class type, parent markup ID, parent relative path, and the original URL. This is to give a more comprehensive context surrounding the error, aiding debugging and issue resolution.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -2478,8 +2478,8 @@ public abstract class Component
 				if (getFlag(FLAG_OUTPUT_MARKUP_ID))
 				{
 					String message = String.format("Markup id set on a component that renders its body only. " +
-					                               "Markup id: %s, component id: %s, class-type: %s, parent markup id: %s, parent relative path: %s, url: %s",
-							getMarkupId(), getId(), getClass(), getParent().getMarkupId(), getParent().getClassRelativePath(), getRequest().getOriginalUrl());
+					                               "Markup id: %s, component id: %s, class-type: %s, path: %s",
+							getMarkupId(), getId(), getClass(), getPage().getPageClass() + ":" + getPageRelativePath());
 					if (notRenderableErrorStrategy == ExceptionSettings.NotRenderableErrorStrategy.THROW_EXCEPTION)
 					{
 						throw new IllegalStateException(message);
@@ -2489,8 +2489,8 @@ public abstract class Component
 				if (getFlag(FLAG_PLACEHOLDER))
 				{
 					String message = String.format("Placeholder tag set on a component that renders its body only. " +
-					                               "Component id: %s, class-type: %s, parent markup id: %s, parent relative path: %s, url: %s\", ",
-							getId(), getClass(), getParent().getMarkupId(), getParent().getClassRelativePath(), getRequest().getOriginalUrl());
+					                               "Component id: %s, class-type: %s, path: %s\", ",
+							getId(), getClass(), getPage().getPageClass() + ":" + getPageRelativePath());
 					if (notRenderableErrorStrategy == ExceptionSettings.NotRenderableErrorStrategy.THROW_EXCEPTION)
 					{
 						throw new IllegalStateException(message);
--- a/wicket-core/src/main/java/org/apache/wicket/Component.java
+++ b/wicket-core/src/main/java/org/apache/wicket/Component.java
@@ -2478,7 +2478,8 @@ public abstract class Component
 				if (getFlag(FLAG_OUTPUT_MARKUP_ID))
 				{
 					String message = String.format("Markup id set on a component that renders its body only. " +
-					                               "Markup id: %s, component id: %s.", getMarkupId(), getId());
+					                               "Markup id: %s, component id: %s, class-type: %s, parent markup id: %s, parent relative path: %s, url: %s",
+							getMarkupId(), getId(), getClass(), getParent().getMarkupId(), getParent().getClassRelativePath(), getRequest().getOriginalUrl());
 					if (notRenderableErrorStrategy == ExceptionSettings.NotRenderableErrorStrategy.THROW_EXCEPTION)
 					{
 						throw new IllegalStateException(message);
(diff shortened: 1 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/Component.java
```diff
@@ -2478,7 +2478,8 @@ protected final void internalRenderComponent()
 				if (getFlag(FLAG_OUTPUT_MARKUP_ID))
 				{
 					String message = String.format("Markup id set on a component that renders its body only. " +
-					                               "Markup id: %s, component id: %s.", getMarkupId(), getId());
+					                               "Markup id: %s, component id: %s, class-type: %s, parent markup id: %s, parent relative path: %s, url: %s",
+							getMarkupId(), getId(), getClass(), getParent().getMarkupId(), getParent().getClassRelativePath(), getRequest().getOriginalUrl());
```
martin-g, line 2482: 1. I don't think we need the url.
2. Instead of ` getParent().getMarkupId(), getParent().getClassRelativePath()` could we use ` getClassRelativePath()`, i.e. `this.getClassRelativePath()` ? It would be simpler

wicket-core/src/main/java/org/apache/wicket/Component.java
```diff
@@ -2488,7 +2489,8 @@ protected final void internalRenderComponent()
 				if (getFlag(FLAG_PLACEHOLDER))
 				{
 					String message = String.format("Placeholder tag set on a component that renders its body only. " +
-					                               "Component id: %s.", getId());
+					                               "Component id: %s, class-type: %s, parent markup id: %s, parent relative path: %s, url: %s\", ",
+							getId(), getClass(), getParent().getMarkupId(), getParent().getClassRelativePath(), getRequest().getOriginalUrl());
```
martin-g, line 2493: Same as above

martin-g (commented):
We need a Jira ticket too. For the changelog.

kbachl:
https://issues.apache.org/jira/browse/WICKET-7112 is created. 
I'd like to keep the URL logged as it has helped me a lot in many loggings over problems I encountered in live deploy. The amount of URL tempering by bots and/ or script kiddies is amazing and leads to all kind of warnings/ errors so I try to at least log it as often as possible since it often saves much time if an issue is required to be investigated at all.
I'll investigate the #2. of your suggestions and report back asap.

martin-g:
Usually the url is "globally logged" as a MDC item

kbachl:
MDC item?

kbachl:
When I do:
getParent().getClassRelativePath() -> parent relative path: org.apache.wicket.settings.ExceptionSettingsNotRenderableErrorStrategyTest$RenderBodyOnlyTestPage
(I can click in my IDE :) )
getClassRelativePath() -> class relative path: org.apache.wicket.markup.html.basic.Label:test 
(it only shows me the Class I already know but not where I put it into it as the "test" is the markup ID and if this is from repeater it wont help me much. All I know then is I set it on a Label in a Repeater anywhere on my page)

martin-g:
> MDC item?

https://www.baeldung.com/mdc-in-log4j-2-logback

martin-g:
Does `getPageRelativePath()` give better results ?

martin-g:
> getParent().getClassRelativePath() -> parent relative path: org.apache.wicket.settings.ExceptionSettingsNotRenderableErrorStrategyTest$RenderBodyOnlyTestPage

This is a private case (a chance) because you debug a component that is a direct child of the page. 
We need to make this useful for any component in the hierarchy.

kbachl:
Regarding the MDC: that was way too simple to be used... I'll remove the URL 👍 

Regarding the "getClassRelativePath() -> class relative path: org.apache.wicket.markup.html.basic.Label:test"
This seems quite useless info to me, I still have no clue about the parent of the component where it has been placed at all. Every component has a parent IMHO, worst case its the page itself. So getParent().getClassRelativePath() shows me *where* the troublemaker is located at, while with "this.getClassRelativePath()" I only know it comes from a Label with id test in our test case?

kbachl:
PS: getPageRelativePath leads to a simple "test", that is because of the only item it is on the page but I dont think this will be helpful at all. I still prefer getParent().getClassRelativePath() since its better to have more detailed info than less...

martin-g:
I this we need: `getPageClass() + ":" + getClassRelativePath()`

kbachl:
getPage().getClass() + ":" + getClassRelativePath() ->
if I have Page A and then 3 levels of components before the failing one I only will get 
A.class:org.apache.wicket.markup.html.basic.Label:test

But not helpful if it fails on Parent LabelRepeaterDummy.class since I never know what the parent is that places and configures the component Label:rfb124 (rfb124 = auto ID from repeater);
So I know its on Page A and a Label:rfb124  but  cant soo where its really at.

I used getPageRelativePath() instead in latest commit.

wicket-core/src/main/java/org/apache/wicket/Component.java
```diff
@@ -2478,7 +2478,8 @@ protected final void internalRenderComponent()
 				if (getFlag(FLAG_OUTPUT_MARKUP_ID))
 				{
 					String message = String.format("Markup id set on a component that renders its body only. " +
-					                               "Markup id: %s, component id: %s.", getMarkupId(), getId());
+					                               "Markup id: %s, component id: %s, class-type: %s, path: %s",
```
martin-g, line 2481:
```suggestion
					                               "Markup id: %s, component id: %s, type: %s, path: %s",
```

wicket-core/src/main/java/org/apache/wicket/Component.java
```diff
@@ -2488,7 +2489,8 @@ protected final void internalRenderComponent()
 				if (getFlag(FLAG_PLACEHOLDER))
 				{
 					String message = String.format("Placeholder tag set on a component that renders its body only. " +
-					                               "Component id: %s.", getId());
+					                               "Component id: %s, class-type: %s, path: %s\", ",
```
martin-g, line 2492:
```suggestion
					                               "Component id: %s, type: %s, path: %s\", ",
```

wicket-core/src/test/java/org/apache/wicket/settings/ExceptionSettingsNotRenderableErrorStrategyTest.java
```diff
@@ -115,8 +115,9 @@ private void assertWicketTagException(WicketRuntimeException wrx) {
     }
 
     private void assertRenderBodyOnlyException(final WicketRuntimeException wrx) {
-        assertThat(wrx.getCause().getMessage(), is(equalTo("Markup id set on a component that renders its "
-                                                           + "body only. Markup id: test1, component id: test.")));
+        assertThat(wrx.getCause().getMessage(), is(equalTo("Markup id set on a component that renders its body only. " +
+                "Markup id: test1, component id: test, class-type: class org.apache.wicket.markup.html.basic.Label, " +
```
martin-g, line 119:
```suggestion
                "Markup id: test1, component id: test, type: class org.apache.wicket.markup.html.basic.Label, " +
```

martin-g (approved):

martin-g:
Thank you, @kbachl !

PR #981, opened 2024-10-02
WICKET-7124 Use `#[[...]]#` for snippet literals in Velocity templates

martin-g:
https://velocity.apache.org/engine/2.3/user-guide.html#literals

### DIFF
--- a/archetypes/quickstart/src/main/resources/archetype-resources/pom.xml
+++ b/archetypes/quickstart/src/main/resources/archetype-resources/pom.xml
@@ -24,7 +24,7 @@
 	<artifactId>${artifactId}</artifactId>
 	<packaging>war</packaging>
 	<version>${version}</version>
-	#literal()
+	#[[
 	<!-- TODO project name  -->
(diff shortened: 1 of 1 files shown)


martin-g:
I am not sure why the `literal` is really needed though. The archetype works even without it.

losmurfs:
This still does not work. When will this work again?

losmurfs:
> I am not sure why the `literal` is really needed though. The archetype works even without it.

Which version of mvn and java is needed for this to work without the literal changes?

martin-g:
> This still does not work. When will this work again?

Define "does not work"!

martin-g:
> > I am not sure why the `literal` is really needed though. The archetype works even without it.
> 
> 
> 
> Which version of mvn and java is needed for this to work without the literal changes?

This is a question for the Maven team. 
We were also surprised that the old syntax doesn't work. 
The Maven archetype plugin upgraded Velocity to some version where the old syntax is no more supported.

PR #1022, opened 2024-10-30
Wicket 7125 set bean manager

pedrosans:
New CdiConfiguration.setBeanManager method enabling unit tests to configure cdi-unit's BeanManager in the application


bitstorm:
I guess we also need to create a separate branch for 10.x since this feature is the first targeted for version 11. No bid deal, I can do it in the next days before merging this PR

martin-g:
I see no reason to keep this only for 11.x. 
@bitstorm What is the API break for 10.x ?

bitstorm:
@martin-g my concern is if it is legit introducing a new module for API compatibility. Maybe we can use japicmp-maven-plugin to test if this change is acceptable.

martin-g:
The new module is only for tests. 
The tests from one module are extracted into their own module. I see no problems here.

martin-g:
@bitstorm  In my humble opinion `branch-10.x` should be deleted and `master` should be used for 10.x releases until we really have an API break that requires the extra branches.
There are not many active developers at the moment and maintaining several branches just adds load.

bitstorm:
sorry @martin-g , I've totally lost your last comment. Yes, I think you are right about branches. I will set back master branch as 10.x branch

bitstorm:
@pedrosans @martin-g I think we can proceed with merge.

bitstorm:
Thank you @pedrosans !

PR #1030, opened 2024-11-04
WICKET-7024 decode PackageResourceReference URL attributes only insid…

pedrosans:
changes in the ticket Wicket-7024 to the wicket-8.x branch


pedrosans:
The PackageResource handling in Wicket 9 and forward have diverged enough that the tests in this pull request won't work, even after reverting the changes in the core classes.

PR #1033, opened 2024-11-14
WICKET-3899 testing Form#wantSubmitOnParentFormSubmit in the form

pedrosans:
visiting

 - removing traversal signaling dead code
 - test if form validator dependent components are enabled and form participants
 - internalMarkFormComponentsValid in postorder
 - updateNestedFormComponentModels in postorder
 - internalOnValidateModelObjects in porstorder

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
@@ -741,7 +741,7 @@ public class Form<T> extends WebMarkupContainer
 	@Override
 	public final void onRequest()
 	{
-		onFormSubmitted(null);
+		onFormSubmitted(findSubmitter());
 	}
 
 	/**
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
@@ -65,10 +65,7 @@ import org.apache.wicket.util.string.PrependingStringBuffer;
 import org.apache.wicket.util.string.Strings;
 import org.apache.wicket.util.string.interpolator.MapVariableInterpolator;
 import org.apache.wicket.util.value.LongValue;
-import org.apache.wicket.util.visit.ClassVisitFilter;
-import org.apache.wicket.util.visit.IVisit;
-import org.apache.wicket.util.visit.IVisitor;
-import org.apache.wicket.util.visit.Visits;
+import org.apache.wicket.util.visit.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -166,22 +163,9 @@ public class Form<T> extends WebMarkupContainer
 		@Override
 		public void component(final FormComponent<?> formComponent, final IVisit<Void> visit)
 		{
-
-			Form<?> form = formComponent.getForm();
-			if (!form.isVisibleInHierarchy() || !form.isEnabledInHierarchy())
-			{
-				// do not validate formComponent or any of formComponent's children
-				visit.dontGoDeeper();
-				return;
-			}
-
 			if (formComponent.isVisibleInHierarchy() && formComponent.isEnabledInHierarchy())
 			{
 				validate(formComponent);
-			}
-			if (formComponent.processChildren() == false)
-			{
-				visit.dontGoDeeper();
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
```diff
@@ -757,7 +741,7 @@ public final boolean isSubmitted()
 	@Override
 	public final void onRequest()
 	{
-		onFormSubmitted(null);
+		onFormSubmitted(findSubmitter());
```
renoth, line 744: This seems a bit problematic. We have failing Tests after upgrading from 10.2. to 10.4 and the reason seems to be that while the submitter is calculated the same as before, there is no check for visibility anymore (like in Line 810ff) and the code does run straight to Line 839 instead of Line 810 because the submitter is not null anymore when checking in line 801

renoth: This might only influence Tests because in reality the user could not click a button that is not visible

martin-g: The best would be to send a PR with a fix and a test!

renoth: https://github.com/apache/wicket/pull/1092

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
```diff
@@ -2059,13 +2020,13 @@ protected final void validateFormValidator(final IFormValidator validator)
 				}
 				// check if the dependent component is visible and is attached to
 				// the page
-				else if (!isFormComponentVisibleInPage(dependent))
+				else if (!dependent.isVisibleInHierarchy() || !dependent.isEnabledInHierarchy() || !dependent.isFormParticipant())
```
mira-silhavy, line 2023: Hi @pedrosans, addition of `!dependent.isEnabledInHierarchy()` makes our migration to Wicket 10 problematic. We have many `FormValidator`s with multiple fields and some are conditionally enabled/disabled, but their values still participate in validation logic.
Following this new logic we'd have to modify `getDependentFormComponents` and check for each field being enabled/disabled.
I don't see any particular benefit by adding `isEnabledInHierarchy` condition.

What was the reason behind this change? Because we'd like this to work without chaning every validator and making sure it follows this `isEnabledInHierarchy` logic.

pedrosans: Hi @mira-silhavy, the idea is to prevent a formvalidator to pass or not based on "dirty" data. The disabled component it depends upon was not processed during the request validating the formvalidator.

pedrosans: > what do you mean by "dirty" data?
Any value returned by dependent formComponent#getInput() or #getConvertedInput() will have not been validated during the form processing before the formvalidator check. wicket-core for instance has formvalidators accessing dependent form component values without checking if they are valid or not, because it should be safe to assume they are valid inside IFormValidator#validate execution.

mira-silhavy: What solution would you recommend if reverting the part of the condition to allow also disabled components is not an option?

papegaaij: IMHO an `IFormValidator` should be executed if at least one of the dependent components is enabled. It should not be necessary for all components to be enabled. We've been bitten by this problem as well.

Suppose you have a form where you can enter a certain property in 2 ways. For example, a form where you can upload a certificate with private key, either in PKCS#12 format (1 single file) or in PEM format (2 files). You can toggle these fields via radio buttons, so only one of the 2 formats is visible at a given moment. In this form, there's also a field for the domain name, which needs to match with the provided certificate. The easiest way would be to have an `IFormValidator` with all these fields as dependent components. However, this does not work as either the PKCS#12 or the PEM fields are not visible, so your validator is never executed.

A problem I do see is that disabled components do not participate in form processing, so their converted input is always null. Your validator needs to be able to deal with this properly. In the example above, this is not an issue, because the validator simply checks the state of the radio buttons before reading either the PKCS#12 or PEM fields.

mira-silhavy: @pedrosans @papegaaij shall I open a PR that uses Wicket 9 behavior in this situation? Or with flag?

papegaaij: @mira-silhavy Yes, that would be great and also make it easier to discuss this change.

mira-silhavy: @papegaaij I opened a PR here https://github.com/apache/wicket/pull/1186, currently only removes the `isEnabledInHierarchy`

PR #1034, opened 2024-11-15
WICKET-7131 Improved accessibility and screen reader support

strido:
Improved accessibility support with the following changes

- The up and down arrow keys in the auto-complete dropdown should not move the cursor to the start or end of the text field
- aria-expanded should be set on the input element to indicate whether the dropdown is expanded or collapsed
- aria-selected should be set on the element that is currently selected in the dropdown
- aria-activedescendant attribute should be used on the input field, pointing to the currently selected item in the dropdown. This allows screen readers to track the active selection in the list.
- aria-live should not be set on the dropdown as it will cause a screen reader to read out everything in the dropdown, which disrupts the navigaition workflow

### DIFF
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/wicket-autocomplete.js
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/wicket-autocomplete.js
@@ -469,7 +469,6 @@
 
 			if (triggerChangeOnHide) {
 				triggerChangeOnHide = false;
-				var input = Wicket.$(ajaxAttributes.c);
 				isTriggeredChange = true;
 				jQuery(input).trigger('change');
 			}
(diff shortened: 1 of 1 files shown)


martin-g (approved):

martin-g:
```
   ../../wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/wicket-autocomplete.js
[INFO]     472 |                var input = Wicket.$(ajaxAttributes.c);
[INFO]                              ^ 'input' is already defined.
[INFO]     626 |            element.firstChild.role = "listbox"
[INFO]                                                         ^ Missing semicolon.
[INFO]     677 |                    node.setAttribute("aria-posinset", i + 1)
[INFO]                                                                       ^ Missing semicolon.
[INFO]     678 |                    node.setAttribute("aria-setsize", elementCount)
[INFO]                                                                             ^ Missing semicolon.
```

strido:
The code style issues should be fixed now

PR #1040, opened 2024-11-19
WICKET-7133 Ability prevent autocomplete to select items when pressing tab

strido:


strido:
This is not relevant for us, I misunderstood the requirements and will rephrase the jira issue.

PR #1041, opened 2024-11-20
WICKET-7133 Ability move focus back to the autocomplete field when selecting an item using the Tab key

strido:
We need the ability to prevent the autocomplete from moving the focus to the next component when the user press "Tab". The marked item should be selected but the focus should go back to the autocomplete input field.

### DIFF
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/AutoCompleteSettings.java
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/AutoCompleteSettings.java
@@ -74,7 +74,7 @@ public final class AutoCompleteSettings implements IClusterable
 
 	private int minInputLength = 1;
 
-	private boolean focusInputOnTabSelection = false;
+	private TabBehavior tabBehavior = TabBehavior.SELECT_FOCUS_NEXT_ELEMENT;
 
 	/**
 	 * Indicates whether the first item in the list is automatically selected when the autocomplete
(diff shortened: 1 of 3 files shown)


strido:
The code style issues should be fixed now

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/AutoCompleteSettings.java
```diff
@@ -379,4 +381,30 @@ public AutoCompleteSettings setMinInputLength(int minInputLength)
...
 	}
+
+	/**
+	 * Indicates how the Tab key should be handled when having an item in the autocomplete list
+	 * selected.
+	 *
+	 * @return <code>true</code> if the focus should return to the input field, <code>false</code>
+	 * moves the focus to the next component.
+	 */
+	public boolean shouldFocusInputOnTabSelection()
```
martin-g, line 392: IMO it would be better to use an enum here instead of boolean.
This way it would be:
1) more clear what the values mean
2) easier to add more values/states later

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/wicket-autocomplete.js
```diff
@@ -183,6 +184,12 @@
 
 							hideAutoComplete();
 
+							if (cfg.focusInputOnTabSelection && keyCode === KEY_TAB) {
```
martin-g, line 187: If you like my idea of using an enum, then here the code would look like:
```suggestion
							if (cfg.focusInputOnTabSelection == "some_enum_value" && keyCode === KEY_TAB) {
```

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/AutoCompleteSettings.java
strido: I like the idea!

I will try look into it during the day.

strido: I've pushed an update, ut are a bit unsure about the naming

wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/AutoCompleteSettings.java
```diff
@@ -379,4 +381,55 @@ public AutoCompleteSettings setMinInputLength(int minInputLength)
...
+	{
+		this.tabBehavior = tabSelectBehavior;
+		return this;
+	}
+
+	/**
+	 * A behavior that can be used to control how the Tab key should be handled when having an item
+	 * in the autocomplete list is marked.
+	 */
+	public enum TabBehavior {
```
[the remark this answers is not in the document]
strido: That seems like a better name!

I have renamed it and pushed the changes.

PR #1049, opened 2024-11-28
WICKET-7135 Work on deprecation warnings

aldaris:
This PR attempts to address most of the deprecated call warnings shown during compilation time. There are a still couple of them left that are a bit more tricky to handle:
* DebugSettings#isOutputMarkupContainerClassName usage in Page - the output strategy is not simple to implement given that Page currently only embeds the classname as a separate comment element, not sure what it should tie to otherwise
* Cookie#setComment, #setVersion usages - don't think anything really needs to be done about these
* CachingResourceStreamLocator#locate - there is a missing @Deprecated annotation. Without proper JavaDoc explanation I didn't want to go down a road for this one
* AccessController usages in JavaSerializer - could just replace them with Runnable#run I suppose, but wasn't sure whether security manager support was there for something specific


bitstorm (changes requested):
Looks good! Thank you! Do you have a JIRA account to create a related issue?

PR #1072, opened 2025-01-07
Wicket 7138 support ur lfragment with page parameters

bitstorm:


wicket-request/src/main/java/org/apache/wicket/request/mapper/parameter/PageParameters.java
```diff
@@ -615,6 +633,14 @@ public String toString()
 				str.append('[').append(entry.getValue()).append(']');
 			}
 		}
+
+		if (str.length() > 0)
+		{
+			str.append(", ");
+		}
+
+		str.append("fragment=").append('\'').append(fragment).append('\'');
```
martin-g, line 642: Do we want to print anything here if the fragment is null ?

bitstorm: I would say to avoid printing anything if fragment is null  to maintain consistency with the rest of the toString code. I fill add a null pointer check.

PR #971, opened 2024-09-23
WICKET-7123 default form submit can now be set one per form

1azyman:
- default form submit can now be set one per form (not form hierarchy)
- hidden input submit in root form now not needed, everything handled in JS
- click invocation mostly reused old part of javascript
- tests green

One thing that I'm missing but couldn't figure out was to write test with form hierarchy (parent/child form) with inputs and simulate filling text field and hitting "enter"...

PR #1076, opened 2025-01-14
WICKET-7140 fix for textarea submit on enter (now ctrl+enter)

1azyman:
https://issues.apache.org/jira/browse/WICKET-7140

Follow up to #971.


wicket-core-tests/src/test/java/org/apache/wicket/markup/html/form/FormHierarchyDefaultButtonTestPage.java
```diff
@@ -33,6 +33,8 @@ public class FormHierarchyDefaultButtonTestPage extends WebPage {
     public final Button childSubmit;
     public final TextField<?> parentInput;
     public final TextField<?> childInput;
+	public final TextArea<?> parentTextarea;
+	public final TextArea<?> childTextarea;
```
theigl, line 37: Please fix the indentation here and also in the HTML files.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
```diff
@@ -1273,9 +1273,11 @@ protected void addDefaultSubmitButtonHandler(IHeaderResponse headerResponse)
 		String submitId = component.getMarkupId();
 
 		AppendingStringBuffer script = new AppendingStringBuffer();
-		script.append("var b = document.getElementById('").append(submitId).append("');");
-		script.append("if (window.getComputedStyle(b).visibility === 'hidden') return;");
-		script.append("if (event.which == 13) {");
+		script.append("if (event.which != 13) return true;");
+		// text area will submit on CTRL+Enter
+		script.append("if (event.target.tagName.toLowerCase() === 'textarea' && !event.ctrlKey) return true;");
```
theigl, line 1278: I don't think this is enough. I'm pretty sure we should to the same for `contenteditable` elements used by most WYSIWYG editors. How was form submit prevented for these elements before you submitted your original PR?

theigl (changes requested):

martin-g:
For some reason the CI didn't run.
I will close/reopen to see whether it will help!

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
1azyman: good idea about that contenteditable. AFAIK previously there were no special handling for textarea/contenteditabl.
It was handled via hidden input that was defined only in root form with onclick event that was handled by script in `Form.addDefaultSubmitButtonHandler()`. Currently it's handled by catching keypress event handler in each relevant form.

1azyman: What do you mean by customizable? What options would you expect/prefer.

I'm currently trying to figure out whether form submit can be done via click on hidden that are still in hierarchy.

1azyman:
I've added improvement that covers cases:

- contenteditable => will not submit
- textarea => will submit on ctrl+enter
- input fields => on enter

however it still seems kinda flimsy. I'll wait for @solomax ideas on how we could configure it better.

solomax:
> however it still seems kinda flimsy. I'll wait for @solomax ideas on how we could configure it better.

plain HTML textarea doesn't handle `Enter`, `Ctrl+Enter`, `Shift+Enter`, `Super+Enter` key presses
And this behavior is expected
Existing Wicket applications are written and tested having this behavior in mind

So I believe there should be at list an option (or better default) to have this "as is"

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.js
```diff
@@ -1,6 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 function handleDefaultSubmit(event, submitId, keyStroke) {
-    const tag = event.target.tagName.toLowerCase();
-    const contenteditable = event.target.getAttribute('contenteditable') === 'true';
+    var tag = event.target.tagName.toLowerCase();
```
solomax, line 19: I believe better to use `let` here :)

solomax (approved):

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.js
1azyman: True, `const` was probably the best option, however current build fails with `const` and `let` with following error: `[ERROR] JSC_LANGUAGE_FEATURE. This language feature is only supported for ECMASCRIPT_2015 mode or better: let declaration.`
Can I updated javascript languageIn version from `ECMASCRIPT3` to at least `ECMASCRIPT_2015` in `resources-optimizer-maven-plugin`?

martin-g: > Can I updated javascript languageIn version from `ECMASCRIPT3` to at least `ECMASCRIPT_2015` in `resources-optimizer-maven-plugin`?

I think it is about time!

theigl:
Let's take a step back and think about this once again.

1. The original behavior prior to your PR #971 was that forms could not be submitted via ENTER from multiline fields. This is what I would consider "default" browser behavior.
2. Your first PR changed that and forms can now be submitted via ENTER from multiline fields. This is problematic.
3. This PR now attempts to solve this by making the submit handling configurable and restoring the original behavior.

For me, this PR is doing too many things. I think we should simply restore the original behavior. Submitting a form from multiline fields is something that is rarely needed, and if it is needed it can be easily added to a textarea as a behavior. We are using this in our project:

```java
public final class KeyboardSubmitBehavior extends WiQueryEventBehavior {

	private static final String SUBMIT_SELECTOR = "input[type='submit'],button[type='submit'],[name=':submit']";

	public KeyboardSubmitBehavior(final KeyboardEvent keyboardEvent) {
		super(new Event(KeyboardEvent.KEYDOWN) {
			@Override
			public JsScope callback() {
				return JsScopeEvent.quickScope("if ((event.which == 10 || event.which == 13) && event.ctrlKey) { " +
						"$(this).closest('form').find(\"" + SUBMIT_SELECTOR + "\").trigger('click'); return false; " +
						"}");
			}
		});
	}
}
```

Then we don't need a separate `form.js` file and most of the complexity goes away.

What do you think @solomax and @martin-g?

1azyman:
> Let's take a step back and think about this once again.
> 
>     1. The original behavior prior to your PR [WICKET-7123 default form submit can now be set one per form #971](https://github.com/apache/wicket/pull/971) was that forms could not be submitted via ENTER from multiline fields. This is what I would consider "default" browser behavior.

No. 
Original problem was that in form hierarchy (eg. parent form with some field and containing another form with more fields), default submit is always executed on last form rendered on page - which is obviously incorrect, and this limitation was also mentioned in docs - see comment here [Form.java](https://github.com/apache/wicket/blob/783722121d55530813f21e5e6bc14e9955777e5c/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java#L505).
I wanted to improve it with a little bit of javascript - remove limitation on form hierarchy.

martin-g:
I do agree with @theigl that we should not try to handle submit for textareas. Especially if it could break custom behaviors.
If an application needs to submit on Ctrl+Enter then it is easy enough to add a Behavior like the one @theigl shared above to do it.

The new Form.js introduces a function in the global space, i.e. `window.**`. If it is really needed then it should be in the `Wicket.Form` namespace. This way it could be overridden by the application if needed.

martin-g:
A warning about `Wicket.Form` - see https://github.com/apache/wicket/blob/c91e34a90234bc091939c27fcc1716f710af5274/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js#L1128
It might be tricky to add a new function to it!
Currently `Wicket.Form` is contributed by wicket-ajax.js, i.e. only then there is an Ajax behavior in the page.
I guess the new function is needed even when there is no need of Ajax.

It would have been better to namespace everything in wicket-ajax.js in Wicket.Ajax.** but it is too late now ...

1azyman:
I see, I just put it into separate file for better readability and not to create unnecessary duplicated JS code on rendered page if there are multiple forms. I can still put it back (embed) it in `Form.java` if necessary.

theigl:
> > Let's take a step back and think about this once again.
> > ```
> > 1. The original behavior prior to your PR [WICKET-7123 default form submit can now be set one per form #971](https://github.com/apache/wicket/pull/971) was that forms could not be submitted via ENTER from multiline fields. This is what I would consider "default" browser behavior.
> > ```
> 
> No. Original problem was that in form hierarchy (eg. parent form with some field and containing another form with more fields), default submit is always executed on last form rendered on page - which is obviously incorrect, and this limitation was also mentioned in docs - see comment here [Form.java](https://github.com/apache/wicket/blob/783722121d55530813f21e5e6bc14e9955777e5c/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java#L505). I wanted to improve it with a little bit of javascript - remove limitation on form hierarchy.

Sorry I wasn't very clear. All your other changes are good as far as I can tell. I'm only talking about how the ENTER key behaved before and after your changes.

1azyman:
> Sorry I wasn't very clear. All your other changes are good as far as I can tell. I'm only talking about how the ENTER key behaved before and after your changes.

No problem at all. Currently there's no submit in textarea or contenteditable by default - even when there's setDefaultButton() used. If user wants, then Form#setMultiRowInputDefaultSubmitKeyStroke() can be used to turn on submit in textarea/contenteditable with different options.

martin-g:
> Or maybe create custom namespace like Wicket.Internal.Form or something like that?

I like it being in its own .js file! Feel free to do it whatever way is easier for you!
Maybe `jQuery.extend()` is smart enough to not override the code from Form.js ... If it is smart then you will need to make Form.js smart too, e.g. with something like https://github.com/apache/wicket/blob/c91e34a90234bc091939c27fcc1716f710af5274/wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js#L31-L33

1azyman:
All right. I'll remove textarea/contenteditable magic and configuration and embedd JS back again.

pom.xml
```diff
@@ -1267,6 +1267,7 @@
 								</excludes>
 							</resourcesSet>
 						</resourcesSets>
+                        <languageIn>ECMASCRIPT_2015</languageIn>
```
theigl, line 1270: @martin-g: Do you think it's a good idea to do such a change in a bug fix that should ideally be released right away?

theigl (approved):
Apart from my comment on the JS language change, LGTM.

pom.xml
1azyman: I'd say it's safe, since it only updates level of javascript that can be used in wicket code base. But I see your point, therefore I'll revert also `const/let` to `var`

martin-g: I think the change is pretty safe these days!
But yes, it could be done in its own PR/ticket.

theigl:
Thank you @1azyman! A

nd sorry for all the back and forth, but I think it was important to do a minimal fix now that solves the problem and has a low chance of introducing new ones ;)

1azyman:
@theigl no problem at all, glad to help. Much better when there's a response, discussion and things are moving.

PR #1092, opened 2025-01-29
Fix Form#onFormSubmitted not considering if submitter is disabled or invisible

renoth:


martin-g (approved):
The code formatting makes it harder to review ...

PR #1093, opened 2025-01-31
WICKET-7145: added `@Nonnull` to parameters

jstuyts:
Added `@Nonnull` to parameters that are checked for `null`. Also included some method overloads/variants.

In addition found some bugs in `null` checks.

See https://issues.apache.org/jira/browse/WICKET-7145 for details.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/request/resource/caching/FilenameWithVersionResourceCachingStrategy.java
+++ b/wicket-core/src/main/java/org/apache/wicket/request/resource/caching/FilenameWithVersionResourceCachingStrategy.java
@@ -18,6 +18,8 @@ package org.apache.wicket.request.resource.caching;
 
 import java.util.regex.Pattern;
 
+import jakarta.annotation.Nonnull;
+
 import org.apache.wicket.request.cycle.RequestCycle;
 import org.apache.wicket.request.http.WebResponse;
 import org.apache.wicket.request.resource.AbstractResource;
@@ -83,8 +85,8 @@ public class FilenameWithVersionResourceCachingStrategy implements IResourceCach
 	 * 
 	 * @see #FilenameWithVersionResourceCachingStrategy(org.apache.wicket.request.resource.caching.version.IResourceVersion) 
 	 */
-	public FilenameWithVersionResourceCachingStrategy(String versionPrefix,
-	                                                  IResourceVersion resourceVersion)
+	public FilenameWithVersionResourceCachingStrategy(@Nonnull String versionPrefix,
+			                                          @Nonnull IResourceVersion resourceVersion)
 	{
 		this.resourceVersion = Args.notNull(resourceVersion, "resourceVersion");
 		this.versionPrefix = Args.notEmpty(versionPrefix, "versionPrefix");
--- a/wicket-core/src/main/java/org/apache/wicket/settings/JavaScriptLibrarySettings.java
+++ b/wicket-core/src/main/java/org/apache/wicket/settings/JavaScriptLibrarySettings.java
@@ -16,6 +16,8 @@
  */
 package org.apache.wicket.settings;
 
+import jakarta.annotation.Nonnull;
+
 import org.apache.wicket.ajax.WicketAjaxJQueryResourceReference;
 import org.apache.wicket.request.resource.ResourceReference;
 import org.apache.wicket.resource.JQueryResourceReference;
@@ -54,7 +56,7 @@ public class JavaScriptLibrarySettings
 	 *            wicket-event and wicket-ajax
 	 * @return {@code this} object for chaining
 	 */
-	public JavaScriptLibrarySettings setJQueryReference(ResourceReference jQueryReference)
+	public JavaScriptLibrarySettings setJQueryReference(@Nonnull ResourceReference jQueryReference)
 	{
 		this.jQueryReference = Args.notNull(jQueryReference, "jQueryReference");
 		return this;
@@ -73,7 +75,7 @@ public class JavaScriptLibrarySettings
 	 *            a reference to the implementation of wicket-ajax.js
 	 * @return {@code this} object for chaining
 	 */
-	public JavaScriptLibrarySettings setWicketAjaxReference(ResourceReference wicketAjaxReference)
+	public JavaScriptLibrarySettings setWicketAjaxReference(@Nonnull ResourceReference wicketAjaxReference)
 	{
 		this.wicketAjaxReference = Args.notNull(wicketAjaxReference, "wicketAjaxReference");
 		return this;
--- a/wicket-native-websocket/wicket-native-websocket-core/pom.xml
+++ b/wicket-native-websocket/wicket-native-websocket-core/pom.xml
@@ -34,6 +34,10 @@
 		<osgi.import.package>!java*,!kotlin*,!sun.nio.ch,org.slf4j*;version="[1.7,3)",*</osgi.import.package>
 	</properties>
     <dependencies>
+		<dependency>
+			<groupId>jakarta.annotation</groupId>
+			<artifactId>jakarta.annotation-api</artifactId>
+		</dependency>
         <dependency>
             <groupId>org.apache.wicket</groupId>
             <artifactId>wicket-core</artifactId>
--- a/wicket-native-websocket/wicket-native-websocket-tester/pom.xml
+++ b/wicket-native-websocket/wicket-native-websocket-tester/pom.xml
@@ -34,6 +34,10 @@
 		<osgi.import.package>!java*,!kotlin*,!sun.nio.ch,org.slf4j*;version="[1.7,3)",*</osgi.import.package>
 	</properties>
     <dependencies>
+		<dependency>
+			<groupId>jakarta.annotation</groupId>
+			<artifactId>jakarta.annotation-api</artifactId>
+		</dependency>
         <dependency>
             <groupId>org.apache.wicket</groupId>
             <artifactId>wicket-core</artifactId>
--- a/wicket-auth-roles/pom.xml
+++ b/wicket-auth-roles/pom.xml
@@ -36,6 +36,10 @@
 	</properties>
 	<dependencies>
 		<dependency>
+			<groupId>jakarta.annotation</groupId>
+			<artifactId>jakarta.annotation-api</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.apache.wicket</groupId>
 			<artifactId>wicket-core</artifactId>
 		</dependency>
--- a/wicket-auth-roles/src/main/java/module-info.java
+++ b/wicket-auth-roles/src/main/java/module-info.java
@@ -20,6 +20,7 @@ module org.apache.wicket.auth.roles {
     requires org.apache.wicket.util;
     requires org.apache.wicket.request;
     requires org.apache.wicket.core;
+    requires jakarta.annotation;
 
     exports org.apache.wicket.authroles.authentication;
     exports org.apache.wicket.authroles.authentication.pages;
--- a/wicket-auth-roles/src/main/java/org/apache/wicket/authroles/authorization/strategies/role/AbstractRoleAuthorizationStrategy.java
+++ b/wicket-auth-roles/src/main/java/org/apache/wicket/authroles/authorization/strategies/role/AbstractRoleAuthorizationStrategy.java
@@ -16,6 +16,8 @@
  */
 package org.apache.wicket.authroles.authorization.strategies.role;
 
+import jakarta.annotation.Nonnull;
+
 import org.apache.wicket.authorization.IAuthorizationStrategy;
 import org.apache.wicket.util.lang.Args;
 
@@ -36,7 +38,7 @@ public abstract class AbstractRoleAuthorizationStrategy extends IAuthorizationSt
 	 * @param roleCheckingStrategy
 	 *            the authorizer delegate
 	 */
-	public AbstractRoleAuthorizationStrategy(IRoleCheckingStrategy roleCheckingStrategy)
+	public AbstractRoleAuthorizationStrategy(@Nonnull IRoleCheckingStrategy roleCheckingStrategy)
 	{
 		Args.notNull(roleCheckingStrategy, "roleCheckingStrategy");
 		this.roleCheckingStrategy = roleCheckingStrategy;
--- a/wicket-auth-roles/src/main/java/org/apache/wicket/authroles/authorization/strategies/role/Roles.java
+++ b/wicket-auth-roles/src/main/java/org/apache/wicket/authroles/authorization/strategies/role/Roles.java
@@ -19,6 +19,8 @@ package org.apache.wicket.authroles.authorization.strategies.role;
 import java.util.Collection;
 import java.util.HashSet;
 
+import jakarta.annotation.Nonnull;
+
 import org.apache.wicket.util.io.IClusterable;
 import org.apache.wicket.util.lang.Args;
 import org.apache.wicket.util.string.StringList;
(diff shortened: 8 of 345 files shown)


wicket-native-websocket/wicket-native-websocket-core/pom.xml
```diff
@@ -34,6 +34,10 @@
 		<osgi.import.package>!java*,!kotlin*,!sun.nio.ch,org.slf4j*;version="[1.7,3)",*</osgi.import.package>
 	</properties>
     <dependencies>
+		<dependency>
+			<groupId>jakarta.annotation</groupId>
+			<artifactId>jakarta.annotation-api</artifactId>
+		</dependency>
         <dependency>
```
mattrpav, line 41: Good time to fix the wicket-native-websocket-core/pom.xml formatting?

mattrpav (commented):
I'm about 1/2 through the review a couple of comments:

1. If there are null check fixes, it would be good if they were in a separate commit so those changes could potentially be back-ported.

2. Please see the code style notes around import static and end-of-file new lines.

wicket-core/src/main/java/org/apache/wicket/settings/JavaScriptLibrarySettings.java
```diff
@@ -73,9 +75,9 @@ public ResourceReference getWicketAjaxReference()
...
 	 */
-	public JavaScriptLibrarySettings setWicketAjaxReference(ResourceReference wicketAjaxReference)
+	public JavaScriptLibrarySettings setWicketAjaxReference(@Nonnull ResourceReference wicketAjaxReference)
 	{
 		this.wicketAjaxReference = Args.notNull(wicketAjaxReference, "wicketAjaxReference");
 		return this;
 	}
-}
\ No newline at end of file
+}
```
mattrpav, line 83: Code style EOF new line

wicket-core/src/main/java/org/apache/wicket/request/resource/caching/FilenameWithVersionResourceCachingStrategy.java
```diff
@@ -83,8 +85,8 @@ public FilenameWithVersionResourceCachingStrategy(IResourceVersion resourceVersi
 	 * 
 	 * @see #FilenameWithVersionResourceCachingStrategy(org.apache.wicket.request.resource.caching.version.IResourceVersion) 
 	 */
-	public FilenameWithVersionResourceCachingStrategy(String versionPrefix,
-	                                                  IResourceVersion resourceVersion)
+	public FilenameWithVersionResourceCachingStrategy(@Nonnull String versionPrefix,
+			                                          @Nonnull IResourceVersion resourceVersion)
```
mattrpav, line 89: Double check for unintended whitespace

wicket-native-websocket/wicket-native-websocket-tester/pom.xml
```diff
@@ -34,6 +34,10 @@
 		<osgi.import.package>!java*,!kotlin*,!sun.nio.ch,org.slf4j*;version="[1.7,3)",*</osgi.import.package>
 	</properties>
     <dependencies>
+		<dependency>
+			<groupId>jakarta.annotation</groupId>
+			<artifactId>jakarta.annotation-api</artifactId>
+		</dependency>
```
mattrpav, line 40: Good time to fix pom formatting?

jstuyts:
> I'm about 1/2 through the review a couple of comments:

Thanks, I'll fix everything once you're done.

> 1. If there are null check fixes, it would be good if they were in a separate commit so those changes could potentially be back-ported.

There is only 1 in production code: `OriginResourceIsolationPolicy` (search for `Checks.notNull`). The other 2 are in test code: `BaseWicketTester` and `WicketTesterTest` (search for `assertNull`). If you want, I can create a separate issue and PRs for backporting those.

> 2. Please see the code style notes around import static and end-of-file new lines.

I'll check the notes and will make sure all changed files comply with them.

theigl:
Is there a reason you went with `jakarta-annotations` and not with [JSpecify](https://jspecify.dev)? 

JSpecify is intended to be the defacto standard and Spring Framework [recently migrated](https://github.com/spring-projects/spring-framework/issues/28797) all their annotations to it. I recently annotated our main project with it but I went the other way and used JSpecify's [recommended strategy](https://jspecify.dev/docs/applying/) of applying `@NullMarked` on the package level und marking `@Nullable` parameters.

mattrpav:
> There is only 1 in production code: `OriginResourceIsolationPolicy` (search for `Checks.notNull`). The other 2 are in test code: `BaseWicketTester` and `WicketTesterTest` (search for `assertNull`). If you want, I can create a separate issue and PRs for backporting those.

That would be great, please do. I think there will be users on Wicket 9.x for a while since it is the javax.* based release.

mattrpav:
> JSpecify is intended to be the defacto standard and Spring Framework [recently migrated](https://github.com/spring-projects/spring-framework/issues/28797) all their annotations to it.

If JSpecify's approach becomes the standard, it will be in an updated release of the jakarta.annotation API. I do not see any reason to bring in a non-standard dependency when the standardized Jakarta-based solution is sufficient.

theigl:
> > JSpecify is intended to be the defacto standard and Spring Framework [recently migrated](https://github.com/spring-projects/spring-framework/issues/28797) all their annotations to it.
> 
> If JSpecify's approach becomes the standard, it will be in an updated release of the jakarta.annotation API. I do not see any reason to bring in a non-standard dependency when the standardized Jakarta-based solution is sufficient.

The whole nullability discussion has been going on for years. There are findbugs annotations, checker framework annotations, jetbrains, nullaway, javax/jakarta and many more. All of them have their advantages and disadvantages. JSpecify is an attempt (by Google, Facebook, Uber, etc) to standardize these annotations and define their exact **semantics**. 

The jakarta.annotation package is simply not enough. There is no way using only these annotations to define the default nullability of a module/package/class. What does it mean now when a parameter is **not annotated**? Does it mean it is nullable? Or is its nullability unspecified? JSpecify solves these ambiguities.

Having said that, annotating with jakarta.annotation is probably still better than nothing, though I would have preferred annotating with `@Nullable` instead.

jstuyts:
> Is there a reason you went with `jakarta-annotations` and not with [JSpecify](https://jspecify.dev)?
> 
> JSpecify is intended to be the defacto standard and Spring Framework [recently migrated](https://github.com/spring-projects/spring-framework/issues/28797) all their annotations to it. I recently annotated our main project with it but I went the other way and used JSpecify's [recommended strategy](https://jspecify.dev/docs/applying/) of applying `@NullMarked` on the package level und marking `@Nullable` parameters.

I have been a Kotlin developer for years, so I have no knowledge of recent annotation libraries. I chose the Jakarta annotations because the library was already in the BOM. But the annotations for nullability had not been used anywhere yet.

As for the other concerns:
* Something that is not annotated will indeed have an ambiguous meaning: it is nullable, or it has not been checked for non-nullability yet. Specifying at some higher level that all of the API is  nullable if marked (or the other way around) won't help here: you still need to add an annotation to each variable or method that is nullable (or not) before it matches how the code behaves. Until that has been done, the other variables and methods are ambiguous too.
* About using `@Nonnull`: in Java the default is that a reference is nullable, so for me it is more logical to state that you cannot assume the standard Java semantics.

mattrpav:
> The whole nullability discussion has been going on for years. There are findbugs annotations, checker framework annotations, jetbrains, nullaway, javax/jakarta and many more. All of them have their advantages and disadvantages. JSpecify is an attempt (by Google, Facebook, Uber, etc) to standardize these annotations and define their exact **semantics**.

There is also a huge graveyard of dead projects from large organizations that take passes at solving for short comings.

I agree, that the spec could be improved. I agree the the Jakarta Annotation Team should take this up-- if a spec doesn't meet the needs of users, it won't be viable.

> The jakarta.annotation package is simply not enough. There is no way using only these annotations to define the default nullability of a module/package/class. What does it mean now when a parameter is **not annotated**? Does it mean it is nullable? Or is its nullability unspecified? JSpecify solves these ambiguities.

If something isn't marked, it should be presumed to be 'nullable', since that is the language default. 

IMO, less noise is better. Nulls are a feature, not a bug. Null values don't use memory or require object allocations that need to be garbage collected. Annotating everything and throwing Optionals everywhere creates non-zero side effects and performance impacts.

Related Jakarta Annotation update requested:
ref: https://github.com/jakartaee/common-annotations-api/issues/145

theigl:
> If something isn't marked, it should be presumed to be 'nullable', since that is the language default.
> IMO, less noise is better. 

Marking with `@Nullable` produces significantly less noise in my experience. While it is true that at the language level almost everything is nullable, in a typical project, the majority of values are never null. I recently annotated hundreds of thousands of LOC in my main project and I would have needed at least 3-5x the amount of annotations if I went with `@Nonnull`.  
But marking as `@Nullable` only makes sense together with a `@Nullmarked` or `@ParametersAreNonnullByDefault` annotation. Otherwise it doesn't provide much value. So if we stick with `jakarta.annotation` I think that only the current `@Nonnull` approach makes sense. 

@jstuyts: Are you consuming Wicket APIs from Kotlin? If so, does adding these `@Nonnull`s already help with null-safety in Kotlin?

jstuyts:
> That would be great, please do. I think there will be users on Wicket 9.x for a while since it is the javax.* based release.

I only found the issues in 9.x. In 8.x the methods are used correctly: https://issues.apache.org/jira/browse/WICKET-7147

jstuyts:
> JSpecify is perfectly usable already but it has a couple of non-intuitive annotation usages. Your example is one of them. See https://jspecify.dev/docs/user-guide/#some-subtler-details.

Got it. I have created a branch using JSpecify: https://github.com/jstuyts/wicket/tree/add-non-null-to-parameters-jspecify

The advantages are:
* It seems like the Java ecosystem is moving to JSpecify as the semantics are more precise and there is more tooling support.
* The Kotlin compiler understands the annotations of JSpecify.

Let me know if you decide you would like to use JSpecify instead of the Jakarta annotations. I will close this PR and create a new one.

Another remark about which annotations to use. I think the annotations should be added incrementally, and both annotations (nullable and not nullable) must be added. Once everything is annotated (which can take quite some time), the decision to switch to package-level (or some other level) annotations can be made. It is then also easy to determine which of the 2 annotations results in the least amount of code. The most occurring annotation can be changed to higher-level annotations.

theigl:
> Let me know if you decide you would like to use JSpecify instead of the Jakarta annotations. I will close this PR and create a new one.

I prefer the JSpecify branch because of the advantages you mentioned!

> Another remark about which annotations to use. I think the annotations should be added incrementally, and both annotations (nullable and not nullable) must be added. Once everything is annotated (which can take quite some time), the decision to switch to package-level (or some other level) annotations can be made. It is then also easy to determine which of the 2 annotations results in the least amount of code. The most occurring annotation can be changed to higher-level annotations.

What I did in my project is this:

1. Mark a package as `@NullMarked` in `package-info.java`
2. For each class in the package:
    - Mark all parameters as `@Nullable` that IntelliJ reported as null being passed or that is null-checked inside the method
    - Mark all return values as `@Nullable` that IntelliJ reported as possibly returning null
3. Repeat until almost all packages are marked and IntelliJ shows no more warnings
4. Optional: Remove all `package-info.java` files and place a single `@NullMarked` annotation in the `module-info.java`. 

This is basically the incremental strategy that JSpecify [recommends](https://jspecify.dev/docs/applying/).

What approach we take will mostly depend on how much time we want to spend.

jstuyts:
> This is basically the incremental strategy that JSpecify [recommends](https://jspecify.dev/docs/applying/).

Yes, but Wicket is a public project. It is undesirable to have things in a state that is incomplete, I think, as this would result in a lot of errors/warnings for users of Wicket.

Doing everything at once would result in a giant PR. I think this one is already stretching the limit. I have a few more coming if this (or the JSpecify variant) is merged, and will make sure those PRs are smaller.

So I think that higher-level declarations of (non-)nullability is not an option for now.

jstuyts:
> I think we can go ahead with your current JSpecify PR and continue improving the annotations in future work.

I'll wait for what the final decision will be: Jakarta or JSpecify.

> The only minor concern I still have with this PR is that we are adding `@NonNull`. From my experience (and also what JSpecify recommends and large projects like Spring are doing) `@Nullable` makes more sense in the long run. At some point, we will have to remove all these `@NonNull` annotations. But this should be a trivial change.

Like I said before: once everything is annotated, it is easy to decide which of the 2 annotations can stay.

> And then there is the question of how we continue with this and what we require in future PRs. Is our goal to completely annotate the whole Wicket codebase? Do we add nullability annotations whenever we touch existing code? Do we require them in PRs? Or is this a one-time effort to slightly improve the developer experience?

A recommendation to add the annotations when adding or modifying code is the least that can be done. Maybe split PRs: first a PR that adds the annotations to existing code, and then a PR that contains clearer code for the modifications because of the added annotations. This should make reviews easier.

Annotating everything would be best, but there are some tricky situations, that force clients that can never see a `null` to work with nullable variables. Here are a few I found so far:

* Theoretically `getComponent()` of a behavior can return `null` (for example, call it before binding the behavior), but I think that practically `getComponent()` is never called when the behavior is unbound. There are a couple of options here, but I want to discuss those when I create an issue for behaviors.
* The `IValidatable` passed to `IValidator` will not have a `null` value, unless the validator implements `INullAcceptingValidator`.

martin-g:
> I'll wait for what the final decision will be: Jakarta or JSpecify.

For me both are meh. 
- The annotations will add a lot of noise
- The changes will make cherry-picking harder, unless someone volunteers to backport them to 9.x too
- I am not coding Java since several years (happy user of `std::result::Result<T, E>`!) but in my past experience the lack of JSR-305 usage and NPEs caused by Wicket APIs was not a (big) problem

But others decided to use it in Wicket then I'd vote for JSpecify.

jstuyts:
> I'm also unsure about proceeding with this if we don't have a concrete and useful final result in mind.

The desired end result would be that for every variable and method in Wicket it is known whether it is nullable or not. This may take a very long time to achieve. And may even be unsatisfactory: for some situations a client is guaranteed that the value will never be `null`, but as there are other situations where it can be `null` the variable or method has to be marked as nullable.

So the perfect state where you know exactly what is nullable will likely never be achieved. Even so, even adding nullable information to the most-used types of Wicket will greatly improve the developer experience. Yes, it is too bad that nobody took the time to add the annotations to less-used parts. And yes, it is unfortunate that a value is marked as nullable that will never be `null` because of the way you set things up.

Both issues can be solved: more effort by people who want to contribute, and new, more strict APIs and backward-incompatible changes to contracts of existing APIs (if acceptable for Wicket users).

In my opinion adding nullability information to even a small number of places, makes Wicket easier to use. This will hopefully help convince developers that Wicket still is a good choice for front end development.

Note that knowing what is nullable is one of the reason I love working with Kotlin: no more going through incomplete documentation, and then through multiple layers of code to figure out whether or not I should handle `null`s. It saves me so much time. This time saving can also be (partially) realized for Wicket developers.

keesvandieren:
For seamless Kotlin integration, having proper @Nullable and @NotNull annotations is extremely helpful.

We're using Kotlin with Wicket, and one challenge we've encountered is the need to manually remove nullability when overriding methods. This adds unnecessary friction and makes the combination of Kotlin and Wicket feel less fluent. Improving support for nullability annotations would greatly enhance the developer experience.

theigl:
I thought about this some more over the weekend: It took me about 30 hours to fully annotate our ~350KLOC codebase by using the approach described above. Wicket is much smaller than that and we probably only need to annotate the most important packages in `wicket-core` and `wicket-request` to get immediate benefits for Kotlin users.

I would suggest the following approach:

1. Create a PR that adds JSpecify but does not annotate anything and merge it
2. Create PRs that annotate a single package or a small group of packages with `@NullMarked` and `@Nullable`

That way individual PRs should be relatively small and easy to review and by using only `@Nullable` the noise should be kept to a minimum.

Ideally, these changes would be driven by someone who actually uses Wicket with Kotlin and can verify the improvements.

PR #1102, opened 2025-02-13
WICKET-7149: Add Integrity and CrossOrigin values to ResourceReference and related code

mpritt:
**Synopsis**  

Added two new variables to the ResourceReference class.  Added code to use these values in both the JavaScriptReferenceHeaderItem and CssReferenceHeaderItem classes.  Also fixed a bug with the AbstractCssReferenceHeaderItem class where it was directly accessing the identity and crossOrigin values directly instead of going through accessor methods.

**In-depth Detail**:

It is easy to create the integrity and cross origin values needed when the we have the control over the creation of the JavaScriptHeaderItem or CssReferenceHeaderItem.  It becomes much more difficult when there are resources that are created and then stored in wicket framework classes (whether resource references are framework or user created), because these resources are used later by the framework when it does the actual creation of the header items (i.e. like JavaScriptHeaderItem.forReference()).

Supporting scenario...try setting the integrity hash for the  "wicket-ajax-jquery.js" file which is referenced from the wicket-core-10.1.0.jar.  This resource is created initially through the WicketAjaxJQueryResourceReference class and used in the JavaScriptLibrarySettings class.    The call getWicketAjaxReference() inside the JavaScriptLibrarySettings is used by the framework in various places, and those places create the actual JavaScriptReferenceHeaderItem  (see OnDomReadyHeaderItem.getDependencies() as an example).   While it is possible to extend a few framework classes to overwrite the default behavior and create the JavaScriptReferenceHeaderItem directly and add the desired hash and cross-origin values, that method is not a desirable or maintainable.

By adding integrity and cross-origin variables to the definition of a ResourceReference class, those values can then be used in the rendering of the JavaScriptReferenceHeaderItem through overriding the getIntegrity() and getCrossOrigin() functions.  If the header item's integrity value was not defined (i.e. null) then use the value defined from the ResourceReference (the same goes for the cross origin value).

It was also noticed that the AbstractCssReferenceHeaderItem's internalRenderCSSReference() function was not correctly referencing the integrity and cross origin values by directly accessing the values rather than using the accessor methods.  Changes were made to this class to correctly use the accessor methods.

### DIFF
--- a/wicket-core-tests/src/test/java/org/apache/wicket/markup/html/JsCssReferenceHeaderItemTest.java
+++ b/wicket-core-tests/src/test/java/org/apache/wicket/markup/html/JsCssReferenceHeaderItemTest.java
@@ -1,0 +1,261 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.markup.html;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.IOException;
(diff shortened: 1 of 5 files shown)


martin-g (approved):
Looks good to me!
It would be nice to have some unit tests though!

bitstorm (approved):

mpritt:
Added unit tests for related changes.

From: Andrea Del Bene ***@***.***>
Sent: Monday, February 17, 2025 9:42 AM
To: apache/wicket ***@***.***>
Cc: Michael Pritt ***@***.***>; Author ***@***.***>
Subject: [EXTERNAL] Re: [apache/wicket] Adding Integrity and CrossOrigin values to ResourceReference and related code (PR #1102)

BE CAREFUL:
This email originated from outside the organization. Do not click links, open attachments, or enter user credentials, unless you recognize the sender and know the content is safe.
BE CAREFUL:


@bitstorm approved this pull request.

martin-g:
Thank you, @mpritt !

PR #1133, opened 2025-03-26
WICKET-7152: HTTP-StatusCodes 307 and 308 now supported by RedirectRequestHandler

hosea:
https://issues.apache.org/jira/browse/WICKET-7152

### DIFF
--- a/wicket-request/pom.xml
+++ b/wicket-request/pom.xml
@@ -38,5 +38,11 @@
       <groupId>org.apache.wicket</groupId>
       <artifactId>wicket-util</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


wicket-request/pom.xml
```diff
@@ -38,5 +38,11 @@
       <groupId>org.apache.wicket</groupId>
       <artifactId>wicket-util</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <scope>test</scope>
```
martin-g, line 45: The `scope` should be defined in the `dependencyManagement` section in the parent pom.

wicket-request/src/main/java/org/apache/wicket/request/http/handler/RedirectRequestHandler.java
```diff
@@ -142,7 +148,7 @@ else if (mode == Mode.STATUS)
 			setStatus(response, requestCycle, location);
 		}
 		// Mode.AUTO
-		else if (status == HttpServletResponse.SC_MOVED_TEMPORARILY)
+		else if (status == HttpServletResponse.SC_FOUND)
```
martin-g, line 151: Why is this being changed ?

wicket-request/pom.xml
hosea: Resolved.  Removed scope (inherit scope from dependencyManagement of parent-pom)

bitstorm (approved):

PR #1146, opened 2025-04-12
[WICKET-7153] 1) make system mapper more reusable by making newResour…

reiern70:
1) make system mapper more reusable by making newResourceReferenceMapper protected and 
2) roll a hash code version of ResourceReferenceMapper that encodes class names by using their hash code.]

See

https://issues.apache.org/jira/browse/WICKET-7153

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/core/request/mapper/HashBasedResourceReferenceMapper.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/request/mapper/HashBasedResourceReferenceMapper.java
@@ -1,0 +1,182 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.core.request.mapper;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+import org.apache.wicket.WicketRuntimeException;
+import org.apache.wicket.request.mapper.ParentPathReferenceRewriter;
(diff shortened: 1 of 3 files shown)


wicket-core/src/main/java/org/apache/wicket/core/request/mapper/HashBasedResourceReferenceMapper.java
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * <p>
+ * Resource reference {@link org.apache.wicket.request.IRequestMapper} that encodes class names as hash codes.
+ * This allows hiding the class name from resource references. i.e., instead of <em>/wicket/resource/org.xxx.yyy.ZPanel/a.js</em> the
+ * URL will display  <em>/wicket/resource/ddd/a.js</em>, where ddd = hash(org.xxx.yyy.ZPanel). This allows globally hiding
+ * class structure of your application (not displaying it via URLs).
+ * </p>
+ *
+ * <p>
+ *     Caveat: we don't take into account has collisions. I.e. to to different class names having the same hash code.
```
martin-g, line 36:
```suggestion
 *     Caveat: we don't take into account hash collisions. I.e. to to different class names having the same hash code.
```

martin-g: `I.e. to to different class` sounds wrong. Please re-phrase it.

wicket-core/src/main/java/org/apache/wicket/core/request/mapper/HashBasedResourceReferenceMapper.java
```diff
@@ -0,0 +1,121 @@
...
+ *            public String getNamespace() {
+ * 				return "xxx";
+ *            }
+ *        };
+ *    }
+ *    </code>
+ *    on your Application class.
+ * </p>
+ */
+public class HashBasedResourceReferenceMapper extends ParentPathReferenceRewriter
```
martin-g, line 52: This class needs unit tests!

wicket-core/src/main/java/org/apache/wicket/core/request/mapper/HashBasedResourceReferenceMapper.java
```diff
@@ -0,0 +1,121 @@
...
+			{
+				return super.getPageClass(name);
+			}
+		}
+
+		@Override
+		protected String getClassName(Class<?> scope)
+		{
+			String name = super.getClassName(scope);
+			long hash = scope.hashCode();
```
martin-g, line 97: The class javadoc says `that encodes class names as hash codes`, i.e. `String.hashCode()`, but it is actualy `Class.hashCode()`

Probably you want:
```suggestion
			long hash = name.hashCode();
```
?!

reiern70:
@martin-g Thanks a lot for review and comments. I have addressed them

- I have added a test
- I re-worked the code so that users can use a different hash "generator".

martin-g (approved):

PR #1160, opened 2025-05-11
[WICKET-7154] use getQueryParameters to avoid tomcat parsing the mult…

reiern70:
Use getQueryParameters to avoid tomcat parsing the multipart request before we create MultipartServletWebRequest.

In tomcat 11 any call to getPostParameters will cause the parts to be parsed and therefore our custom logic to read parameters and files via fileupload2 will fail. It is not clear to me if this is a tomcat defect... But changes above are needed in able to have file upload to a resource working in our app (plus modification to our app to avoid any call triggering POST parsing by tomcat).

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/resource/AbstractFileUploadResource.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/resource/AbstractFileUploadResource.java
@@ -94,7 +94,7 @@ public abstract class AbstractFileUploadResource extends AbstractResource
 		final ServletWebRequest webRequest = (ServletWebRequest) attributes.getRequest();
 
 		// get the ID of the upload field (it should be unique per application)
-		String uploadId = webRequest.getRequestParameters().getParameterValue(UPLOAD_ID).toString("resource");
+		String uploadId = getUploadId(webRequest);
 
 		Bytes maxSize = getMaxSize(webRequest);
 		Bytes fileMaxSize = getFileMaxSize(webRequest);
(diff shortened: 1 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/markup/html/form/upload/resource/AbstractFileUploadResource.java
```diff
@@ -269,18 +270,31 @@ private Bytes getMaxSize(ServletWebRequest webRequest)
...
+		long fileMaxSize = webRequest.getQueryParameters().getParameterValue("fileMaxSize").toLong(-1);
 		return fileMaxSize > 0 ? Bytes.bytes(fileMaxSize) : null;
 	}
 
+	/**
+	 * Return the unique ID identifying the upload.
+	 *
+	 * @return String
+	 */
+	private String getUploadId(ServletWebRequest webRequest)
```
martin-g, line 283: Is this method really needed ? It is called just once and its body is a single line ?

martin-g: IMO it would be better to have a method that returns `webRequest.getQueryParameters()` with a Javadoc explaining why reading POST parameters should be avoided.

martin-g (approved):
IMO this change could be done in `master`, i.e. 10.x. There is no real need to create `wicket-11.x` branch yet. If you agree then please delete it!

reiern70:
> IMO this change could be done in `master`, i.e. 10.x. There is no real need to create `wicket-11.x` branch yet. If you agree then please delete it!

Yes. This is correct. Initially I thought the problem was the we are using fileupload2 for servelt5 specification and that servelet 6 version was needed. 

image

But this was not correct. Using 6 does not solves the problem. After that I tried to use tomcat own upload machinery,

https://github.com/reiern70/file-upload-broken/blob/master/src/main/java/com/mycompany/TomcatMultipartServletWebRequestImpl.java

This didn't work either... Thus, I realized the problem was some change on behavior in tomcat 11. Another side effect is wicket filter becomes not so nice as you can't use @UploadConfig on filters (and neither use API/code to specify upload properties on a filter).

Regarding using tomcat file upload machinery

https://github.com/reiern70/file-upload-broken/blob/master/src/main/java/com/mycompany/TomcatMultipartServletWebRequestImpl.java#L138

it allows to set a "progress" listener. Thus, you can till support upload progress.

PR #1167, opened 2025-05-18
[WICKET-7154] provide a way to hook into tomcat native multipart proc…

reiern70:
Provide a way to hook into tomcat native multipart processing and at the same time do upload progress reporting. This is needed because with tomcat 11.x tomcat will parse multipart whenever getParameters is called and logic using fileupload2 is very error-prone (it can be rather easily broken on applications if "someone" calls getParameters before wicket form processing takes place).

This should be working for Tomcat 11.x after 

https://github.com/apache/tomcat/pull/856

See

https://bz.apache.org/bugzilla/show_bug.cgi?id=69690


reiern70:
Closing PR but keeping branch

PR #1186, opened 2025-06-13
WICKET-7159 - Remove isEnabledInHierarchy condition from validate Form

mira-silhavy:
https://issues.apache.org/jira/browse/WICKET-7159

Reverting the breaking change introduced in https://github.com/apache/wicket/pull/1033#discussion_r2104060257

After upgrading to Wicket 10.x, form validators (IFormValidator) are not executed if any dependent form component is disabled, due to the addition of an isEnabledInHierarchy() check. This breaks backward compatibility for applications that rely on validators with conditionally enabled/disabled fields, as the validator is now skipped in these cases. This impacts forms with cross-field validation logic and causes migration issues for existing applications. A flag or configuration to restore the previous behavior, or reverting the change, is needed for backward compatibility

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
@@ -1994,9 +1994,9 @@ public class Form<T> extends WebMarkupContainer
 				// no enabled dependents, so no need to validate
 				if (log.isWarnEnabled())
 				{
-                    log.warn("IFormValidator in form `" +
-                        getPageRelativePath() +
-                        "` depends on components, but none have been enabled.");
+					log.warn("IFormValidator in form `" +
+						getPageRelativePath() +
+						"` depends on components, but none have been enabled.");
 				}
 				return;
 			}
@@ -2015,10 +2015,10 @@ public class Form<T> extends WebMarkupContainer
 				{
 					if (log.isWarnEnabled())
 					{
-                        log.warn("IFormValidator in form `" +
-                            getPageRelativePath() +
-                            "` depends on a component that has been removed from the page or is no longer visible. " +
-                            "Offending component id `" + dependent.getId() + "`.");
+						log.warn("IFormValidator in form `" +
+							getPageRelativePath() +
+							"` depends on a component that has been removed from the page or is no longer visible. " +
+							"Offending component id `" + dependent.getId() + "`.");
 					}
 					validate = false;
 					break;


wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
```diff
@@ -1986,7 +1986,7 @@ protected final void validateFormValidator(final IFormValidator validator)
 				}
 				// check if the dependent component is visible and is attached to
 				// the page
-				else if (!dependent.isVisibleInHierarchy() || !dependent.isEnabledInHierarchy() || !dependent.isFormParticipant())
```
papegaaij, line 1989: The problem I see with this approach is that the validator will also perform validation if all components are disabled. I don't think this is desired. Validation should only be performed if at least one component is enabled.

papegaaij: I think this change looks good now. The warn below does need a small change, and I think it would be a good idea to also log a warning where you put the comment `// no enabled dependents, so no need to validate`. After all, the user did add this validator to the form, so he/she probably expected it to do something.

papegaaij: Thanks for the updates. This looks fine now. @pedrosans and @martin-g what do you think about this change? It's a minor change in behavior compared to previous 10.x versions, but IMHO this appraoch makes much more sense.

martin-g (approved):

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
mira-silhavy: Hi @papegaaij, @martin-g I've pushed a fix to an older test. So that it does not return null from getDependentFormComponents.

papegaaij: Yes, agreed. If the validator does not depend on any components, then it should always be run. And this should work regardless of the returned value being null or empty.

papegaaij: From what I can see in the code, the default is to validate: https://github.com/apache/wicket/blob/wicket-9.x/wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java#L2083 . This is the only place where getDependentFormComponents is used (besides in validators themselves).

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
```diff
@@ -1986,24 +2009,26 @@ protected final void validateFormValidator(final IFormValidator validator)
...
 
-		if (validate)
-		{
-			validator.validate(this);
+			if (validate)
+			{
+				validator.validate(this);
+			}
+		} else
+			validator.validate(this);{
```
papegaaij, line 2031: This doesn't compile. But I think it would be much cleaner to only have one place where `validator.validate(this)` is called. You could move the `if (validate)` out the enclosing `if (dependents != null)` and set `validate` to true when `dependents == null`

mira-silhavy: Thanks @papegaaij, this has been changed in the last commit.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
mira-silhavy: Javadoc has been updated in the last commit.

mira-silhavy: I couldn't find the guidelines. Finally found it in wicket-core module. Updated the condition and reformatted the new code.

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
```diff
@@ -1986,14 +2011,14 @@ protected final void validateFormValidator(final IFormValidator validator)
...
 				// the page
-				else if (!dependent.isVisibleInHierarchy() || !dependent.isEnabledInHierarchy() || !dependent.isFormParticipant())
+				else if (!dependent.isVisibleInHierarchy() || !dependent.isFormParticipant())
 				{
 					if (log.isWarnEnabled())
 					{
-						log.warn("IFormValidator in form `" +
-							getPageRelativePath() +
-							"` depends on a component that has been removed from the page or is no longer visible/enabled. " +
-							"Offending component id `" + dependent.getId() + "`.");
```
papegaaij, line 1996: You seem to have replaced tabs with spaces here.

papegaaij:
This pull request looks good to me. When the last comment is addressed, I would like to merge it.

papegaaij (approved):

wicket-core/src/main/java/org/apache/wicket/markup/html/form/Form.java
mira-silhavy: Fixed

papegaaij:
Thanks @mira-silhavy for the work

andruhon:
Thank you everyone, this issue was causing us pains! Where can I learn about release plans?

PR #1385, opened 2026-02-28
WICKET-7177: alternative replacement method support

jstuyts:
One implementation supporting the replacement of XML is provided.


jstuyts:
@kbachl @martin-g what can I do to get this approved?

PR #1440, opened 2026-04-30
Add opt-in CSP embedding and plugin restrictions

jmestwa-coder:
## Summary

Add an opt-in method to apply stricter CSP directives for embedding and plugin restrictions.

## Changes

* Introduce `enableStrictEmbeddingProtection()` in `CSPHeaderConfiguration`
* Adds:

  * `frame-ancestors 'self'`
  * `object-src 'none'`

## Why

* `frame-ancestors 'self'` limits embedding to the same origin (clickjacking protection)
* `object-src 'none'` disables plugin-based content
* Keeping this opt-in avoids breaking existing applications

## Impact

* No default behavior changes
* Fully backward compatible
* Provides a simple way to enable stricter CSP when needed


martin-g (approved):

jmestwa-coder:
@martin-g any updates on this PR .?

PR #1439, opened 2026-04-30
Use SecureRandom for cryptographic operations and remove SHA1PRNG dependency

jmestwa-coder:
## Summary

Replace `java.util.Random` with `SecureRandom` in cryptographic usage and remove hardcoded `"SHA1PRNG"`.

## Changes

* Use `SecureRandom` for salt generation in `SunJceCrypt`
* Replace `"SHA1PRNG"` with `new SecureRandom()` in `DefaultSecureRandomSupplier`

## Why

* `Random` is predictable and not suitable for cryptographic use
* Salt generation requires high-entropy randomness
* Avoid hardcoded PRNG and let JVM choose the appropriate provider

## Impact

* No API or behavior changes
* Backward compatible
* Improves randomness quality in security-sensitive paths

PR #1448, opened 2026-05-06
Fix NoSuchAlgorithmException SHA1PRNG on fips enabled system.

mortenlud:
When using wicket in fips enabled systems where SHA1PRNG is removed, Wickert has no strong security fallback, and then refuses to start.


martin-g:
Please see https://github.com/apache/wicket/pull/1361 - you could configure a custom one in Wicket 10.9.0.

Also duplicates https://github.com/apache/wicket/pull/1439

PR #1451, opened 2026-05-08
Wicket 8.x security fixes

pedrosans:


pedrosans:
@bitstorm , I made one more change in the branch (committed directly) to take in consideration differences in Link + ExternalLink from their implementation in Wicket 9 forwards + fixed a few compiling errors in the tests

PR #1477, opened 2026-06-04
Fix for WICKET-7182

jackeri:
Wrapped the fetched element with jQuery to use the event call, like with other calls on the script.

### DIFF
--- a/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/wicket-autocomplete.js
+++ b/wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/wicket-autocomplete.js
@@ -665,7 +665,7 @@
 
 					if (document.activeElement !== input) {
 						ignoreOneFocusGain = true;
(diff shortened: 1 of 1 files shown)


wicket-extensions/src/main/java/org/apache/wicket/extensions/ajax/markup/html/autocomplete/wicket-autocomplete.js
```diff
@@ -665,7 +665,7 @@
 
 					if (document.activeElement !== input) {
 						ignoreOneFocusGain = true;
-						input.trigger('focus');
+						jQuery(input).trigger('focus');
```
martin-g, line 668: Better use https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus
All major browsers support it.

jackeri: True, but then it would be better to upgrade the whole style of the script. This keeps the same style as with the other calls.

PR #1485, opened 2026-06-11
WICKET-7186: correctly delegate calls to wicket cleaner

lalmeras:
FileCleaningTrackerAdapter does not delegate some calls to IFileCleaner.

These call are then performed by FileCleaningTracker superclass. Each FileCleaningTrackerAdapter then create a Reaper thread that is leaked (FileCleaningTrackerAdapter is request-scope).

Intended behavior is to delegate to IFileCleaner that is a singleton (there is at most 1 Reaper by wicket application).

The issue appears on wicket 10.9.x as commons-fileupload2-M5 refactors DiskFileItem and now calls the Path variant of track(...) method.

This fix adds the missing delegate methods.

---

I successfully tested this fix on wicket-10.x branch with one of our application and a breakpoint on `FileCleaningTracker$Reaper`. Before fix, one Reaper is created for each file upload. After fix, Reaper is created only on the first call.

---

https://issues.apache.org/jira/browse/WICKET-7186

### DIFF
--- a/wicket-util/src/main/java/org/apache/wicket/util/file/FileCleaner.java
+++ b/wicket-util/src/main/java/org/apache/wicket/util/file/FileCleaner.java
@@ -17,6 +17,7 @@
 package org.apache.wicket.util.file;
 
 import java.io.File;
+import java.nio.file.Path;
 
 import org.apache.commons.io.FileCleaningTracker;
 import org.apache.commons.io.FileDeleteStrategy;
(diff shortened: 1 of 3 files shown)


bitstorm (changes requested):
Thank you @lalmeras ! You patch looks good. could you just modify the indentation style for consistency (i.e. put the opening braces on a new line )?

lalmeras:
Thank you for your feedback.

I just amend my commit to fix the indentation style.

bitstorm (approved):

Issue #1492, opened 2026-06-19
AjaxRequestHandler generation of PartialPageUpdate is not extensible


reiern70:

reiern70:
Fiexed

PR #1493, opened 2026-06-19
[https://github.com/apache/wicket/issues/1492] make partial page upda…

reiern70:
make partial page update creation pluggable.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/ajax/AjaxRequestHandler.java
+++ b/wicket-core/src/main/java/org/apache/wicket/ajax/AjaxRequestHandler.java
@@ -80,13 +80,13 @@ public class AjaxRequestHandler extends AbstractPartialPageRequestHandler implem
 	/**
 	 * Collector of page updates.
 	 */
-	private final PartialPageUpdate update;
+	private PartialPageUpdate update;
 
 	/** a set of listeners */
-	private Set<AjaxRequestTarget.IListener> listeners = null;
+	protected Set<AjaxRequestTarget.IListener> listeners = null;
 
 	/** */
-	private final Set<ITargetRespondListener> respondListeners = new HashSet<>();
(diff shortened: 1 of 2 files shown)


wicket-core/src/main/java/org/apache/wicket/ajax/AjaxRequestHandler.java
```diff
@@ -104,7 +104,18 @@ public AjaxRequestHandler(final Page page)
 	{
 		super(page);
 
-		update = new XmlPartialPageUpdate(page)
+		update = newPartialPageUpdate(page);
```
martin-g: https://stackoverflow.com/questions/3404301/whats-wrong-with-overridable-method-calls-in-constructors

It will be better to add another constructor `public AjaxRequestHandler(final Page page, final PartialPageUpdate ppu)`

reiern70: Fine by me... checking

ernestosemedy: Factory method is better because constructor argument does not allow access to protected methods

martin-g: The problem is that the overrideable method is called on an instance with unfinished construction.

ernestosemedy: Moved creation to getter

ezvangiggv-ship-it:
> ایجاد به‌روزرسانی جزئی صفحه را قابل اتصال کنید.

dollar

ezvangiggv-ship-it:
Dollar 💰****

reiern70:
@martin-g now creating is done at the getter ...

martin-g (approved):

PR #1498, opened 2026-06-21
WICKET-7187 type for JavaScript content item, and new JavaScript import map item

jstuyts:
Added type to JavaScript content header items, and added a JavaScript import map header item.

The first attempt was to make a base class for `AbstractJavaScriptReferenceHeaderItem` and `JavaScriptContentHeaderItem`, and move the type to there. But `setType(...)` in the former returns `AbstractJavaScriptReferenceHeaderItem`. Moving the type information up the hierarchy would break backward compatibility. The (simple) code for managing the type has simply been copied now.

`JavaScriptReferenceType` should be called `JavaScriptType` now, but that would also break backward compatibility. This has been documented.

If the new features are not used, almost everything works as before. The only change to current behavior is that `equals(...)` and `hashCode(...)` of `JavaScriptContentHeaderItem` now also use the type.

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/markup/head/JavaScriptDataBlockType.java
+++ b/wicket-core/src/main/java/org/apache/wicket/markup/head/JavaScriptDataBlockType.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.wicket.markup.head;
+
+import java.util.Locale;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.apache.wicket.util.lang.Args;
+
+/**
+ * A <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type#value">data block</a>
+ * type (that is, any other value than a browser-processed value) for {@link JavaScriptContentHeaderItem}.
+ */
+public class JavaScriptDataBlockType implements JavaScriptContentType
+{
+    private static final Pattern TYPE_PREFIX_PATTERN = Pattern.compile("^[-!#$%&'*+.0-9A-Z^_`a-z{|}~]+/");
+    private static final Set<String> JAVASCRIPT_MEDIA_TYPES = Set.of(
+            "text/javascript",
+            "application/javascript",
+            "application/ecmascript",
+            "application/x-ecmascript",
+            "application/x-javascript",
+            "text/ecmascript",
+            "text/javascript1.0",
+            "text/javascript1.1",
+            "text/javascript1.2",
+            "text/javascript1.3",
+            "text/javascript1.4",
+            "text/javascript1.5",
+            "text/jscript",
+            "text/livescript",
+            "text/x-ecmascript",
+            "text/x-javascript"
+    );
+
+    private final String type;
+
+    /**
+     * Create a new data block type with the given media type.
+     * <p>
+     * A rudimentary check is done to ensure <code>type</code> is a media type. It must start with a
+     * <a href="https://www.rfc-editor.org/info/rfc9110/#media.type">token followed by a slash</a>. No check is done for
+     * what follows the slash.
+     * <p>
+     * The type may also not be one of the JavaScript media types defined in
+     * <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types#textjavascript"><i>text/javacript</i>
+     * on MDN</a>.
+     *
+     * @param type the media type of the data block.
+     */
+    public JavaScriptDataBlockType(String type)
+    {
+        Args.notNull(type, "type");
+        Args.isTrue(TYPE_PREFIX_PATTERN.matcher(type).find(),
+                "'type' must be a media type (that is: start with a type followed by a slash).");
+        Args.isFalse(JAVASCRIPT_MEDIA_TYPES.contains(type.toLowerCase(Locale.ENGLISH)),
+                "'type' may not be a JavaScript media type.");
+
+        this.type = type;
+    }
+
+    @Override
+    public String getType()
+    {
+        return type;
+    }
+}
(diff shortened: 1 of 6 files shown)


bitstorm (commented):
Hi @jstuyts ,

I also think moving 'type' up int he hierarchy is a good idea, but I would prefer JavaScriptHeaderItem instead of JavaScriptContentHeaderItem as target class. We can do it on master branch where backward compatibility is not required.
For Wicket 10.x branch you solution looks fine, but I wouldn't change  JavaScriptContentHeaderItem adding 'type' field. I'd rather keep this class untouched  for Wicket 10.x

jstuyts:
@bitstorm I've had to do some research to understand what is possible with `<script>`. As usual with HTML, it is quite complex. References can only be classic scripts or modules, while inline content can be classic scripts, modules, import maps, speculation rules and data blocks: https://html.spec.whatwg.org/multipage/scripting.html#attr-script-type

So I think the current implementation of reference header item classes with `JavaScriptReferenceType`, correctly reflects what is allowed.

What I think is needed are:

* The addition of a attribute `type` to `JavaScriptContentHeaderItem`, with its own type hierarchy instead of reusing `JavaScriptReferenceType`.
* Specialized classes for import maps and speculation rules, to provide a nice API to construct these instead of having to generate JSON yourself with `JavaScriptContentHeaderItem`. Because of the flexibility of import maps and (probably) speculation rules, these specialized classes may have limitations.

With these changes Wicket will prevent the user from making (some) mistakes, while still allowing full control if needed.

It will look like this. Green types already exists, purple ones are new:
afbeelding

Edit: including PlantUML source for accessibility:

```plantuml
@startuml

class JavaScriptHeaderItem

class JavaScriptContentHeaderItem {
    type
}
JavaScriptHeaderItem <|-- JavaScriptContentHeaderItem

note left of JavaScriptContentHeaderItem
    Gives full control for any inline
    `<script` element, but specialized
    classes for import maps and
    speculation rules are easier to
    work with.
end note

class AbstractJavaScriptReferenceHeaderItem
JavaScriptHeaderItem <|-- AbstractJavaScriptReferenceHeaderItem

class JavaScriptReferenceHeaderItem
AbstractJavaScriptReferenceHeaderItem <|-- JavaScriptReferenceHeaderItem

class JavaScriptUrlReferenceHeaderItem
AbstractJavaScriptReferenceHeaderItem <|-- JavaScriptUrlReferenceHeaderItem

class JavaScriptImportMapHeaderItem << (C,BB33FF) >>
JavaScriptHeaderItem <|-- JavaScriptImportMapHeaderItem

note bottom of JavaScriptImportMapHeaderItem
    Nicer to work with than with
    `JavaScriptContentHeaderItem`
    for import maps. But may have
    limitations.
end note

class JavaScriptSpeculationRulesHeaderItem << (C,BB33FF) >>
JavaScriptHeaderItem <|-- JavaScriptSpeculationRulesHeaderItem

note bottom of JavaScriptSpeculationRulesHeaderItem
    Nicer to work with than with
    `JavaScriptContentHeaderItem`
    for speculation rules. But may
    have limitations.
end note

class JavaScriptReferenceType
AbstractJavaScriptReferenceHeaderItem --> JavaScriptReferenceType

interface JavaScriptContentType << (C,BB33FF) >>
JavaScriptContentHeaderItem --> JavaScriptContentType

enum JavaScriptWellKnownContentType << (E,BB33FF) >> {
    IMPORT_MAP
    JAVASCRIPT
    MODULE
    SPECULATION_RULES
}
JavaScriptContentType <|-- JavaScriptWellKnownContentType

class JavaScriptDataBlockType << (C,BB33FF) >> {
    mimeType
}
JavaScriptContentType <|-- JavaScriptDataBlockType

note bottom of JavaScriptDataBlockType
    Checks `mimeType` is
    not a well-known type.
end note

@enduml
```

bitstorm:
@jstuyts hi see the point. The PR looks good to me but as I mentioned in the previous comment I would apply this "as is" to the wicket-10.x branch, while I think we might rework it for branch master where compatibility constraints don't apply. What do you think if we use class JavaScriptHeaderItem as target base class for  type field?

jstuyts:
@bitstorm in my opinion, moving `type` to `JavaScriptHeaderItem` is the wrong approach. There are certain types (currently import maps and speculation rules) that can really benefit from a dedicated API. These sub types of `JavaScriptHeaderItem` would also inherit `type`, but ignore its value because they already know their type.

With the type hierarchy in the schema above, the API is nicer: the user is prevented from making mistakes. Also, there is no need for breaking compatibility.

The only "downside" is that 2 enum values of `JavaScriptReferenceType` (`TEXT_JAVASCRIPT` and `MODULE`) have to be duplicated in `JavaScriptWellKnownContentType`. I feel that the improvement of the API is worth it.

What I can do is: make a PR for everything in the hierarchy above, except for `JavaScriptImportMapHeaderItem` and `JavaScriptSpeculationRulesHeaderItem` (which I was not going to build as I have found no use for it yet). I can create a separate PR for the import map later. This way, the type for content items will be available sooner.

jstuyts:
Split the changes into multiple parts:

* Adding a type to `JavaScriptContentHeaderItem`. (This PR.)
* A dedicated API for import maps. (A future PR. This could take a while.)
* A dedicated API for speculation rules. (I have no need for this now, so will not submit a PR.)

wicket-core/src/main/java/org/apache/wicket/markup/head/JavaScriptDataBlockType.java
```diff
@@ -0,0 +1,83 @@
...
+package org.apache.wicket.markup.head;
+
+import java.util.Locale;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.apache.wicket.util.lang.Args;
+
+/**
+ * A <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type#value">data block</a>
```
bitstorm, line 26: Is this the correct URL or shall we use 'https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types#textjavascript' also here?

bitstorm:
@jstuyts the new PR looks good to me! There are just a couple of minor issues to address. 

1.  Check the URL I've left a note about above
2.  Check braces indentation to conform to the code standard (open braces must go on a new line)

Thank you!

jstuyts:
@bitstorm I do not understand which URL you mean. I checked here and Jira for it.

I certainly can create a PR for the 10.x branch too.

bitstorm:
Hi @jstuyts 

> @bitstorm I do not understand which URL you mean. I checked here and Jira for it.
> 

I mean the URL [here](https://github.com/apache/wicket/pull/1498/changes/BASE..cba7fc34c3a41a1a3f2160721de0e4e0baf98621#r3528201880)  
Isn't more appropriate [this link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types#textjavascript)  to MIME types?

jstuyts:
@bitstorm sorry, I rebased and force pushed for the latest version of this PR. That must have broken the link. Which class and method/constructor is this about? Thanks.

bitstorm:
@jstuyts dont't worry, I also wasn't so clear...I mean the url in the javadoc at class level in JavaScriptDataBlockType

jstuyts:
@bitstorm the URL in the documentation is correct. It discusses the allowed values of `type`, and explains when `<script>` is considered a data block.

I did extend and simplify the (class and constructor) documentation a bit (and fixed a typo).

The commits have been squashed and force-pushed again.

bitstorm (approved):

bitstorm (approved):

bitstorm:
@jstuyts introducing JavaScriptContentType is a nice idea.  On master branch it would be nice if:

- JavaScriptReferenceType would also implement it.
- AbstractJavaScriptReferenceHeaderItem#type could be changed to JavaScriptReferenceType

PR #1504, opened 2026-07-04
WICKET-7183 fix popupsettings#setTarget regression

pedrosans:
target is now processed as the parameter described in the javadoc (using single quotes)

### DIFF
--- a/wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
+++ b/wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
@@ -96,6 +96,33 @@ public class JavaScriptUtils
 	}
 
 	/**
+	 * Removes the surrounding single quotes from a string after trimming
+	 * leading and trailing whitespace.
+	 *
+	 * <p>Examples:</p>
+	 * <pre>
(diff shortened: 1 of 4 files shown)


wicket-core/src/main/java/org/apache/wicket/core/util/string/JavaScriptUtils.java
```diff
@@ -95,6 +95,29 @@ public static CharSequence escapeQuotesAndBackslash(final CharSequence input)
...
+	 * unquoteSingleQuoted("'url'")     -> "url"
+	 * unquoteSingleQuoted("''")        -> ""
+	 * </pre>
+	 *
+	 * @param s the single-quoted string
+	 * @return the contents between the outer single quotes
+	 */
+	public static String unquoteSingleQuoted(String s)
+	{
+		if (s != null && s.trim().startsWith("'") && s.trim().endsWith("'"))
```
solomax: `trim()` is used 3 times, maybe it's better to introduce variable for trimmed string?

pedrosans: Hi, it's. I changed the code to use an auxiliary variable and to work like neighboring methods (assigning the return value)

solomax (approved):

PR #1516, opened 2026-07-14
Validate file upload resource constraints server-side

M0nd0R:
## Summary
This change stops `FileUploadToResourceField` from trusting client-controlled upload limit parameters.

## Root Cause
The upload resource accepted `maxSize`, `fileMaxSize`, and `fileCountMax` directly from the request URL. A client could tamper with those values and relax the server-side limits that were rendered by the component.

## Fix
Render a signed `uploadToken` alongside the configured limits and reject uploads when the submitted limits do not match the signed values.

## Validation
- `mvn -pl wicket-core-tests -am -Dtest=org.apache.wicket.markup.html.form.upload.resource.FileUploadToResourceFieldSecurityTest -Dsurefire.failIfNoSpecifiedTests=false clean test`


reiern70:
@M0nd0R thanks for the PR... I will review ASAP and test our app, that uses this functionality, still works.

reiern70 (approved):
LGTM. Thanks!

PR #1517, opened 2026-07-14
Reject appended plaintext listener parameters

M0nd0R:
## Summary
This change rejects plaintext request-listener parameters that are appended to an otherwise encrypted bookmarkable page URL.

## Root Cause
`decryptEntireUrl(...)` decrypted the protected URL and then merged additional query parameters from the incoming request. A plaintext `PageComponentInfo` parameter could therefore survive outside the encrypted payload and still affect request mapping.

## Fix
Reject incoming query parameters that parse as `PageComponentInfo` before merging the request query parameters back into the decrypted URL.

## Validation
- `mvn -pl wicket-core-tests -am -Dtest=org.apache.wicket.core.request.mapper.CryptoMapperTest -Dsurefire.failIfNoSpecifiedTests=false clean test`


M0nd0R:
Closing.

PR #1523, opened 2026-07-22
WICKET-7192 restrict SourcesPage to resources of the page's own package

fr-manvi:
The `source` page parameter of the examples source browser goes straight into `Class.getResourceAsStream()`, so it is not limited to the package it claims to show:
- `?SourcesPage_class=org.apache.wicket.examples.helloworld.HelloWorld&source=/META-INF/NOTICE` returns any class path resource, absolute names included
- `source=../style.css` walks out of the package the same way
- `.class` files are readable too, although the file browser deliberately filters them out

The page already computes the list of files it is willing to show, so the requested name is now checked against that list. Building the list also had to stop looking at only the first class path entry of the package, otherwise a package present in more than one entry loses files it should offer.

### DIFF
--- a/wicket-examples/src/main/java/org/apache/wicket/examples/source/SourcesPage.java
+++ b/wicket-examples/src/main/java/org/apache/wicket/examples/source/SourcesPage.java
@@ -101,6 +101,12 @@ public class SourcesPage extends WebPage
 			try
 			{
 				source = (name != null) ? name : sourceParam.toString();
+				if (!packagedResources.getObject().contains(source))
+				{
+					log.error("user is trying to access resource: {} which is not part of the package of {}",
+						source, getPageTargetClass().getName());
+					return "Unable to read the source for " + source;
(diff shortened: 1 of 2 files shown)


fr-manvi:
CI is red on AsynchronousPageStoreTest.storeAsynchronousContextClosed, a flaky async timing test in wicket-core-tests. It's unrelated to this change, which only touches wicket-examples, and the reactor fails in core before it ever reaches the examples module. Would appreciate a review when you get a chance.

fr-manvi:
The red check was an unrelated flake in AsynchronousPageStoreTest (async page-store timing on another thread), not from this change; the examples tests here pass locally. I rebased onto current master to get a clean run, but it's sitting in action_required, so it needs a maintainer to approve the workflow before it can go green.

bitstorm (approved):
Looks good to me. Thank you!

PR #1577, opened 2026-09-07
WICKET-7204 Don't use INJECTION for ByteBuddy proxy creation

bitstorm:
The current proxy creation with byte buddy uses ClassLoadingStrategy.Default.INJECTION class loading strategy, but this relies on Java Unsafe support, which is deprecated and disabled with the newer release of ByteBuddy.

I'm not a bytecode expert but looks like it's the minimal change required.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -159,7 +159,7 @@
 		<aspectj.version>1.9.25.1</aspectj.version>
 		<assertj-core.version>3.27.7</assertj-core.version>
 		<bouncycastle.version>1.85.2</bouncycastle.version>
-		<byte-buddy.version>1.18.13</byte-buddy.version>
+		<byte-buddy.version>1.18.8</byte-buddy.version>
 		<cdi-unit.version>5.0.0-EA7</cdi-unit.version>
 		<commons-collections4.version>4.6.0</commons-collections4.version>
 		<commons-fileupload.version>2.0.0-M5</commons-fileupload.version>
--- a/pom.xml
+++ b/pom.xml
@@ -159,7 +159,7 @@
 		<aspectj.version>1.9.25.1</aspectj.version>
 		<assertj-core.version>3.27.7</assertj-core.version>
 		<bouncycastle.version>1.85.2</bouncycastle.version>
-		<byte-buddy.version>1.18.8</byte-buddy.version>
+		<byte-buddy.version>1.18.13</byte-buddy.version>
 		<cdi-unit.version>5.0.0-EA7</cdi-unit.version>
 		<commons-collections4.version>4.6.0</commons-collections4.version>
 		<commons-fileupload.version>2.0.0-M5</commons-fileupload.version>
--- a/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
+++ b/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
@@ -17,8 +17,10 @@
 package org.apache.wicket.proxy.bytebuddy;
 
 import java.io.Serializable;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Modifier;
 import java.util.function.Function;
 
 import org.apache.wicket.Application;
@@ -100,8 +102,10 @@ public class ByteBuddyProxyFactory implements IProxyFactory
 
 	@SuppressWarnings("unchecked")
 	public static <T> Class<T> createOrGetProxyClass(Class<T> type)
-	{
+	{		
+		ClassLoadingStrategy<ClassLoader> loadingStrategy = resolveLoadingStrategy(type);
 		ClassLoader classLoader = resolveClassLoader();
+		
 		return (Class<T>) DYNAMIC_CLASS_CACHE.findOrInsert(classLoader,
 				new TypeCache.SimpleKey(type),
 				() -> BYTE_BUDDY
@@ -116,8 +120,24 @@ public class ByteBuddyProxyFactory implements IProxyFactory
 						.implement(InterceptorMutator.class).intercept(FieldAccessor.ofBeanProperty())
 						.implement(Serializable.class, IWriteReplace.class, ILazyInitProxy.class).intercept(MethodDelegation.toField(INTERCEPTOR_FIELD_NAME))
 						.make()
-						.load(classLoader, ClassLoadingStrategy.Default.INJECTION.allowExistingTypes())
+						.load(classLoader, loadingStrategy)
 						.getLoaded());
+	}
+
+	private static ClassLoadingStrategy<ClassLoader> resolveLoadingStrategy(Class<?> type) 
+	{
+		try 
+		{
+			int modifiers = type.getModifiers();
+			
+			return !Modifier.isPublic(modifiers)
+				   ? ClassLoadingStrategy.UsingLookup.of(MethodHandles.privateLookupIn(type, MethodHandles.lookup())) 
+				   : ClassLoadingStrategy.Default.WRAPPER.allowExistingTypes();
+		} 
+		catch (IllegalAccessException e) 
+		{
+			throw new WicketRuntimeException(e);
+		}
 	}
 
 	private static ClassLoader resolveClassLoader()
@@ -205,7 +225,9 @@ public class ByteBuddyProxyFactory implements IProxyFactory
 		for (Constructor<?> constructor : type.getDeclaredConstructors())
 		{
 			if (constructor.getParameterTypes().length == 0)
+			{
 				return true;
+			}
 		}
 
 		return false;
--- a/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
+++ b/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
@@ -119,8 +119,9 @@ public class ByteBuddyProxyFactory implements IProxyFactory
 						.implement(InterceptorMutator.class).intercept(FieldAccessor.ofBeanProperty())
 						.implement(Serializable.class, IWriteReplace.class, ILazyInitProxy.class).intercept(MethodDelegation.toField(INTERCEPTOR_FIELD_NAME))
 						.make()
-						.load(classLoader, loadingStrategy)
-						.getLoaded());
+						.load(classLoader, resolveLoadingStrategy(type))
+						.getLoaded(),
+				DYNAMIC_CLASS_CACHE);
 	}
 
 	private static ClassLoadingStrategy<ClassLoader> resolveLoadingStrategy(Class<?> type) 
--- a/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
+++ b/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
@@ -102,10 +102,9 @@ public class ByteBuddyProxyFactory implements IProxyFactory
 
 	@SuppressWarnings("unchecked")
 	public static <T> Class<T> createOrGetProxyClass(Class<T> type)
-	{		
-		ClassLoadingStrategy<ClassLoader> loadingStrategy = resolveLoadingStrategy(type);
+	{
 		ClassLoader classLoader = resolveClassLoader();
-		
+
 		return (Class<T>) DYNAMIC_CLASS_CACHE.findOrInsert(classLoader,
 				new TypeCache.SimpleKey(type),
 				() -> BYTE_BUDDY
--- a/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
+++ b/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
@@ -232,9 +232,7 @@ public class ByteBuddyProxyFactory implements IProxyFactory
 		for (Constructor<?> constructor : type.getDeclaredConstructors())
 		{
 			if (constructor.getParameterTypes().length == 0)
-			{
 				return true;
-			}
 		}
 
 		return false;
--- a/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
+++ b/wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
@@ -124,19 +124,27 @@ public class ByteBuddyProxyFactory implements IProxyFactory
 				DYNAMIC_CLASS_CACHE);
 	}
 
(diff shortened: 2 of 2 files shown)


wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
```diff
@@ -116,10 +120,26 @@ public static <T> Class<T> createOrGetProxyClass(Class<T> type)
 						.implement(InterceptorMutator.class).intercept(FieldAccessor.ofBeanProperty())
 						.implement(Serializable.class, IWriteReplace.class, ILazyInitProxy.class).intercept(MethodDelegation.toField(INTERCEPTOR_FIELD_NAME))
 						.make()
-						.load(classLoader, ClassLoadingStrategy.Default.INJECTION.allowExistingTypes())
+						.load(classLoader, loadingStrategy)
 						.getLoaded());
 	}
 
+	private static ClassLoadingStrategy<ClassLoader> resolveLoadingStrategy(Class<?> type)
```
reiern70: @bitstorm javadoc?

reiern70 (approved):
Also not an expert on this. Thus a mild approval

wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
```diff
@@ -17,8 +17,10 @@
 package org.apache.wicket.proxy.bytebuddy;
 
 import java.io.Serializable;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Modifier;
```
papegaaij: `Modifier` is not needed once the branch is on the package rather than on the visibility.

```suggestion
```

wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
```diff
@@ -100,8 +102,10 @@ public <T> T createProxy(final Class<T> type, final IProxyTargetLocator locator)
 
 	@SuppressWarnings("unchecked")
 	public static <T> Class<T> createOrGetProxyClass(Class<T> type)
-	{
+	{		
+		ClassLoadingStrategy<ClassLoader> loadingStrategy = resolveLoadingStrategy(type);
 		ClassLoader classLoader = resolveClassLoader();
+
```
papegaaij: Resolving the strategy here runs `privateLookupIn` on every `createProxy` call, not just on a cache miss, and makes it throw even when the proxy class is already cached. Moving it into the lambda also lets the trailing whitespace go.

```suggestion
	{
		ClassLoader classLoader = resolveClassLoader();

```

wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
```diff
@@ -116,10 +120,26 @@ public static <T> Class<T> createOrGetProxyClass(Class<T> type)
 						.implement(InterceptorMutator.class).intercept(FieldAccessor.ofBeanProperty())
 						.implement(Serializable.class, IWriteReplace.class, ILazyInitProxy.class).intercept(MethodDelegation.toField(INTERCEPTOR_FIELD_NAME))
 						.make()
-						.load(classLoader, ClassLoadingStrategy.Default.INJECTION.allowExistingTypes())
+						.load(classLoader, loadingStrategy)
 						.getLoaded());
```
papegaaij: The monitor argument makes class creation for one `(loader, type)` pair mutually exclusive. Without it, `UsingLookup` has no `allowExistingTypes()` equivalent and concurrent first use of a package-private type defines the class twice — WICKET-7005 again.

```suggestion
						.load(classLoader, resolveLoadingStrategy(type))
						.getLoaded(),
				DYNAMIC_CLASS_CACHE);
```

wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
```diff
@@ -116,10 +120,26 @@ public static <T> Class<T> createOrGetProxyClass(Class<T> type)
...
+			
+			return Modifier.isPublic(modifiers)
+				   ? ClassLoadingStrategy.Default.WRAPPER.allowExistingTypes()
+				   : ClassLoadingStrategy.UsingLookup.of(MethodHandles.privateLookupIn(type, MethodHandles.lookup()));
+		} 
+		catch (IllegalAccessException e) 
+		{
+			throw new WicketRuntimeException(e);
+		}
+	}
```
papegaaij: This is the core of it: branch on whether the proxy keeps the type's package, not on the type's visibility. Only a `java.*` proxy is renamed into another package by `WicketNamingStrategy`, and that is the only case that needs a class loader of its own — `privateLookupIn` cannot serve it anyway, since `java.base` does not open its packages.

The message on the exception matters here: when it fires, it is because a named module does not `open` its package to wicket-ioc, and a bare `IllegalAccessException` gives the user nothing to act on.

```suggestion
	/**
	 * The proxy has to be defined in the same runtime package as the type it proxies, or the
	 * package private methods it overrides are not overridden at all. Only a proxy for a
	 * <em>java.**</em> type is renamed into another package, and needs a class loader of its own.
	 */
	private static ClassLoadingStrategy<ClassLoader> resolveLoadingStrategy(Class<?> type)
	{
		if (type.getName().startsWith("java."))
		{
			return ClassLoadingStrategy.Default.WRAPPER.allowExistingTypes();
		}

		try
		{
			return ClassLoadingStrategy.UsingLookup
				.of(MethodHandles.privateLookupIn(type, MethodHandles.lookup()));
		}
		catch (IllegalAccessException e)
		{
			throw new WicketRuntimeException("Cannot create a proxy for " + type.getName()
				+ ", because its package is not open to " + ByteBuddyProxyFactory.class.getModule(), e);
		}
	}
```

wicket-ioc/src/main/java/org/apache/wicket/proxy/bytebuddy/ByteBuddyProxyFactory.java
```diff
@@ -205,7 +225,9 @@ private static boolean hasNoArgConstructor(Class<?> type)
 		for (Constructor<?> constructor : type.getDeclaredConstructors())
 		{
 			if (constructor.getParameterTypes().length == 0)
+			{
```
papegaaij: Unrelated to the fix — worth dropping to keep the commit to one thing.

pom.xml
```diff
@@ -159,7 +159,7 @@
 		<aspectj.version>1.9.25.1</aspectj.version>
 		<assertj-core.version>3.27.7</assertj-core.version>
 		<bouncycastle.version>1.85.2</bouncycastle.version>
-		<byte-buddy.version>1.18.8</byte-buddy.version>
+		<byte-buddy.version>1.18.13</byte-buddy.version>
```
papegaaij: This bump is separable from the fix: 1.18.13 does not disable `INJECTION` (I checked on JDK 25), so it is not what forces the change. No objection to bumping, but it reads as part of the fix here.

papegaaij (changes requested):
I had a close look at this and reproduced **two regressions against `master`**, both in `wicket-ioc`. One of them is silent. The direction is right — moving off `INJECTION` is worth doing — but I think the strategy is being selected on the wrong property.

## 1. Package-private methods of a public class are silently no longer intercepted

`WRAPPER` defines the proxy in a fresh child `ClassLoader`. `WicketProxy_Foo` then shares the package *name* with `Foo` but sits in a different **runtime package**, so a package-private method is no longer an override: the call resolves to the superclass method and runs against the Objenesis-allocated proxy, whose fields are all null. No exception, just a wrong value.

That is exactly what `WicketNamingStrategy` promises in its javadoc — *"This way the generated proxy class could still access package-private members of sibling classes"* — and what WICKET-7025 added the `isPackagePrivate()` matcher for.

With a public class exposing a package-private getter:

| | `proxy.internalGetMessage()` |
| --- | --- |
| `master` | `"public-concrete"` |
| this branch | `null` |

Test at the bottom (`PublicPackagePrivateMethodTest`).

## 2. WICKET-7005 comes back for package-private types

`ClassLoadingStrategy.UsingLookup` has no `allowExistingTypes()` — only `Default` implements `Configurable` — and `TypeCache.findOrInsert` is called without a monitor. Concurrent first use of one type therefore defines the class twice:

```
LinkageError: loader 'app' attempted duplicate class definition for
org.apache.wicket.proxy.packageprivate.WicketProxy_PackagePrivateConcreteObject
```

A 16-thread stress test on `createOrGetProxyClass`: **15 of 16 threads fail on this branch, green on `master`.** `ParallelInjectionTest` does not catch it, because its beans are `public static` and take the `WRAPPER` path.

Test at the bottom (`PackagePrivateParallelTest`).

## The condition wants to be about the package, not the visibility

What decides the strategy is not whether the type is public, it is whether the proxy keeps the type's package. `WicketNamingStrategy` relocates only `java.*` types, into `bytebuddy_generated_wicket_proxy.…`. Measured, each in its own JVM:

| type | `INJECTION` (today) | `WRAPPER` | `UsingLookup` |
| --- | --- | --- | --- |
| public application class | pkg-private intercepted | **not intercepted** | intercepted |
| package-private application class | works | inaccessible | works |
| `java.util.ArrayList` | works | works | `IllegalAccessException` — `java.base` does not open `java.util` |

So `java.*` → `WRAPPER`, everything else → `UsingLookup`.

The inline suggestions carry that, plus a monitor on `findOrInsert` for #2 and the strategy resolution moved inside the cache lambda (as written it runs `privateLookupIn` on every `createProxy`, and throws even when the proxy class is already cached). **They are one coherent change — applying only some of them will not compile.**

With all of them applied: **wicket-ioc 16/16, wicket-spring 54/54, wicket-guice 7/7 green**, including `ParallelInjectionTest` and both new tests. `wicket-spring` proxies `ArrayList<String>`, so the `java.*` branch really is exercised. I did not run a full `mvn clean verify`.

One caveat I should be honest about: the monitor closes the concurrency race but not soft-reference expunction. `TypeCache.Sort.SOFT` can drop a cached class while it is still defined in the loader, and regeneration would hit the same `LinkageError` that `allowExistingTypes()` used to absorb. Matching the old robustness fully needs a `Class.forName` recovery or a strong cache.

## Smaller points

- `UsingLookup` **ignores the `ClassLoader` passed to `load()`** — I passed a `URLClassLoader` and the class still landed in the app loader. So the proxy is defined next to the type rather than into the `IClassResolver` loader, while `DYNAMIC_CLASS_CACHE` stays keyed by the resolver's loader. That divergence is inherent to the approach, and it is what makes package access work, but it is worth a thought about classloader lifetime on redeploy.
- On the module path, `privateLookupIn` needs the target package `open`ed to wicket-ioc's module, where `INJECTION` did not. Worth a line in the migration guide.
- Both commits have empty bodies, and `WICKET-7204 logic refactoring` looks like a fixup to squash.
- There is trailing whitespace on six of the new lines (two tabs after `createOrGetProxyClass`'s opening brace, two tab-only blank lines, trailing spaces after the method signature, `try` and `}`), and the ternary is indented tabs-then-spaces. The suggestions drop it.

## On the premise

I could not reproduce `INJECTION` being disabled. On JDK 25 with byte-buddy 1.18.13 it still works. It does trigger the `sun.misc.Unsafe::objectFieldOffset` terminal-deprecation warning — and still does with `--add-opens java.base/java.lang=ALL-UNNAMED`, so that warning is the Unsafe dispatcher being initialised rather than proof that the definition goes through it. Worth doing regardless, since `Unsafe` is going away, but it does not look like it is fixing a present breakage at 1.18.13.

## Tests

New files, so they cannot be inline suggestions.

<details>
<summary><code>wicket-ioc/src/test/java/org/apache/wicket/proxy/util/PublicObjectWithPackagePrivateMethod.java</code></summary>

```java
package org.apache.wicket.proxy.util;

/**
 * A public mock dependency with a package private method, to verify that such methods are
 * intercepted on the proxy as well (WICKET-7025).
 */
public class PublicObjectWithPackagePrivateMethod
{
	private String message;

	public PublicObjectWithPackagePrivateMethod()
	{
	}

	public PublicObjectWithPackagePrivateMethod(final String message)
	{
		this.message = message;
	}

	public String getMessage()
	{
		return message;
	}

	String internalGetMessage()
	{
		return message;
	}
}
```

</details>

<details>
<summary><code>wicket-ioc/src/test/java/org/apache/wicket/proxy/util/PublicPackagePrivateMethodTest.java</code></summary>

```java
package org.apache.wicket.proxy.util;

import static org.junit.jupiter.api.Assertions.assertEquals;

import org.apache.wicket.proxy.IProxyTargetLocator;
import org.apache.wicket.proxy.LazyInitProxyFactory;
import org.junit.jupiter.api.Test;

class PublicPackagePrivateMethodTest
{
	private static final PublicObjectWithPackagePrivateMethod TARGET =
		new PublicObjectWithPackagePrivateMethod("public-concrete");

	private static final IProxyTargetLocator LOCATOR = new IProxyTargetLocator()
	{

		@Override
		public Object locateProxyTarget()
		{
			return TARGET;
		}
	};

	@Test
	void packagePrivateMethodOfPublicClassIsIntercepted()
	{
		PublicObjectWithPackagePrivateMethod proxy = (PublicObjectWithPackagePrivateMethod)LazyInitProxyFactory
			.createProxy(PublicObjectWithPackagePrivateMethod.class, LOCATOR);

		assertEquals("public-concrete", proxy.getMessage());
		assertEquals("public-concrete", proxy.internalGetMessage());
	}
}
```

</details>

<details>
<summary><code>wicket-ioc/src/test/java/org/apache/wicket/proxy/packageprivate/PackagePrivateParallelTest.java</code></summary>

```java
package org.apache.wicket.proxy.packageprivate;

import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.CountDownLatch;

import org.apache.wicket.proxy.bytebuddy.ByteBuddyProxyFactory;
import org.junit.jupiter.api.Test;

class PackagePrivateParallelTest
{
	@Test
	void concurrentProxyClassCreation() throws Exception
	{
		int n = 16;
		CountDownLatch start = new CountDownLatch(1);
		List<Throwable> failures = new ArrayList<>();
		List<Thread> threads = new ArrayList<>();
		for (int i = 0; i < n; i++)
		{
			Thread t = new Thread(() -> {
				try
				{
					start.await();
					ByteBuddyProxyFactory.createOrGetProxyClass(PackagePrivateConcreteObject.class);
				}
				catch (Throwable e)
				{
					synchronized (failures)
					{
						failures.add(e);
					}
				}
			});
			t.start();
			threads.add(t);
		}
		start.countDown();
		for (Thread t : threads)
		{
			t.join();
		}
		if (!failures.isEmpty())
		{
			throw new AssertionError(failures.size() + "/" + n + " threads failed, first: "
				+ failures.get(0), failures.get(0));
		}
	}
}
```

</details>

pom.xml
bitstorm: It fails on JDK > 25 (i.e. 26) starting from version 1.18.9. See PR https://github.com/apache/wicket/pull/1574 . So yes, we better keep version bumping out of this commit but I had to do it to test the result on JDK 26.

PR #1579, opened 2026-09-07
Document the conventions an agent needs to work in this repository

papegaaij:
Nothing in the repository said how a change is made here, so a coding agent had to infer it, and the parts that matter most are the ones least visible from the source: that tests for a module usually live in a different module, that pom section and dependency order is enforced before anything compiles, that a class not named *Test is never run, and that pushing is not the agent's call.

AGENTS.md records those, along with the branch and cherry-pick order, the commit message shape, what an API change on master owes its users, and the comment, Javadoc and language rules. It deliberately does not repeat what is already readable from pom.xml, the CI workflow or SECURITY.md, because a second copy of a version number or a policy only goes stale and then misleads. CLAUDE.md just includes it, so Claude Code and any agent reading AGENTS.md see the same file.

This changes nothing about the build. A contributor who ignores both files is in exactly the position they were in before.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,135 @@
+# Apache Wicket
+
+Component-based Java web framework. `master` is the 11.x development line; the maintained
+release lines live on `wicket-10.x`, `wicket-9.x` and `wicket-8.x`. `SECURITY.md` says which of
+them still receive fixes.
+
+Versions — the Java release, the Servlet level, library and plugin versions — are not repeated
+here, because they go stale and then mislead. Read them from `pom.xml`.
+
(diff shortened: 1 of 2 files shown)


reiern70 (commented):
LGTM. We lacck the SECURITY.md

reiern70 (approved):

papegaaij:
@reiern70 Actually, we do have a SECURITY.md for some time now. It should help a bit keeping the bad security reports away.

reiern70:
> @reiern70 Actually, we do have a SECURITY.md for some time now. It should help a bit keeping the bad security reports away.

ok

PR #1580, opened 2026-09-07
Add a jQuery-free vanilla JS implementation of wicket-ajax.js

reiern70:
Wicket's client-side Ajax/event support (Wicket.Ajax, Wicket.Event, Wicket.DOM, Wicket.Form, Wicket.Head, Wicket.Focus, ...) has always required jQuery. This adds a second, plain-JavaScript implementation exposing the exact same API, so applications that don't want to ship jQuery at all can opt in via:

    getJavaScriptLibrarySettings().setWicketAjaxReference(WicketAjaxVanillaResourceReference.get());

jQuery stays the default; existing applications are unaffected.

Core changes:
- wicket-ajax-vanilla.js: full port of wicket-ajax-jquery.js onto native XMLHttpRequest/DOM/event APIs (WicketAjaxVanillaResourceReference.java is its resource reference).
- Removed jQuery from every other framework JS file that only used it incidentally (not as its actual implementation choice), so the whole stack can run without jQuery once the vanilla engine is selected: the extensions Ajax components (autocomplete, palette, upload progress bar, ajax download, trap focus), the dev debug bar, native WebSockets support, and FileUploadToResourceField.js - all verified to still work correctly under the default jQuery engine too, and their public extension points/callback signatures are unchanged.
- Cleaned up several now-dead legacy-browser fallbacks and deprecated Web APIs encountered along the way (window.escape/unescape, window.event, Attr.specified, IE-only branches).

Testing:
- Ported the wicket-core QUnit suite (channels/dom/event/form/head/ajax/timer) off jQuery-based test-authoring helpers so the same tests run against either engine; all.html now takes ?vanilla alongside the existing ?<jquery-version>, and the Gruntfile has a matching qunit:vanilla target.
- Ran the full suite for real (jQuery 3.7.1, jQuery 4.0.0, vanilla) via a portable Node build + jsdom against a real HTTP server, since no JS runtime was available here otherwise - this caught and fixed several real bugs in the vanilla engine (event "extra data" colliding with native Event.detail, beforeSend's settings.url/data not reflecting the resolved request, unsupported dataType 'json', a DOM.show() default-display mismatch, and scripts in replaced/appended markup never executing).

Examples:
- wicket-examples can now be started with either engine via a single JVM-wide switch (-Dwicket.examples.use=JQUERY|VANILLA, default JQUERY), applied through AjaxEngineSelector from every example Application's init(), with a startup console banner (mirroring Wicket's own dev-mode warning) confirming which engine is active.
- The shared example page header also shows the active engine and lets a visitor switch it for their session only, live, without an application restart - implemented via an IHeaderResponseDecorator rather than mutating the shared JavaScriptLibrarySettings, so concurrent sessions with different choices never interfere with each other. Verified end-to-end with WicketTester (both switch directions render the correct script tags in the correct order, and the toggle link updates the session as expected).

Documentation: new user guide section (ajax_11.adoc) explaining both implementations and how to switch; corrected IAjaxCallListener/ajax_5/ajax_6 docs that assumed jqXHR/jqEvent are always jQuery-specific objects.


wicket-examples/src/main/java/org/apache/wicket/examples/WicketExampleApplication.java
```diff
@@ -60,10 +62,7 @@ protected void init()
 		getDebugSettings().setDevelopmentUtilitiesEnabled(true);
 		
 		getResourceSettings().setCssCompressor(new CssUrlReplacer());
-		getCspSettings().blocking().strict().reportBack()
-		                .add(CSPDirective.STYLE_SRC,
-						"https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css")
-		                .add(CSPDirective.FONT_SRC, "https://maxcdn.bootstrapcdn.com");
+		getCspSettings().blocking().disabled();
```
reiern70: @papegaaij I would appreciate your help here: with strict the new VANILA js is no loaded. I could not figure out why

reiern70:
I will do some more testing but this seems to be working. Next steps

1. Imporve examples to have more realistic APP
2. Create selenium tests and infra to run them

papegaaij:
@reiern70 The problem was in the decorator. It did not preserve the nonce. The decorator was not needed at all. I've just pushed a fix on your branch that resolves the engine directly from the session and it works now.

reiern70:
> @reiern70 The problem was in the decorator. It did not preserve the nonce. The decorator was not needed at all. I've just pushed a fix on your branch that resolves the engine directly from the session and it works now.

Many thanks!

testing/wicket-js-tests/Gruntfile.js
```diff
@@ -20,8 +20,10 @@ module.exports = function(grunt) {
 		coreJs = [
 			'../../wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery-debug.js',
 			'../../wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-jquery.js',
+			'../../wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-vanilla.js',
```
mattrpav: Minor nit: How about simply calling it wicket-ajax.js?

mattrpav:
Love this!

testing/wicket-js-tests/Gruntfile.js
reiern70: Ok. Fine by me :-)

reiern70:
> Love this!

Thanks. It seems to be working. I need some more testing. If someone else tries it I would be happy to receive feedback.

wicket-core/src/main/java/org/apache/wicket/ajax/res/js/wicket-ajax-vanilla.js
```diff
@@ -0,0 +1,2821 @@
...
+
+	if (typeof(Wicket) === 'undefined') {
+		window.Wicket = {};
+	}
+
+	if (typeof(Wicket.Head) === 'object') {
+		return;
+	}
+
+	var isUndef = function (target) {
```
solomax: Maybe it worth to use modern `let/const` syntax? :)

reiern70: Modernized files.

mattrpav (approved):
Excited for this!

reiern70:
> Excited for this!

Thanks. I will wait for feedback of other developers too. e.g. pleople trying this branch locally. I will creat a port of this for wicket 10.x... and I will try to try that locally on our project..

@bitstorm 

Do we have some kind of staging nexus where we can publish snapshots? I would need this to run company tests against those changes for wicket 10.x. 

It would also be nice to profile these changes. Any help welcomed.

reiern70:
> Excited for this!

As I see it this could make adoption for new projects easy as you don't need to rely on jQuery. It would also be good to check wicketstuff for things that depend on wicket jQUERY Ajax implementation and roll a plain javascript version.

mattrpav:
> Do we have some kind of staging nexus where we can publish snapshots? I would need this to run company tests against those changes for wicket 10.x.
> 

apache had public facing maven snapshot repositories.

PR #1581, opened 2026-09-08
WICKET-7205: Resolve IRequestHandlerDelegate-Chain in CSP-Processing

hosea:


wicket-core-tests/src/test/java/org/apache/wicket/csp/CSPHeaderWriterTest.java
```diff
@@ -115,6 +123,53 @@ void addCspDirectiveToStatelessPageAfterNoRedirect()
...
+	void addCspDirectiveToStatelessPageEvenIfWrappedInMultipleIRequestHandlerDelegates()
+	{
+		tester.getApplication().mount(new MountedMapper("withdelegate/page", NoopMockPage.class)
+		{
+			@Override
+			public IRequestHandler mapRequest(final Request request)
+			{
+				final IRequestHandler requestHandler = super.mapRequest(request);
+				final IRequestHandler result;
+				if (requestHandler instanceof RenderPageRequestHandler renderPageRequestHandler && NoopMockPage.class.equals(renderPageRequestHandler.getPageClass()))
```
pedrosans: Hi, thank you for the PR. I think the test could be simplified by:
- just testing if requestHandler is != null. if so, it's ok to add wrap it two time.
- if null, it's ok to just return
- Page can be used instead of NoopMockPage

hosea: @pedrosans: Thank you for review. You are right. I simplified the test.

PR #1595, opened 2026-09-11
Component state rework (WICKET-6774) and further performance improvements

papegaaij:
Draft, for discussion. This collects the WICKET-6774 component state rework together with a
handful of further allocation and CPU savings found by profiling a real application, and adds
the benchmark module they were all measured with.

## wicket-benchmarks

A new module, in the default reactor so that the benchmarks keep compiling against the current
API. It is never released and has no unit tests; see its README for how to run it.

The component benchmarks used to investigate WICKET-6774 only ever existed as attachments on
the issue, and they no longer compile: `WicketTester` has moved to its own module, and since
JDK 23 javac no longer runs annotation processors found on the classpath, so JMH silently
produces no `BenchmarkList` and the run executes nothing. Keeping them in the reactor means
they keep compiling.

It now holds three kinds of tool:

* **JMH benchmarks** - `ComponentStateBenchmark`, `PageRenderBenchmark`,
  `ResourceNameIteratorBenchmark`, `PageSerializationBenchmark`, `PageEncryptionBenchmark`.
* **`ComponentFootprint`** - retained heap via JOL and serialized size, per state shape,
  against an identical stateless tree so the difference isolates the state itself.
* **`WicketContext`** - the shared harness.

## Component state (WICKET-6774)

A component's state is no longer an `Object[]` with a packing convention, but a
`ComponentState` instance. One final class rather than one subclass per combination: every
unpacking call site otherwise dispatches over four implementations of the same six accessors,
which is past the point where HotSpot stops inlining, and a real page interleaves shapes even
where a benchmark feeding one shape at a time would not show it.

Reading state over an array holding every shape at once, 2 forks, ns/op, master -> this branch:

| | master | this branch | |
|---|---|---|---|
| `readBehaviorsMixedShapes` | 31.50 | 21.35 | -32% |
| `readMetaDataMixedShapes`  | 25.27 | 17.66 | -30% |
| `readModelMixedShapes`     | 30.17 | 16.54 | -45% |
| `readBehaviors[AJAX_BEHAVIOR]` | 11.15 | 1.97 | -82%, 72 -> 24 B/op |
| `readBehaviorById` | 2.82 | 1.62 | -43% |

For a component carrying a real `AjaxEventBehavior` the tree is 39.8% smaller serialized, -72
bytes retained heap and -32 bytes serialized per component - comparable to the -36.2% reported
on the issue in 2020.

**Behaviour change:** behavior ids are only maintained for stateful behaviors. Ids can change
for other behaviors, also when combined on the same component. This is documented in the guide
commit on this branch.

## Resource name iteration

`ResourceStreamLocator` walks a list of candidate filenames for every property and markup
lookup, once per registered properties loader. Misses are the common case, because a key is
resolved by climbing the component hierarchy, so every class above the one that declares it
contributes a full traversal that finds nothing.

`toString()` built each candidate from four `prepend()` calls, and the locale part went through
`getLocale()` - a `Locale.of()` cache lookup - even though `LocaleResourceNameIterator.next()`
had just built the identical suffix and thrown it away. `ResourceUtil#rejectPathSeparators`
separately ran `Locale#toString()` on every call.

`nl_NL` without a style, 3 forks:

| | master | this branch | |
|---|---|---|---|
| `walkAllCandidates` (miss) | 923.8 ns, 1224 B | 139.9 ns, 672 B | -85% / -45% |
| `firstCandidate` (hit) | 775.2 ns, 600 B | 405.6 ns, 368 B | -48% / -39% |

Both changes were verified by enumeration rather than by argument: every candidate name with
its locale, style, variation and extension over 6 paths x 2 styles x 2 variations x 7 locales x
5 extension lists x strict/non-strict (9918 names), and 4918 locales for `rejectPathSeparators`
- byte-identical before and after.

One note for reviewers: `toString()` now takes the locale segment from `next()` instead of from
`getLocale()`. A subclass supplied through the protected `newLocaleResourceNameIterator` hook
that overrode `getLocale()` without also overriding `next()` would see a difference; nothing in
the tree does that.

## Page store

Serialization and encryption run on the request thread after the response has been flushed, so
this is throughput and thread occupancy rather than user-visible latency.

`JavaSerializer` wrote into a `java.io.ByteArrayOutputStream` with no initial size, which
starts at 32 bytes and grows by copying everything written so far into a buffer of twice the
size. Wicket already has a `ByteArrayOutputStream` that chains a new buffer instead.

`SchemeCrypt` prefixes every ciphertext with a one-byte marker naming the scheme, and did so by
allocating a new array one byte longer and copying the whole payload in - and on the way back
copying everything after the marker out again. `AbstractAesGcmCryptScheme` added a third copy.
`ICryptScheme#encrypt`/`#decrypt` now carry the offsets needed to avoid all three; the old
signatures remain as default methods. The marker stays `SchemeCrypt`'s concern, as documented
in the user guide - a scheme still knows nothing about it, only that the first few bytes of
what it returns are not its own.

| | master | this branch | |
|---|---|---|---|
| serialize, 500 components (37,924 bytes) | 256,714 B/op | 214,763 B/op | -16% |
| encrypt, 40kB | 18.28 us, 124,496 B | 12.58 us, 44,416 B | -31% / -64% |
| decrypt, 40kB | 14.07 us, 84,408 B | 11.17 us, 44,360 B | -21% / -47% |

Encrypting now allocates 1.11x the payload rather than 3.1x. Serialization time is unchanged;
what remains there is `ObjectOutputStream`'s handle table, which scales with the number of
objects in the page graph rather than with its byte count - which is the thing the component
state rework above reduces.

## How the non-6774 hotspots were found

JFR on a large production application under its Selenium suite, keeping only samples whose
stack contains `WicketFilter.doFilter`, charging JDK frames to the nearest non-JDK caller and
bucketing by package. Inside Wicket request handling: page store 23.2% CPU / 12.4% allocation,
render 16.8% / 15.5%, localizer 6.1% / 15.0%.

Two attempts were measured and dropped rather than included: folding the resource name
iterators into precomputed lists (6x slower on the hit path, because `getLocale()` costs a
cache lookup per state and the hit path stops early), and a `reset()`-based restart of the same
iterators (no time win, three new subclass contract hazards).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wicket-benchmarks/src/main/java/org/apache/wicket/benchmarks/PageSerializationBenchmark.java
```diff
@@ -18,14 +18,11 @@
 
 import java.util.concurrent.TimeUnit;
 
-import org.apache.wicket.Component;
 import org.apache.wicket.MarkupContainer;
 import org.apache.wicket.markup.IMarkupResourceStreamProvider;
 import org.apache.wicket.markup.html.WebMarkupContainer;
 import org.apache.wicket.markup.html.WebPage;
 import org.apache.wicket.markup.html.basic.Label;
-import org.apache.wicket.markup.html.list.ListItem;
```
reiern70: Maybe we should add this to agents.md? I have encountered the same problem in classes I edited (becase intelij warns me about this)

papegaaij: Formatting and imports is a real issue with these agents, and we have a very simple solution for that at Topicus: make the build fail if done incorrectly. Use use the spotless-maven-plugin for that. You can give it presets and let it fail the build if something is wrong with imports or formatting. It also has a spotless:apply goal that fixes everyting and you can put in the AGENTS.md file that your agent needs to run it after making changes. I think that's something worth adding to Wicket as well, but I'd rather not do it now, because it requires a full reformat of the entire code base.

reiern70: Clear! And may thanks for letting me know about spotless-maven-plugin

wicket-benchmarks/src/main/java/org/apache/wicket/benchmarks/ComponentStateBenchmark.java
```diff
@@ -0,0 +1,466 @@
...
+		}
+
+		Component newComponent(String id)
+		{
+			Component c = new WebMarkupContainer(id);
+			populate(c);
+			return c;
+		}
+
+		void populate(Component c)
```
reiern70, line 151: Maybe:

1. Make create  interface IShape with a method populate(Component c)
2. Then you make Shape implement this interface and a given enum contant implements populate

Then you don't need the booleans and each populate clearly show what it is populating?

wicket-core-tests/src/test/java/org/apache/wicket/behavior/BehaviorTest.java
```diff
@@ -89,6 +89,21 @@ public void consecutiveTemporaryBehaviorsAreRemoved() {
 		assertFalse(container.getBehaviors().contains(temp2));
 	}
 
+	@Test
+	void nullBehaviorIsRejectedBeforeAnythingIsStored() {
```
reiern70, line 93: JavaDoc?

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
```diff
@@ -89,12 +89,6 @@ protected void onBind()
 		final Component component = getComponent();
 		
 		component.setOutputMarkupId(true);
-
```
reiern70, line 92: Why is this no longer needed?

wicket-core/src/main/java/org/apache/wicket/core/util/resource/locator/ExtensionResourceNameIterator.java
```diff
@@ -42,13 +42,14 @@ public class ExtensionResourceNameIterator implements Iterator<String>
 	public ExtensionResourceNameIterator(final Iterable<String> extensions)
 	{
 		// Fail safe: hasNext() needs to return at least once with true
-		if (extensions == null || !extensions.iterator().hasNext())
+		Iterator<String> extensionIterator = extensions == null ? null : extensions.iterator();
+		if (extensionIterator == null || !extensionIterator.hasNext())
 		{
 			this.iterator = NULL_ITERABLE.iterator();
```
reiern70, line 48: Is this thread safe? If not... is this not important?

wicket-core/src/main/java/org/apache/wicket/ComponentState.java
```diff
@@ -0,0 +1,871 @@
...
+import org.apache.wicket.model.IModel;
+
+/**
+ * This class keeps track of the flexible state of a component: model, behaviors and meta data.
+ * These types of state vary per component. Not every component contains these elements or their
+ * numbers differ. The state is stored in the {@code data} field in {@link Component}. To keep the
+ * size of this state as small as possible, the following cases are identified:
+ * <ul>
+ * <li>No state at all: {@code data} is {@code null}
+ * <li>Only a model: {@code data} contains the model
```
reiern70, line 35: Not belonging to this PR... but many times we have components with several models and then we need to remmeber to manually detach them... maybe wicket should support registering other models and detaching them?

reiern70 (approved):
I'm approving. But this is very difficullt to review. I just made some comments that might make sense or not. It would be nice to port this to 10.x and also host a 10.x. SNAPSHOT somewhere so that we can test  this with other applications (e.g. our application). Thanks for all your work!

wicket-core/src/main/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehavior.java
papegaaij: Because an id no longer has to be handed out in order to exist.

On master an id is a position in a `BehaviorIdList` kept in the component's meta data, allocated in the order `getBehaviorId` is first called. That order is whatever rendering happens to do, which is not reproducible for a stateless page: the page is thrown away and rebuilt on the next request, while the id is already baked into the callback url in the markup. Calling `getBehaviorId` from `onBind` pinned the id to *add* order instead, which is reproducible. That is what the block was for.

On this branch an id *is* the behavior's index in the component's behavior array, so it is add order by construction and there is nothing left to force.

Doing it eagerly would now cost something, too. `getBehaviorId` compacts the behavior array one last time and then sets `FLAG_BEHAVIOR_IDS_FIXED`, which stops it from ever being compacted again so that the ids it hands out stay put. Calling it at bind time would set that flag on every component carrying an ajax behavior, so it would keep the gaps left by removed behaviors for the rest of its life - exactly the footprint this PR is trying to get rid of. The stability guarantee itself is unchanged; it just starts at first use rather than at bind. I wrote the contract down on `IRequestableComponent#getBehaviorId` in 7323aca.

wicket-core/src/main/java/org/apache/wicket/core/util/resource/locator/ExtensionResourceNameIterator.java
papegaaij: It is, and this change does not move that either way.

`ExtensionResourceNameIterator` instances are per lookup - `ResourceNameIterator` constructs one per `locate()` - so they are never shared between threads. The only shared state is `NULL_ITERABLE`, which predates this branch: a one-element `Arrays.asList` that is never mutated, and whose `iterator()` hands out a fresh `ArrayItr` with its own cursor on every call. `ArrayItr` does not implement `remove()`, so `ExtensionResourceNameIterator.remove()` throws `UnsupportedOperationException` on that path rather than touching the shared list.

What the change does is ask `extensions` for an iterator once instead of twice. `hasNext()` does not advance, so reusing the iterator the check was made on is equivalent to throwing it away and asking for a second one.

wicket-core/src/main/java/org/apache/wicket/ComponentState.java
papegaaij: Agreed that it is a real gap. `IModel` has `detach()`, but nothing calls it unless the model is the component's default model or is wrapped by something that passes the call on, so every extra model a component keeps is a detach you have to remember to write yourself.

It is new API rather than a side effect of this rework though, so I would rather not fold it into this PR. Could you open a GitHub issue for it? JIRA was shut down yesterday, so that is where it would live now.

papegaaij:
> I'm approving. But this is very difficullt to review. I just made some comments that might make sense or not. It would be nice to port this to 10.x and also host a 10.x. SNAPSHOT somewhere so that we can test this with other applications (e.g. our application). Thanks for all your work!

It should not be that hard to backport most of this to 10.x, but I'd rather not ship it in a 10.x release. I'm pretty confident that the code is functionally equivalent, but there may be very subtle changes in behavior. One such change is that behaviors can get different ids. This should normally not make any difference, as long as the id is stable over requests, but someone might depend on these numbers. I can create a branch on 10.x for you to test your application against.

wicket-benchmarks/src/main/java/org/apache/wicket/benchmarks/ComponentStateBenchmark.java
papegaaij: You are right about the symptom: `AJAX_BEHAVIOR(false, false, false, true, true)` was unreadable, and two of the three constructors existed only to fill in the defaults the shorter constants did not name.

I went for a variation on your suggestion rather than constant-specific `populate` bodies. Eight of the eleven constants are exactly the combinations of three independent ingredients, so a `populate` per constant would repeat the same three snippets across those combinations, and `hasModel()` - which the `ModelShapes` and `NoModelShapes` states use to split the constants - would need an override per constant as well.

So the ingredients became a `Trait` enum, collected into an `EnumSet` by a varargs constructor:

```java
MODEL_BEHAVIOR(Trait.MODEL, Trait.BEHAVIOR),
STABLE_ID_BEHAVIOR(Trait.STABLE_ID),
AJAX_BEHAVIOR(Trait.STABLE_ID, Trait.AJAX);
```

One constructor, no positional booleans, each constant naming what it carries, `populate` still in one place and `hasModel()` reduced to `traits.contains(Trait.MODEL)`. The javadoc that hung on the stable-id and ajax constants moved onto the traits it actually describes. Constant order is unchanged, so the ordinals and the `@Param` strings the benchmarks are run with still line up. 3ef485a.

wicket-core-tests/src/test/java/org/apache/wicket/behavior/BehaviorTest.java
papegaaij: Added in f2c78b1. It now says what the test protects rather than only what it asserts: `add` validates the whole argument list before it stores any of it, because a behavior's id is its position in the list, so one that was already stored when a later `null` was rejected would keep the position it took and shift the id of everything added after it.

reiern70:
> > I'm approving. But this is very difficullt to review. I just made some comments that might make sense or not. It would be nice to port this to 10.x and also host a 10.x. SNAPSHOT somewhere so that we can test this with other applications (e.g. our application). Thanks for all your work!
> 
> It should not be that hard to backport most of this to 10.x, but I'd rather not ship it in a 10.x release. I'm pretty confident that the code is functionally equivalent, but there may be very subtle changes in behavior. One such change is that behaviors can get different ids. This should normally not make any difference, as long as the id is stable over requests, but someone might depend on these numbers. I can create a branch on 10.x for you to test your application against.

yes please... It would also be nice if we could publish those temporary fixes somewhere with /branchname/m2 repository, this way I can point our gladle to that repository and build as we nomally build (similar to the staging repositories whne releases are voted).

wicket-core/src/main/java/org/apache/wicket/ComponentState.java
papegaaij: As you may have noticed, this Claude did the replying on your comments. Maybe I should have reviewed them first :joy:

wicket-benchmarks/src/main/java/org/apache/wicket/benchmarks/PageSerializationBenchmark.java
solomax: `checkstyle` might also help here :)

